PR #1, opened 2015-01-20
Add .gitignore with anchored directories

LeoVerto:
Mostly generated by https://gitignore.io, thanks to kuno for pointing out that directories weren't anchored, which could have prevented users from noticing messed up source trees.


LordSputnik:
Looks good to me, merged!

PR #2, opened 2015-01-25
Set up serve-favicon and return to present

LeoVerto:


LordSputnik:
Looks good to me, thanks (just read up on serve-favicon, looks like it will be beneficial). Might change the year to "2014 - 2015" though - will commit this myself.

PR #35, opened 2015-03-09
Implement passport.js authentication

leftmostcat:
This doesn't currently make use of the auth.isAuthenticated() function, which will be done on a route-by-route basis, but it does appear to work with all existing code, with exceptions noted in the message for the second commit.


leftmostcat:
Accidentally closed. Should not have been merged.

PR #37, opened 2015-03-10
Add rudimentary config helper

leftmostcat:
This currently only works for server-side scripts and requires that configuration be specified in config/cfg.js. The code should be adapted to allow pulling in configuration per-environment and to allow the server admin to specify their own config file location.


LordSputnik:
Looks good to me, although we'll need to modify the travis configuration to avoid testing errors caused by missing config files.

PR #53, opened 2015-04-08
Restructure the root directory of the site

LordSputnik:
As discussed in http://chatlogs.musicbrainz.org/bookbrainz-devel/2015/2015-04/2015-04-06.html#T22-34-32-498230


leftmostcat:
Parts of this I'm not sure how they're working and some things I think we ought to rework a little. rootRequire(), for example, no longer uses even src/ as a root, but whatever is in app.js. This means we get weird things like rootRequire('../../config') when config/ is in the root and rootRequire('helpers/bbws') working even though helpers is not in the root. This probably merits some rethinking of how we structure requires.

LordSputnik:
I'm thinking that the root directory for the source should be src/server, and not the top directory, since that contains very little code of its own. I wouldn't be against moving the config into src/server too.

PR #58, opened 2015-06-02
Move a space out of an a tag

mineo:


LeoVerto:
Yeah, that's a common one.

PR #77, opened 2015-12-27
BB-101: Added guess button and logic to alias sortName

opatel99:
I first added a buttonAfter to the Alias sortName input. Afterwards, I created a guess() function that reads the name of the alias and sets the sortName input value to a generates a "guess" based on naming standards.
Although this guess method will not correctly apply to some cases, many entries will benefit from the usage of a quick generator.

http://tickets.musicbrainz.org/browse/BB-101

### DIFF
--- a/src/client/components/forms/parts/aliases.jsx
+++ b/src/client/components/forms/parts/aliases.jsx
@@ -72,6 +72,25 @@ const AliasRow = React.createClass({
 			this.refs.name.getValue() && this.refs.sortName.getValue()
 		);
 	},
+	guess() {
+		'use strict';
+		const name = this.refs.name.refs.input;
+		const sortName = this.refs.sortName.refs.input;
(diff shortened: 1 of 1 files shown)


LeoVerto:
I'd suggest renaming the function "guess" to "guessNames".
Adding a "title" attribute to the button for better accessibility (for example for screen readers) probably also wouldn't be a bad idea.

Other than that this looks pretty good, but I'd say we don't merge until your style PR is accepted and we can come up with definitive tests.

LordSputnik:
This looks good to me. Since it isn't "Guess Capitalization" I don't think it'll be affected by the style PR.

I'll merge this now then update it with some of the code from the other submission for this GCI task.

opatel99:
I hereby license the code encompassed in this pull request under the GPLv2+ license.

PR #78, opened 2016-01-02
BB-154: Align navbar search input with button by forcing line-height and height

opatel99:
I wasn't particularly fond of using !important to force the line-height and height, but this seems to be an issue with Bootstrap itself, so issuing an override is what I deemed best.

The issue can be replicated by visiting [Bootstrap Input-Groups](http://getbootstrap.com/components/#input-groups-basic) and manipulating zoom percentage.

As of now, I propose this solution.

nav-search

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -45,6 +45,11 @@ body {
 .whole-page-form {
 	margin-top: 3em;
 }
+.navbar-form .form-control,
+.navbar-form .input-group-btn>.btn {
+  line-height: 1 !important;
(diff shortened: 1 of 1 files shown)


mineo:
Just to clarify: I don't see any broken behaviour visiting that bootstrap page at any zoom level I tried.

LeoVerto:
I can reproduce it on that site but only at certain zoom levels, it looks good overall.

I think this isn't a Bootstrap issue but exclusive to [lobes](https://github.com/metabrainz/lobes).

PR #79, opened 2016-01-03
BB-152: Fix navigation menu staying depressed

opatel99:


LordSputnik:
As commented on JIRA, this doesn't seem to fix the issue (see http://bookbrainz.mbsandbox.org)

opatel99:
There seems to have been a slight miscommunication on my end. What I attempted to fix was the "hamburger" menu when the navbar collapses into one option. That menu button also stayed depressed upon initial click. That has been fixed. However, it is my understanding that the intended buttons cannot be fixed simply. There is a JS script that could emulate this request, but I have not found any pure CSS solution as of yet.

PR #83, opened 2016-03-18
Minor changes to instructions from readme

heshoots:
I have just set up the development environment and have found a two sections slightly confusing that required checking some outside resources, included are the commands I used to get it working


LordSputnik:
Looks good to me :+1:

PR #85, opened 2016-04-10
April 2016 React Cleanup

LordSputnik:
These few commits reduce the number of eslint warnings from 200+ to about 60. String literal refs (such as `ref="input"`) have been removed, since these are considered legacy by React (https://facebook.github.io/react/docs/more-about-refs.html#the-ref-string-attribute). Callback props have been named on<XYZ> consistently, and now always call functions called handle<ABC>. Also introduced stateless functional components (https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components), and deleted unused Jade templates which have been ported to React components.

The main thing we still need to fix is migrating old `React.createClass` components to use ES6 classes. However, this needs us to rethink our build/development/production process a little bit, since we'll need to transpile both client and server code with babel in production, and run something like babel-node in development (ES6 classes aren't yet available in Node 5.x).

Also need to refactor components to avoid using `bind` and storing arrays of refs to components.


leftmostcat:
Is there any reason we can't use ES6 classes with 'use strict'? We could retain function-level strict mode with an anonymous function call.

LordSputnik:
That seems a bit hacky to me, shouldn't be too much effort to update the build process and we'll want to do it to get other ES6 features too.

leftmostcat:
It's a little bit hacky, but it avoids adding to the build process (which is already a bit frustrating to me for iterative development and testing) and it also provides a clear path for future build trimming. (Part of the work I was doing with bower was with the goal in mind of breaking apart the bundle and directly using vendor JS.)

LordSputnik:
What do you find frustrating about the current process?

leftmostcat:
My preference would be to minimize the amount of build needed for dev iteration, in part by eliminating bundling and directly including vendor JS plus only compiling JSX for the individual files that require it.

The punchline is that I'd eventually like to see ES6-classed react components with global 'use strict'. We can browserify it however we need to, but I don't think we should require further building for server-side dev.

LordSputnik:
But what if there are features in ES6 that we want to use that aren't in node?

My envisaged development process is this:
1. Use babel-node to transpile ES6/7 syntax on the fly
2. Modify "npm start" to run nodemon + babel-node
3. Set up a separate npm command to run watchify to auto-build updated bundles and auto-compile LESS
4. Write commands to build server and client code for production

The advantage to bundling locally installed JS is that you don't need an internet connection to develop. Pulling vendor JS from a CDN would require you to be online.

PR #86, opened 2016-04-10
Pull in entity data set schema changes

leftmostcat:
There are two major changes wrapped up in this branch: one, the changes directly related to the entity set implementation in -sql and -data-js, which covers pulling in the model relations that were set up and appropriately changing references; and two, changes to editing to both accommodate and take advantage of these changes. This allows entity editing to be much more generic and, as a major bonus, more easily calculate whether there are changes during editing that necessitate a new revision.

N.B.: While the bulk of these changes have been tested reasonably thoroughly, Edition and Work creation and editing need testing and a general smoketest is needed to make sure that nothing else broke horribly.

### DIFF
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -437,162 +437,183 @@ function formatScalarChange(change, label) {
 	return [formatChange(change, label, (side) => side && [side])];
 }
 
-function formatEditionLanguages(change) {
+function formatNewReleaseEventSet(change) {
 	const rhs = change.rhs;
-
-	if (rhs && rhs.length) {
+	if (rhs.releaseEvents && rhs.releaseEvents.length > 0) {
 		return [formatRow(
-			'N', 'Language', null, rhs.map(
-				(language) => language.name
+			'N', 'Release Date', null, rhs.releaseEvents.map(
+				(releaseEvent) => releaseEvent.date
 			)
 		)];
-	}
-
-	return null;
-}
-
-function formatEditionLanguageAddOrDelete(change) {
-	return [
-		formatChange(
-			change.item,
-			`Language ${change.index}`,
-			(side) => side && [side]
-		)
-	];
-}
-
-function formatEditionLanguageModified(change) {
-	return [
-		formatChange(
-			change,
-			`Language ${change.path[1]}`,
-			(side) => side && [side]
-		)
-	];
-}
-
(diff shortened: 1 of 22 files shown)


src/server/routes/revision.js
```diff
@@ -437,162 +437,183 @@ function formatScalarChange(change, label) {
...
+	const lhs = change.item.lhs && [change.item.lhs.date];
+	const rhs = change.item.rhs && [change.item.rhs.date];
+
 	return [
-		formatChange(
-			change.item,
-			`Language ${change.index}`,
-			(side) => side && [side]
-		)
+		formatRow(change.item.kind, 'Release Date', lhs, rhs)
```
leftmostcat: Could we use `formatChange()` here? Should be as simple as `formatChange(change.item, 'Release Date', (side) => side && [side.date])` if I'm not missing something.

src/server/routes/revision.js
```diff
@@ -437,162 +437,183 @@ function formatScalarChange(change, label) {
...
+	const lhs = change.item.lhs && [change.item.lhs.name];
+	const rhs = change.item.rhs && [change.item.rhs.name];
+
 	return [
-		formatChange(
-			change.item,
-			`Publisher ${change.index}`,
-			(side) => side && [side.bbid]
-		)
+		formatRow(change.item.kind, `Language ${change.index}`, lhs, rhs)
```
leftmostcat: Same here. ``formatChange(change.item, `Language ${change.index}`, (side) => side && [side.name])`` unless I'm missing something.

src/server/routes/revision.js
```diff
@@ -437,162 +437,183 @@ function formatScalarChange(change, label) {
...
 }
 
-function formatReleaseEventAddOrDelete(change) {
+function formatPublisherAddOrDelete(change) {
+	const lhs = change.item.lhs && [change.item.lhs.bbid];
+	const rhs = change.item.rhs && [change.item.rhs.bbid];
+
 	return [
-		formatChange(change.item, 'Release Date', (side) => side && [side.date])
+		formatRow(change.item.kind, 'Publisher', lhs, rhs)
```
leftmostcat: As above.

leftmostcat:
A couple comments on the diff code, and as mentioned on IRC, I'd like the revision rearranging to be temporary, but otherwise the changes seem merge-ready.

PR #89, opened 2016-04-29
BB-186 Add error classes for major error categories

leftmostcat:
### Problem

Error handling is inconsistent across the site.

### Solution

Add a handful of convenience classes for handling errors, as well as functions to facilitate display.

### Areas of Impact

Error handling throughout the server-side code has been touched. Login, registration, and entity deletion forms have all been rewritten or heavily refactored to accommodate changes to error handling on form handlers.

Entity creation and editing, including the relationship editor, should be largely unchanged; error handling on the client-side for these remains largely broken for now.


src/server/routes/entity/entity.js
```diff
@@ -312,7 +324,6 @@ function processFormAliases(
 		}
 	);
 }
-module.exports.processFormAliases = processFormAliases;
```
LordSputnik: Are these exports unused?

app.js
```diff
@@ -111,38 +110,18 @@ app.use((req, res, next) => {
...
-app.use((err, req, res) => {
-	res.status(err.status || status.INTERNAL_SERVER_ERROR);
+const error = require('./src/server/helpers/error');
 
-	res.render('error', {
-		message: err.message,
-		error: {}
-	});
+// Error handler; arity MUST be 4
+app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
```
LordSputnik: Why do we need to define next here?

leftmostcat: The arity of the handler has to be four or express doesn't recognize it as an error handler. If the prototype is (err, req, res) then we're actually getting (err=req, req=res, res=next).

src/server/routes/entity/entity.js
leftmostcat: They are, and they strike me as internal functionality.

PR #91, opened 2016-05-02
BB-190 Convert React components to one-per-file ES6 classes

leftmostcat:
### Problem

React components are currently implemented with React.createClass(). ES6 classes will provide easier inheritance and are more future-proof. Additionally, many files declare multiple React components.

### Solution

Port all React components to ES6 classes and split multi-component files.

### Areas of Impact

All non-static React components.


LordSputnik:
Couple of things:

1. Why switch all the filenames from camelCase to kebab-case?
2. Does browserify move the global 'use strict' to a function 'use strict'?

leftmostcat:
1. This was mostly just personal preference and a desire for consistency (which a couple of files lacked). Camel case has never made a lot of sense to me on the filesystem. I took the opportunity, since I was splitting files, to make the change across all react components. It's not an important change, and can be undone if preferred, but it does seem that—for files, at least—even node projects have a preference for kebab case at that level.
2. Since it is an immediate invocation, it's counted as function-scoped. Using the immediate invocation is a bit of a hack, but I went with it since it allows us to move forward. My own preference would be to find a way to move toward global-scoped use strict in these components so we could avoid the hack, but for now it's a fairly simple workaround for the issue.
3. I've done cursory testing on most of the forms and some more extensive testing on components that had more invasive changes (splitting, binding, that sort of thing). The forms are definitely rendered server-side, but I haven't been able to verify that isomorphic rendering is working. I can certainly do this testing, but I could some suggestions on how if you have them.

PR #92, opened 2016-06-28
Single-sign on using MB credentials

LordSputnik:
### Problem

The BookBrainz account system is currently separate from the MusicBrainz account system, creating an obstacle to new editors signing up for BookBrainz. Additionally, the separate system requires users to trust BookBrainz with some private information (password, email), which is no longer necessary if this information is stored in MusicBrainz. 

### Solution

This PR introduces a single-sign-on solution to allow MusicBrainz editors to register for BookBrainz using their MusicBrainz account, and subsequently log in to BookBrainz using their MusicBrainz credentials. Existing BB users can link their BB accounts to MB accounts through a link on the profile page. Public profile information, such as display name and gender, can be set on BookBrainz independent of information stored in MusicBrainz.

### Areas of Impact
Affects the log in, registration and authentication code and the base layout template (logged in user display).


LordSputnik:
@leftmostcat any feedback on this? Thanks!

LordSputnik:
Self-reviewed and merged - as far as I can tell, all changes are sane and user-friendly.

PR #93, opened 2016-06-29
Gamification dummy functions

heshoots:
### Problem
Produce start of achievement module for BookBrainz

### Solution
Implements a "Revisionist" achievement awarded to anyone with a revision attached to their account

### Areas of Impact
Effects the editing of an entity, when an entity is edited, the processEdit function is called which will call all process<Achievements> later into the modules development

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "body-parser": "^1.14.1",
-    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git#gamification",
+    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git",
     "bootstrap": "^3.3.2",
     "connect-redis": "^3.0.1",
     "debug": "^2.2.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -8,10 +8,13 @@
...
     "test": "npm run lint && gulp"
   },
+  "directories": {
+	"test": "tests"
+  },
   "dependencies": {
     "bluebird": "^3.3.4",
     "body-parser": "^1.14.1",
-    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git",
+    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git#gamification",
```
heshoots: I will need to change this once the bookbrainz-data-js pull request is merged

package.json
```diff
@@ -8,6 +8,9 @@
     "lint": "eslint --ext .js --ext .jsx src test templates scripts app.js gulpfile.js",
     "test": "npm run lint && gulp"
   },
+  "directories": {
+	"test": "tests"
```
LordSputnik: What's this bit for?

test/testAchievement.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
```
LordSputnik: Missing a copyright header here!

PR #96, opened 2016-07-29
Achievement jsdoc

heshoots:
### Problem

Add HTML documentation to the BookBrainz project

### Solution

Adds JSdoc to the project and document the achievement module

### Areas of Impact

Package.json
Achievement Module

### DIFF
--- a/src/server/helpers/achievement.js
+++ b/src/server/helpers/achievement.js
@@ -56,6 +56,14 @@ function awardUnlock(UnlockType, awardAttribs) {
 		);
 }
 
+/**
+ * Awards an Achievement
+ * @param {int} editorId - The editor the achievement will be awarded to
+ * @param {string} achievementName - Name of achievement in database
+ * @returns {object} - {achievementName: unlock} where unlock is JSON returned
+ * from awardUnlock
+ * @memberof module:Achievement
+ */
 function awardAchievement(editorId, achievementName) {
 	return new AchievementType({name: achievementName})
 		.fetch({require: true})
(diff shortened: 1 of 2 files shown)


src/server/helpers/achievement.js
```diff
@@ -94,9 +121,19 @@ function awardTitle(editorId, tier) {
 	return titlePromise;
 }
 
-function awardListToAwardObject(tier) {
+/**
+ * Reformats a list of list of achievements objects to an object keyed by the
+ * achievement names
```
LordSputnik: Could you explain in the comment what this is used for/why it is needed?

src/server/helpers/achievement.js
```diff
@@ -140,7 +191,13 @@ function testTiers(signal, editorId, tiers) {
...
 }
 
-// returns the number of typeRevisions an editor has
+/**
+ * Returns number of revisions of a certain type there are for the specified
+ * editor
+ * @param {string} type - Camelcase name of the type to query
+ * @param {int} editor - Editor id being queried
+ * @returns {int} - Number of revisions of type (type)
+ */
```
LordSputnik: Unrelated to this PR, but we should probably move this function to a property of the Editor module in -data-js, so we can just do Editor.getTypeRevisions(...) (possibly useful in the WS too).

src/server/helpers/achievement.js
```diff
@@ -222,16 +279,12 @@ function processSprinter(editorId) {
...
-	Object.keys(achievementUnlock).forEach((key) => {
-		if (achievementUnlock[key].id) {
-			unlockIds.push(String(achievementUnlock[key].id));
-		}
-	});
-	return unlockIds;
-}
-
+/**
+ * Gets number of distinct days the editor made an edit on within set limit
```
LordSputnik: "made an edit on within set limit" could be replaced with "created revisions within specified time period"

PR #97, opened 2016-08-02
Revert partial-date to React.createClass form

heshoots:
### Problem
The ES6 version of partial-date does not work and prevents users from creating editions

### Solution

Reverts the change back to React.createClass form

### Areas of Impact

Partial-date class in client/components/input/

It would be preferable to stick with es6 syntax if a fix is found but this could be an issue if pushed to the live server.

### DIFF
--- a/src/client/components/input/partial-date.jsx
+++ b/src/client/components/input/partial-date.jsx
@@ -25,102 +25,91 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
 const ymRegex = /^\d{4}-\d{2}$/;
 const yRegex = /^\d{4}$/;
 
-(() => {
-	'use strict';
-
-	class PartialDate extends React.Component {
-		static validate(value) {
-			if (!value) {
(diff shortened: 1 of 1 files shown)


src/client/components/input/partial-date.jsx
```diff
@@ -25,102 +25,91 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
...
-				value: this.props.defaultValue,
-				valid: PartialDate.validate(this.props.defaultValue)
-			};
-
-			// React does not autobind non-React class methods
-			this.handleChange = this.handleChange.bind(this);
-		}
-
-		getValue() {
-			return this.validate(this.input.getValue().trim());
```
heshoots: This line is the problem. However once this is changed this component returns true/false instead of the date, which causes problems in other areas

LordSputnik:
Fixed without reverting in 2ec0e6303a17b095345a14f8562f0684f740d01e

PR #99, opened 2016-08-09
Time achievements

heshoots:
### Problem

Site does not encourage making edition close to or before the release date

### Solution

Add Hot off the press and Time traveller achievement

### Areas of Impact

Achievement module
Unit test

### DIFF
--- a/src/server/routes/entity/entity.js
+++ b/src/server/routes/entity/entity.js
@@ -535,7 +535,7 @@ module.exports.createEntity = (
 			);
 	});
 
-	entityCreationPromise.then((creationJSON) =>
+	const achievementPromise = entityCreationPromise.then((creationJSON) =>
 		achievement.processEdit(
 			creationJSON.editorJSON.id,
 			creationJSON.entityJSON.revisionId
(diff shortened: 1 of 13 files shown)


src/server/routes/entity/entity.js
```diff
@@ -519,20 +519,30 @@ module.exports.createEntity = (
...
+		achievement.processEdit(
+			creationJSON.editorJSON.id,
+			creationJSON.entityJSON.revisionId
+		)
+			.then(() => creationJSON.entityJSON)
+	);
+
 	return handler.sendPromiseResult(
 		res,
 		entityCreationPromise,
```
LordSputnik: Hmm, this should be the promise for the achievement system processing the edit, not `entityCreationPromise`.

heshoots: The reason I did this was because I need access to the revision in order to get information from it in the achievement system. Moving it elsewhere means the revision is not yet in the database. How would you recommend handling this otherwise?

LordSputnik: The processing is fine, but just assign this new promise to a variable, then send that to the handler, so that the route doesn't respond to the client before it's finished processing:

    const achievementPromise = entityCreationPromise.then((creationJSON) => ...

        res,
        achievementPromise,

PR #100, opened 2016-08-09
Title integration

heshoots:
### Problem

Titles are only integrated into the profile page, it would be good to have them in other areas of the site

### Solution

Integrates titles into revision page such that you can hover over them and see the users title

### Areas of Impact

Revision page
Revision route

### DIFF
--- a/src/client/components/pages/revision.jsx
+++ b/src/client/components/pages/revision.jsx
@@ -96,6 +96,18 @@ const EntityLink = require('../entity-link.jsx');
 			return _compact(result);
 		}
 
+		static formatTitle(author) {
+			let title;
+			if (author.titleUnlock.title) {
+				const authorTitle = author.titleUnlock.title;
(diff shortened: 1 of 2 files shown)


src/client/components/pages/revision.jsx
```diff
@@ -117,6 +117,9 @@ const EntityLink = require('../entity-link.jsx');
 				</div>
 			));
 
+			const editorTitle =
```
LordSputnik: This is slightly problematic - it will soon be possible for editors to reply to notes on revisions, so there will be notes by several different editors. So the title should be retrieved on a note-by-note basis. If there's a `title` relation on the editor data model, then that should be fairly easy to do by modifying the server code that fetches the note author to also fetch `author.title`.

heshoots: Fixed in the latest commit, the title-integration pull request on data-js will have to be merged beforehand though

PR #101, opened 2016-08-09
Drag and Drop achievement form

heshoots:
### Problem

Achievement badge seletor is currently a select form that could be more 'fun' to use

### Solution

Convert it to a drag and drop menu that fits with modern expectations

### Areas of Impact

Achievement form on profile page


templates/layout.jade
```diff
@@ -5,6 +5,7 @@ html
 		link(rel='stylesheet', href='/stylesheets/font-awesome.css')
 		link(rel='stylesheet', href='/stylesheets/style.css')
 		link(rel='stylesheet', href='/stylesheets/lib/select2.css')
+		link(rel='stylesheet', href='/stylesheets/lib/dndSelector.css')
```
LordSputnik: dndSelector.css doesn't seem to be added in the commit or generated by any code

heshoots: It was added here, https://github.com/bookbrainz/bookbrainz-site/pull/101/commits/627e6c339064402421958bf1fbb4d75cbb544d1c
I think git refused to add it for a while because of the gitignore.

PR #102, opened 2016-08-10
Achievement Notification

heshoots:
### Problem

Users are not notified when they get an achievement. 

### Solution

Add an alert when an achievement is unlocked

### Areas of Impact

Entity route
Entity Forms

### DIFF
--- a/templates/entity/view/entity.jade
+++ b/templates/entity/view/entity.jade
@@ -22,7 +22,7 @@ block body
 						each achievement in alert
 							p #{achievement.name}
 					div
-						strong Go to your profile to display
+						strong To showcase the badges for your new achievements go to your profile page
 	.row
 		.col-md-12
 			h1


src/server/routes/entity/entity.js
```diff
@@ -54,10 +56,62 @@ module.exports.displayEntity = (req, res) => {
...
 		);
 
-	res.render(
-		`entity/view/${entity.type.toLowerCase()}`,
-		{identifierTypes}
-	);
+	let editorEntityVisitPromise;
+	if (res.locals.user) {
+		editorEntityVisitPromise = new EditorEntityVisits({
+			editor_id: res.locals.user.id,
```
LordSputnik: Think this should be camelCase not snake_case

src/server/routes/entity/entity.js
```diff
@@ -526,12 +580,9 @@ module.exports.createEntity = (
 						transacting
 					})
 			)
-			.then((entity) =>
-				({
-					entityJSON: entity.toJSON(),
-					editorJSON
-				})
-
+			.then((entity) => entity.toJSON())
```
LordSputnik: What's the reason for this change?

templates/entity/view/entity.jade
```diff
@@ -13,6 +13,16 @@ mixin begin_end
...
+	if alert
+		.row
+			.col-md-12
+				.alert.alert-info
+					div
+						strong Achievement unlocked:
+						each achievement in alert
+							p #{achievement.name}
+					div
+						strong Go to your profile to display
```
LordSputnik: I think this guideline is a bit confusing - if I were a new user, I'd think "display what?". Maybe showcase would be a better word? I don't know, try to think of a different way of saying it! :)

PR #103, opened 2016-08-10
Explorer achievement

heshoots:
### Problem

No achievement to encourage looking around the entities on the site

### Solution

Explorer achievement is obtained by viewing many entities on the site

### Areas of Impact

Entity route
Entity page
Unit tests
Achievement module

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "body-parser": "^1.14.1",
-    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git#explorer",
+    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git",
     "bootstrap": "^3.3.2",
     "co": "^4.6.0",
     "connect-redis": "^3.0.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
   "dependencies": {
     "bluebird": "^3.3.4",
     "body-parser": "^1.14.1",
-    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git",
+    "bookbrainz-data": "git+https://github.com/bookbrainz/bookbrainz-data-js.git#explorer",
```
heshoots: Change back before merging

src/server/helpers/achievement.js
```diff
@@ -113,7 +118,10 @@ function awardTitle(editorId, tier) {
 						const out = {};
 						out[tier.titleName] = unlock;
 						return out;
-					});
+					})
+					.catch((err) =>
+						console.log(err)
```
LordSputnik: Are you sure the default error handler doesn't print them out (in app.js)?

heshoots: Its not something I can test in the actual application since its unlikely that an incorrect editor Id will get passed in. But it caused a massive problem in writing unit tests since I was getting an UnlockNotFound error, which was hiding the actual issue of the editorId not being found, I Will document this to make note of it, but I personally think it should stay.

LordSputnik: Hmm, is there no way to detect the editor ID error and report it by throwing a different kind of error?

PR #104, opened 2016-09-03
Remove option to log in using local strategy

LordSputnik:
### Problem
BookBrainz recently allowed editors to log in using MusicBrainz credentials. A sufficient proportion of active users have migrated to the new login system, and manual methods exist for moving legacy users to the new system. Therefore, the local (username/password) method of logging in to BookBrainz is redundant. It is also preventing us from writing a simple database export script to create regular database dumps (passwords make things complex), and that in turns stops us creating Docker images for BookBrainz.

### Solution
Remove the UI, routes and helpers allowing users to log in locally. Corresponding PRs to the data and sql repositories will remove relevant model methods and schema fields.

### Areas of Impact
Login and registration


leftmostcat:
Looks good to me.

PR #105, opened 2016-09-12
New build system

LordSputnik:
### Problem

The current build system relies on gulp and related modules to programmatically compile and process resources. This means that additional code must be maintained (the gulpfile) and the build system must remain compatible with the additional dependencies. On top of this, the gulp process must be started separately from the app process, and gulp does not watch and rebuild LESS files.

Additionally, running the code natively with node.js means that the supported set of ES6 features is entirely dependent on the version of node.js being used. If more advanced ES6 features are used, the range of supported Node.js environments is significantly decreased.
### Solution

This PR removes the previous gulp-based build system and replaces it with a new CLI-based build system, defined in the package.json file. The new build system:
- Uses babel-node to perform just-in-time compilation of ES6 JS during development
- Uses babel to transpile ES6 JS to JS compatible with whatever version of Node is running in production
- Watches server JS, client JS and LESS files to determine when to rebuild/reload
- Allows the use of JSX files without the JSX prefix, simplifying require/import statements
- Reduces the number of dependencies required in all environments
- Copies up-to-date client JS dependencies from node_modules instead of storing fixed versions in the repository
### Areas of Impact

All client JS files have changed minimally. The entry point of the app has changed from scripts/www to src/index.js (to allow this to be correctly transpiled). And (obviously!) the build system has been rewritten.


package.json
```diff
@@ -3,31 +3,51 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "start": "node ./scripts/www",
-    "debug": "cross-env DEBUG=bbsite nodemon ./scripts/www",
-    "lint": "eslint --ext .js --ext .jsx src test templates scripts app.js gulpfile.js",
+    "copy-bootstrap": "cp -R node_modules/bootstrap/less/ src/client/stylesheets/bootstrap",
```
leftmostcat: A lot of these are pretty gnarly. It would be very helpful to break them out into sh scripts or the like to make them easier to maintain.

.eslintrc
```diff
@@ -5,13 +5,10 @@ ecmaFeatures:
   - generators
 env:
   node: true
+  es6: true
```
leftmostcat: Most of these changes look fine, but CI is heavily broken due to many of them. I would like one of three things to happen: this PR depends on one which adjusts `.eslintrc` and cleans up those failures, this PR cleans up those failures itself, or the bulk of the `.eslintrc` changes are left out for now and addressed later.

src/server/app.js
```diff
@@ -65,13 +61,13 @@ Promise.config({
 const app = express();
 
 // Set up jade as view engine
-app.set('views', path.join(__dirname, 'templates'));
+app.set('views', path.join(__dirname, '../../templates'));
```
leftmostcat: It feels a little bad to be making this so relative, but could we mitigate by having a single `rootPath` const that encapsulates the `__dirname` and `../../` parts, then join that with the relative paths for the other bits?

.eslintrc
LordSputnik: I'm currently fixing all of these in the new-entity-editing branch. I should've cleaned them up before I opened that. I'd like to leave it broken for now and fix it in the next PR, to avoid a ton of merge conflicts.

src/server/app.js
LordSputnik: That sounds like a good plan.

.eslintrc
leftmostcat: How many of the settings changes are necessary to accommodate the changes to the build? Would it be possible to pull them into the other PR and drop them from this one?

scripts/watch-less.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd src/client/stylesheets
+catw -c 'lessc --include-path=bootstrap -' \
```
leftmostcat: Is there a particular reason that bootstrap and font-awesome are handled differently for include? If so, it might be good to add a comment about that.

scripts/copy-client-scripts.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+cp -R node_modules/bootstrap/less/ src/client/stylesheets/bootstrap
```
leftmostcat: Should the copy scripts implicitly clean before copying over?

LordSputnik: Don't think so. They're only called by prepublish, which is already cleaning and setting up directories first. This is only an issue if someone calls copy-client-scripts.sh directly, and if they do that, they should have a good reason.

package.json
```diff
@@ -3,31 +3,49 @@
...
   "scripts": {
-    "start": "node ./scripts/www",
-    "debug": "cross-env DEBUG=bbsite nodemon ./scripts/www",
-    "lint": "eslint --ext .js --ext .jsx src test templates scripts app.js gulpfile.js",
-    "test": "npm run lint && gulp",
+    "copy-client-scripts": "./scripts/copy-client-scripts.sh",
+    "mkdirs": "mkdir -p static/js/editor static/js/entity static/stylesheets",
+    "clean": "./scripts/clean.sh",
+    "prepublish": "npm run clean && npm run mkdirs && npm run copy-client-scripts",
+    "build-server-js": "babel src --out-dir lib",
```
leftmostcat: It seems like this creates an implicit, undocumented dep on a global babel install (which is also currently not reflected in the travis config). Could we depend on a local babel install for this?

LordSputnik:
Yes, in fact I added that on a commit on the new-entity-editing branch last
week. Can add it here too.

LordSputnik:
No need to put it in a separate PR - it fits in fine as part of the new
build system changes :)

PR #106, opened 2016-10-29
Remove usage of IIFEs for class definition

leftmostcat:
### Problem

Immediately-invoked function expressions are a hack to avoid the node 4.x requirement of declaring a file to be strict mode before defining classes. With babel being used as a default part of the build, it is avoidable.

### Solution

Remove all IIFEs.

### Areas of Impact

React components.

### Dependencies
Dependent on https://github.com/bookbrainz/bookbrainz-site/pull/105.


LordSputnik:
Looks good to me once #105 is in!

PR #109, opened 2016-11-06
New entity editing

LordSputnik:
### Problem
The current entity editing forms in BookBrainz are overly complex. They don't flow well, use tabs although they aren't really needed, and it isn't really clear which information is required and which is optional.

### Solution
This PR introduces a new entity editor form for the creator entity, as a prototype that can be developed further based on feedback and extended to other entities prior to the next release. The prototype drops tabs, displays only a minimum of inputs by default and is based on last year's GCI UI concept drawings and concepts drawn up by myself and Leftmost 

### Areas of Impact
The client code for the creator editor has been replaced with this new code, and an additional function has been added to the server code processing the creator editor data to transform the new form data to the shape of the old form data. Style and syntax issues left over from #105 are also addressed.


src/client/entity-editor/common/error-text.js
```diff
@@ -0,0 +1,43 @@
...
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import {Alert} from 'react-bootstrap';
+import React from 'react';
+import {connect} from 'react-redux';
+
+function ErrorText({
```
leftmostcat: `ErrorToast`?

src/client/entity-editor/common/language-field.js
```diff
@@ -0,0 +1,37 @@
...
+import VirtualizedSelect from 'react-virtualized-select';
+
+function LanguageField({
+	...props
+}) {
+	return (
+		<Input
+			label="Language"
+		>
+			<VirtualizedSelect {...props}/>
```
leftmostcat: It means more boilerplate, but how would you feel about explicitly specifying these props? It makes it a little difficult at this level to see what behavior will be affected at the `VirtualizedSelect` level.

src/client/entity-editor/common/name-field.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+function NameField({
+	empty,
+	error,
+	onChange,
+	...rest
+}) {
+	const label = (
+		<ValidationLabel
+			empty={empty}
```
leftmostcat: Unclear what the purpose of this is from its name. Since it's a bool, I assume `isEmpty` is the intent.

src/client/entity-editor/common/name-field.js
```diff
@@ -0,0 +1,61 @@
...
+			error={error}
```
leftmostcat: Since this is also a bool, `hasError`?

src/client/entity-editor/common/sort-name-field.js
```diff
@@ -0,0 +1,145 @@
...
+ * string.
+ *
+ * @param {String} name the input string to strip
+ * @returns {String} the string with dots removed
+ */
+function stripDot(name) {
+	return name.replace(/\./g, '');
+}
+
+function makeSortName(name) {
```
leftmostcat: We'll want to give this code a more thorough review at some point. It may make the most sense to make it a specifiable parameter, since different entity types (and even different form field values) will require different behaviors.

src/client/entity-editor/common/sort-name-field.js
```diff
@@ -0,0 +1,145 @@
...
+	let lastName = words.splice(INDEX_BEFORE_END);
+	if (suffixWords.length > 0) {
+		lastName += ` ${suffixWords.join(' ')}`;
+	}
+
+	return `${lastName}, ${words.join(' ')}`;
+}
+
+/**
+ * Presentational component. Renders the sort name field for the alias section
```
leftmostcat: I'm liking these descriptive components. I am in favor of adding them to the components that are missing them as well. 👍

src/client/entity-editor/common/submit-button.js
```diff
@@ -0,0 +1,88 @@
...
+
+function submit(url, data) {
+	return request.post(url).send(data)
+		.promise()
+		.then((response) => {
+			if (!response.body) {
+				window.location.replace('/login');
+			}
+
+			const redirectUrl = `/creator/${response.body.bbid}`;
```
leftmostcat: Since this is a generic entity form, we'll need to parameterize this.

src/client/entity-editor/common/submit-button.js
```diff
@@ -0,0 +1,88 @@
...
+			{errorElement}
+		</div>
+	);
+}
+SubmitButton.displayName = 'SubmitButton';
+SubmitButton.propTypes = {
+	error: React.PropTypes.string
+};
+
+function submit(url, data) {
```
leftmostcat: A lot of this behavior seems like it will be shared by all of our forms. The basic pattern of "check authentication, log in if lacking, continue". Seems like factoring this out would be good. We may also want to modalize login. All this is outside the scope of this PR, though.

src/client/entity-editor/creator-data.js
```diff
@@ -0,0 +1,98 @@
...
+
+
+import AliasEditor from './alias-editor/alias-editor';
+import CreatorData from './creator-data/creator-data';
+import IdentifierEditor from './identifier-editor/identifier-editor';
+import {Panel} from 'react-bootstrap';
+import React from 'react';
+import SharedData from './shared-data/shared-data';
+import {connect} from 'react-redux';
+
```
leftmostcat: #nit: Excess whitespace.

src/client/entity-editor/creator-data.js
```diff
@@ -0,0 +1,98 @@
...
+
+let Wrapper = ({
```
leftmostcat: Not sure I like the name `Wrapper`. If that's all this is, I wonder if it should be split into multiple components or subsumed by another.

src/client/entity-editor/creator-data/begin-field-container.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+import DateField from './date-field';
+import _debounce from 'lodash.debounce';
+import {connect} from 'react-redux';
+import {updateBeginDate} from '../actions';
+
+const KEYSTROKE_DEBOUNCE_TIME = 250;
+
+function isPartialDateValid(value) {
```
leftmostcat: This seems like functionality that belongs as part of the date field component.

src/client/entity-editor/creator-data/creator-data.js
```diff
@@ -0,0 +1,82 @@
+/*
```
leftmostcat: How does this file differ from `src/client/entity-editor/creator-data.js`? The naming is a little problematic.

src/client/entity-editor/shared-data/shared-data.js
```diff
@@ -0,0 +1,82 @@
...
+import SortNameField from './sort-name-field-container';
+
+function SharedData({
+	disambiguationVisible,
+	languageOptions
+}) {
+	return (
+		<div>
+			<h2>
+				What is the Creator called?
```
leftmostcat: Mention of creators in a generic form. Also, how do you feel about adding little hint popups with this sort of text next to a basic "Name" field?

src/server/app.js
```diff
@@ -135,7 +135,8 @@ app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
 
 const debug = require('debug')('bbsite');
 
-app.set('port', process.env.PORT || 9099);
+const DEFAULT_PORT = 9099;
```
leftmostcat: Make it a config option?

src/server/helpers/achievement.js
```diff
@@ -18,6 +18,8 @@
 
 'use strict';
 
+/* eslint prefer-spread: 1, prefer-reflect: 1, no-magic-numbers: 0 */
```
leftmostcat: Is there a particular reason these are specified here? Are they intended to be temporary pending style cleanup? If so, a `XXX` comment would be helpful.

src/server/helpers/achievement.js
```diff
@@ -52,16 +58,16 @@ function awardUnlock(UnlockType, awardAttribs) {
...
-			else {
+			if (award === null) {
 				unlockPromise = new UnlockType(awardAttribs)
 					.save(null, {method: 'insert'})
 					.then((unlock) =>
 						unlock.toJSON()
 					);
 			}
+			else {
+				unlockPromise = Promise.resolve('already unlocked');
```
leftmostcat: Why resolve with a user-invisible string?

src/server/helpers/search.js
```diff
@@ -132,6 +132,7 @@ search.indexEntity = (entity) =>
 search.refreshIndex = () =>
 	_client.indices.refresh({index: _index});
 
+/* eslint camelcase: 0, no-magic-numbers: 1 */
```
leftmostcat: Same question about eslint exceptions as above.

src/server/routes/entity/creator.js
```diff
@@ -139,16 +139,54 @@ const additionalCreatorProps = [
...
 
-router.post('/create/handler', auth.isAuthenticatedForHandler, (req, res) =>
-	entityRoutes.createEntity(
-		req, res, 'Creator', _.pick(req.body, additionalCreatorProps)
-	)
-);
 
-router.post('/:bbid/edit/handler', auth.isAuthenticatedForHandler, (req, res) =>
-	entityRoutes.editEntity(
+function transformNewForm(data) {
```
leftmostcat: Is the idea here to make this a temporary shim before we move to better data backend stuff? If so, please comment to that effect.

src/client/entity-editor/alias-editor/reducer.js
```diff
@@ -16,6 +16,10 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+import {
+	ADD_ALIAS, REMOVE_ALIAS, UPDATE_ALIAS_LANGUAGE,
+	UPDATE_ALIAS_NAME, UPDATE_ALIAS_PRIMARY, UPDATE_ALIAS_SORT_NAME
+} from './actions';
 import Immutable from 'immutable';
 
 const DEFAULT_ALIAS = Immutable.Map({
```
leftmostcat: The naming of this confused me below. Instead of `DEFAULT`, how about `EMPTY`?

src/client/entity-editor/alias-editor/alias-editor.js
```diff
@@ -16,16 +16,18 @@
...
+import AliasRowContainer from './alias-row-container';
 import React from 'react';
 import {connect} from 'react-redux';
-import {hideAliasEditor} from '../actions';
+
 
 const AliasEditor = ({
+	aliases,
 	languageOptions,
+	onAddButtonClick,
```
leftmostcat: Would it make sense to name this as `onAddAlias`? We only care about the button click internally. Externally, we only care that some action has been performed to add an alias.

LordSputnik:
The importance of keeping ItemRow and ItemRowContainer separate is to ensure there's a separation of concerns between the layout (ItemRow) and the logic and state (ItemRowContainer). It may be possible to merge the container into the editor, but I'd need to have a look. It's also useful because it may be possible to make ItemRow the same component in both the alias and identifier editors in the future.

leftmostcat:
I don't see a strong motivation for separating presentation and logic in this case, particularly given that the logic is a simple set of `mapDispatchToProps` and `mapStateToProps` call. I still think this is unnecessary indirection, and I would resist merging the presentation components of two distinct data types; it would complicate future changes. Aliases and identifiers aren't the related concepts; the sets are.

src/client/entity-editor/actions.js
```diff
@@ -0,0 +1,216 @@
...
+	return {
+		type: ADD_ALIAS,
+		id: nextAliasId++
+	};
+}
+
+export function removeAlias(index) {
+	return {
+		type: REMOVE_ALIAS,
+		index
```
LordSputnik: Nope, once an index is assigned, it stays with the associated alias.

PR #111, opened 2016-12-05
Added JSDoc annotations - GCI

Arthelon:
### Problem
Lack of documentation makes maintaining the project harder and puts a higher bar of entry for new developers.

### Solution
Adds JSDoc style annotations to a few files in the codebase, outlining the purpose, params, and return values.

### Areas of Impact
Annotations should not alter the behaviour of any existing code.

### DIFF
--- a/src/client/components/forms/publication.js
+++ b/src/client/components/forms/publication.js
@@ -16,7 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-
+/* eslint valid-jsdoc: "warn" */
 const Icon = require('react-fontawesome');
 const React = require('react');
 const request = require('superagent-bluebird-promise');
@@ -56,12 +56,10 @@ class PublicationForm extends React.Component {
 	}
 
 	/**
-	 * Sets 'tab' state variable to the first parameter and checks if
-	 * alias form is valid, storing the result as a boolean 'aliasesValid' in
-	 * state
+	 * Changes the current tab to the one specified and checks
+	 * if the alias form is valid.
 	 *
 	 * @param {number} tab - Indicates the current tab
-	 * @returns {undefined}
 	 */
 	handleTabSelect(tab) {
 		this.setState({
@@ -71,29 +69,26 @@ class PublicationForm extends React.Component {
 	}
 
 	/**
-	 * Decrements 'tab' state variable and moves to previous tab form
+	 * Event handler that moves user to previous tab
 	 *
-	 * @returns {undefined}
 	 */
 	handleBackClick() {
 		this.handleTabSelect(this.state.tab - 1);
 	}
 
 	/**
-	 * Increments 'tab' state variable and moves to following tab form
+	 * Event handler that moves user to following tab
 	 *
-	 * @returns {undefined}
 	 */
 	handleNextClick() {
 		this.handleTabSelect(this.state.tab + 1);
 	}
 
 	/**
-	 * Extracts user input from forms, formats the changes to publisher,
-	 * and submits it to the server.
+	 * Extracts user revisions to publications and sends it to the server
+	 * to be saved
 	 *
 	 * @param {object} evt - Event object passed in by event dispatcher
-	 * @returns {undefined}
 	 */
 	handleSubmit(evt) {
 		evt.preventDefault();
@@ -140,7 +135,7 @@ class PublicationForm extends React.Component {
 	/**
 	 * Renders the component: Includes a top nav bar, 3 forms that change
 	 * visibility based the 'tab' state variable, and a loading spinner that
-	 * appears when the 'waiting' state variable is true.
+	 * appears for certain async operations
 	 *
 	 * @returns {object} - JSX to render
 	 */
--- a/src/client/components/forms/publication.js
+++ b/src/client/components/forms/publication.js
@@ -16,7 +16,8 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-/* eslint valid-jsdoc: "warn" */
+/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
+
 const Icon = require('react-fontawesome');
 const React = require('react');
 const request = require('superagent-bluebird-promise');
@@ -56,10 +57,10 @@ class PublicationForm extends React.Component {
 	}
 
 	/**
-	 * Changes the current tab to the one specified and checks
-	 * if the alias form is valid.
+	 * Changes the selected tab and refreshes validity
+	 * of aliases form into state
 	 *
-	 * @param {number} tab - Indicates the current tab
+	 * @param {number} tab - Indicates the selected tab
 	 */
 	handleTabSelect(tab) {
 		this.setState({
--- a/src/client/components/forms/publication.js
+++ b/src/client/components/forms/publication.js
@@ -30,6 +30,14 @@ const PublicationData = require('./parts/publication-data');
 const RevisionNote = require('./parts/revision-note');
 
 class PublicationForm extends React.Component {
+
+	/**
+	 * Initializes component state to default values and binds class
+	 * methods to proper context so that they can be directly invoked
+	 * without explicit binding
+	 *
+	 * @param {object} props - Properties object passed down from parents
+	 */
 	constructor(props) {
 		super(props);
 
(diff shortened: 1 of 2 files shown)


src/client/components/forms/publication.js
```diff
@@ -47,21 +55,46 @@ class PublicationForm extends React.Component {
 		this.handleSubmit = this.handleSubmit.bind(this);
 	}
 
+	/**
+	 * Sets 'tab' state variable to the first parameter and checks if
+	 * alias form is valid, storing the result as a boolean 'aliasesValid' in
+	 * state
+	 *
+	 * @param {number} tab - Indicates the current tab
+	 * @returns {undefined}
```
leftmostcat: No return value, so `@returns` should be omitted.

src/client/components/forms/publication.js
```diff
@@ -47,21 +55,46 @@ class PublicationForm extends React.Component {
...
-			aliasesValid: this.aliases.valid(),
-			dataValid: this.data.valid()
+			aliasesValid: this.aliases.valid()
 		});
 	}
+
+	/**
+	 * Decrements 'tab' state variable and moves to previous tab form
+	 *
+	 * @returns {undefined}
```
leftmostcat: No return value, so `@returns` should be omitted.

src/client/components/forms/publication.js
```diff
@@ -47,21 +55,46 @@ class PublicationForm extends React.Component {
...
+	 */
 	handleBackClick() {
 		this.handleTabSelect(this.state.tab - 1);
 	}
 
+	/**
+	 * Increments 'tab' state variable and moves to following tab form
+	 *
+	 * @returns {undefined}
```
leftmostcat: No return value, so `@returns` should be omitted.

src/client/components/forms/publication.js
```diff
@@ -47,21 +55,46 @@ class PublicationForm extends React.Component {
...
 	handleNextClick() {
 		this.handleTabSelect(this.state.tab + 1);
 	}
 
+	/**
+	 * Extracts user input from forms, formats the changes to publisher,
+	 * and submits it to the server.
+	 *
+	 * @param {object} evt - Event object passed in by event dispatcher
+	 * @returns {undefined}
```
leftmostcat: No return value, so `@returns` should be omitted.

src/server/helpers/utils.js
```diff
@@ -100,6 +151,12 @@ function incrementEditorEditCountById(id, transacting) {
 		});
 }
 
+/**
+ * Removes all rows from a selection of database tables
+ * @param {object} Bookshelf - Bookshelf instance connected to database
+ * @param {string[]} tables - List of tables to truncate
+ * @returns {undefined}
```
leftmostcat: No return value, so `@returns` should be omitted.

src/client/components/forms/publication.js
```diff
@@ -16,7 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-
+/* eslint valid-jsdoc: "warn" */
```
LordSputnik: Hopefully no need for this if we set "requireReturn" to false (http://eslint.org/docs/rules/valid-jsdoc)

src/client/components/forms/publication.js
```diff
@@ -47,21 +55,41 @@ class PublicationForm extends React.Component {
 		this.handleSubmit = this.handleSubmit.bind(this);
 	}
 
+	/**
+	 * Changes the current tab to the one specified and checks
+	 * if the alias form is valid.
+	 *
+	 * @param {number} tab - Indicates the current tab
```
LordSputnik: I'd call this "the tab to be selected" - the current tab is ambiguous - it could be the one prior to selection.

src/client/components/forms/publication.js
```diff
@@ -47,21 +55,41 @@ class PublicationForm extends React.Component {
 		this.handleSubmit = this.handleSubmit.bind(this);
 	}
 
+	/**
+	 * Changes the current tab to the one specified and checks
+	 * if the alias form is valid.
```
LordSputnik: "refreshes the validity of the aliases tab in the form state" might give more context here

src/server/helpers/utils.js
```diff
@@ -16,7 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-
+/* eslint valid-jsdoc: "warn" */
```
LordSputnik: See first comment - in this case, change the rule so that returns aren't required, since that seems fairly sensible.

src/server/helpers/utils.js
```diff
@@ -27,11 +27,22 @@ const Publisher = require('bookbrainz-data').Publisher;
 const Work = require('bookbrainz-data').Work;
 const Promise = require('bluebird');
 
+/**
+ * Returns an API path for interacting with the given entity object
```
LordSputnik: Entity object -> "Bookshelf entity model" in this comment to be more specific about what's expected

src/server/helpers/utils.js
```diff
@@ -27,11 +27,22 @@ const Publisher = require('bookbrainz-data').Publisher;
...
+ * @param {object} entity - Entity object
+ * @returns {string} - URL path to interact with entity
+ */
 function getEntityLink(entity) {
 	const bbid = entity.bbid;
 	return `/${entity.type.toLowerCase()}/${bbid}`;
 }
 
+/**
+ * Returns all entity models defined with the Bookshelf ORM
```
LordSputnik: I'd prefer "within bookbrainz-data-js"

src/server/helpers/utils.js
```diff
@@ -42,6 +53,15 @@ function getEntityModels() {
 	};
 }
 
+/**
+ * Retrieves a Bookshelf ORM model given the model name
```
LordSputnik: I'd suggest "Bookshelf entity model" to make clear it only works for entities

src/server/helpers/utils.js
```diff
@@ -52,13 +72,36 @@ function getEntityModelByType(type) {
 	return entityModels[type];
 }
 
+/**
+ * Regular expression for valid BookBrainz IDs
```
LordSputnik: "BookBrainz IDs" -> "UUIDs". It might also be worth renaming this to _uuidRegex (if you feel like it!), since it isn't specific to BBIDs (could work for MBIDs too)

src/server/helpers/utils.js
```diff
@@ -52,13 +72,36 @@ function getEntityModelByType(type) {
...
+ * is passed in at a later time to be interpolated into a
+ * string
+ *
+ * @param {string[]} strings - Array of string literals from template
+ * @returns {function(*)} - Takes an object/array as an argument.
+ * When invoked, it will return a string with all the key names from
+ * the tagged template literal replaced with their corresponding values
+ * from the newly passed in object.
+ */
 // Cribbed from MDN documentation on template literals
```
LordSputnik: If we can find a link for this documentation, it'd be good to include it in the JSDoc comment - @leftmostcat I think this came from you?

src/server/helpers/utils.js
```diff
@@ -91,6 +143,13 @@ function createEntityPageTitle(entity, titleForUnnamed, templateForNamed) {
 	return title;
 }
 
+/**
+ * Adds to the edit count of the specified editor
```
LordSputnik: "Adds 1" would be more specific here

src/server/helpers/utils.js
```diff
@@ -91,6 +143,13 @@ function createEntityPageTitle(entity, titleForUnnamed, templateForNamed) {
...
+ *
+ * @param {string} id - ID of editor object/row
```
LordSputnik: how about "row ID of editor to be updated"

src/server/helpers/utils.js
```diff
@@ -91,6 +143,13 @@ function createEntityPageTitle(entity, titleForUnnamed, templateForNamed) {
...
+ * @param {object} transacting - Bookshelf transaction object
```
LordSputnik: Also would be good to note this has to be a transaction in progress

src/server/helpers/utils.js
Arthelon: BBID is being used a lot in this project so I'll avoid changing any potentially breaking identifiers for now.

Arthelon:
Thanks for these revisions. I've made changes as stated.

LordSputnik:

PR #112, opened 2016-12-07
Add JSDoc to relationship.js (GCI)

nnaik39:
### Problem

The file relationship.js needs JSDoc documentation.

### Solution

I documented relationship.js with JSDoc.

### Areas of Impact

Only the file relationship.js is affected, but all of my edits are comments so the functionality
should not be impaired.

I am new to JSDoc, so could my documentation please be checked? If any editing is needed, I would be happy to fix my mistakes. Also, if I have misunderstood any of the code or not been thorough enough, could you please tell me where I went wrong so I can improve next time?

### DIFF
--- a/src/client/components/forms/relationship.js
+++ b/src/client/components/forms/relationship.js
@@ -33,13 +33,31 @@ const validators = require('../../helpers/react-validators');
 
 const RelationshipRow = require('./parts/relationship-row');
 
+/**
+ * This function checks if the relationship is new
+ * @param: {boolean} initialType - Initial type of the relationship.
+ * @param: {boolean} initialTarget - Initial target of the relationship.
+ * Returns boolean value if initialType and initialTarget are both false
+ * If there is a new relationship, then in the render function,
+   the relationship type of the RelationshipRow
+   is set to a non-deprecated type. (const typesWithoutDeprecated)
+ */
 function isRelationshipNew(initialType, initialTarget) {
 	'use strict';
-
 	return !(initialType || initialTarget);
 }
-
+/**
+ * This class makes it possible for a user to edit a relationship.
+ * The render() function in this class generates an
+   HTML div with a RelationshipEditor.
+*/
 class RelationshipEditor extends React.Component {
+	/**
+	* This is a constructor.
+	* @param: {props} props - from React.Component constructor
+	* Returns a RelationshipEditor.
+	* @constructor
+	*/
 	constructor(props) {
 		super(props);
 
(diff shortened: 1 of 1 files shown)


src/client/components/forms/relationship.js
```diff
@@ -102,7 +129,15 @@ class RelationshipEditor extends React.Component {
 					dataHelper.getEntityLink(this.props.entity);
 			});
 	}
-
+	/**
+	* This function goes through the list of updated relationships
```
LordSputnik: Gets a list of relationships on the form from getValue(), and adds additional information derived from the state that is used internally

src/client/components/forms/relationship.js
```diff
@@ -141,7 +182,10 @@ class RelationshipEditor extends React.Component {
 			rowsSpawned
 		});
 	}
-
+	/**
+	* This function deletes all relationships which have not been selected.
```
LordSputnik: Hmm, not exactly. If you look at the code, a list of relationships is generated with unselected ones replaced with `null`. Then the `_reject` function will filter out the `null` values, leaving only the selected relationships, which are assigned to `relationshipsToDelete`. So the selected relationships are the ones being deleted. You should also mention what form event triggers this handler.

src/client/components/forms/relationship.js
```diff
@@ -153,7 +197,13 @@ class RelationshipEditor extends React.Component {
 			this.refs[idx].handleDeleteClick();
 		});
 	}
-
+	/**
+	* This function takes in a row index, checks if it is the last relationship
+	 and if it is, then a new row is added.
+	* @param: {array} updatedRelationships - relationships with correct changed and valid booleans.
+	* @param: {integer} (index) changedRowIndex - Row index which has been recently changed.
+	* Returns the total number of added rows.
```
LordSputnik: Returns the total number of rows added since the form was loaded

src/client/components/forms/relationship.js
```diff
@@ -170,7 +220,12 @@ class RelationshipEditor extends React.Component {
 
 		return rowsSpawned;
 	}
-
+	/**
+	* This takes in a row index and checks if it is new.
```
LordSputnik: if the row has been added since the form was loaded

src/client/components/forms/relationship.js
```diff
@@ -170,7 +220,12 @@ class RelationshipEditor extends React.Component {
...
+	* If it is, then it is deleted and numSelected decreases by 1.
```
LordSputnik: the number of selected rows is decremented if the row was part of the selection

src/client/components/forms/relationship.js
```diff
@@ -188,7 +243,12 @@ class RelationshipEditor extends React.Component {
 			});
 		}
 	}
-
+	/**
+	* This takes in a row index, finds out the updated relationships,
```
LordSputnik: Gets the most current information about relationships in the form, and saves it in the form state, to update the state following a change and propagate the change to the displayed UI.

src/client/components/forms/relationship.js
```diff
@@ -200,7 +260,14 @@ class RelationshipEditor extends React.Component {
 			rowsSpawned
 		});
 	}
-
+	/**
+	* This function takes in a row index, checks if it is selected,
+	and updates the state with the new number.
```
LordSputnik: new number of selected relationships.

LordSputnik:
Closed since work was resubmitted in PR #113

PR #113, opened 2016-12-09
Add edited JSDoc comments to relationship.js

nnaik39:
Thanks for the feedback on the previous documentation! Here are the edited JSDoc comments. I wasn't quite sure how to talk about how a function works in the bigger scale of the program, so I may need to go back and fix that if it's not clear enough. Let me know if there's anything which needs to be changed.
Thanks!

### DIFF
--- a/src/client/components/forms/relationship.js
+++ b/src/client/components/forms/relationship.js
@@ -16,6 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+ /* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
 
 const React = require('react');
 const request = require('superagent-bluebird-promise');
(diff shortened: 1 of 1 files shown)


LordSputnik:
Hi, it looks like your latest commit has removed most of the whitespace in the file - please check your editor settings, then fix!

Also, no need to create a separate branch and PR - if you update the original branch, the original PR will be updated to reflect that, and I can see all the changes in one place :) (much preferred!)

Thanks for the work so far.

src/client/components/forms/relationship.js
```diff
@@ -34,7 +34,8 @@ const validators = require('../../helpers/react-validators');
 const RelationshipRow = require('./parts/relationship-row');
 
 /**
- * Checks if the relationship is new by checking that the fields weren't set when the form was loaded
+ * Checks if the relationship is new by checking that
+   the fields weren't set when the form was loaded
```
LordSputnik: Missing asterisk `*` at the start here, and on the other lines you've created in this commit

src/client/components/forms/relationship.js
```diff
@@ -111,6 +115,7 @@ class RelationshipEditor extends React.Component {
 	/**
 	 * Submits the changed, valid relationships.
 	 * It is triggered by the onClick event of the SUBMIT button.
+	 * @returns {void}
```
LordSputnik: No need for this - sorry, the rules for documentation are quite new, and this is one of the things we're planning to change - for now, please add `/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */` to the top of the file (below the copyright header and license), then remove the `@returns {void}` lines from the file.

LordSputnik:

PR #114, opened 2016-12-10
Updated forms and server to support area editing

Arthelon:
### Problem
The site does not include a way to modify location-based fields.

### Solution
- Allows EntitySearch component to have its search narrowed down to a particular collection
- Builds on top of the existing schema support for areas to add additional area input components into forms that require it. Also modifying the data sent up to the server to include the results from the new inputs.
- Modifies the server-side entity edit handlers to accept the new area fields
- Adds area relations into a few indexed entity models

### Areas of Impact
This PR affects a few client-side forms, server indexing code, and entity edit handlers.

### Gotchas
- Current indexing code is buggy
- Error is thrown when edit handler of Creator entity is requested without any of its data being changed (successfully replicated on live site).
    - **UPDATE**: is this intended behaviour? https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/routes/entity/entity.js#L724

### DIFF
--- a/src/client/components/input/entity-search.js
+++ b/src/client/components/input/entity-search.js
@@ -171,9 +171,11 @@ class EntitySearch extends React.Component {
 	}
 }
 
+
 EntitySearch.displayName = 'EntitySearch';
 EntitySearch.propTypes = {
 	bsStyle: React.PropTypes.string,
+	collection: React.PropTypes.string,
 	defaultValue: React.PropTypes.shape({
 		bbid: React.PropTypes.string
 	}),
--- a/src/client/components/input/entity-search.js
+++ b/src/client/components/input/entity-search.js
@@ -171,7 +171,6 @@ class EntitySearch extends React.Component {
 	}
 }
 
-
 EntitySearch.displayName = 'EntitySearch';
 EntitySearch.propTypes = {
 	bsStyle: React.PropTypes.string,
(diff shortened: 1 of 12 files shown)


src/client/components/input/entity-search.js
```diff
@@ -171,9 +171,11 @@ class EntitySearch extends React.Component {
 	}
 }
 
+
```
leftmostcat: #nit: Extraneous whitespace.

src/server/helpers/search.js
```diff
@@ -129,6 +130,7 @@ search.indexEntity = (entity) =>
 		body: entity
 	});
 
+
```
leftmostcat: #nit: Extraneous whitespace.

src/server/helpers/search.js
```diff
@@ -207,6 +209,7 @@ search.generateIndex = () => {
 				throw err;
 			}
 		})
+		// GOTCHA: index creation is buggy
```
leftmostcat: Could you create a ticket in JIRA and add a link to it here?

src/server/helpers/search.js
```diff
@@ -26,6 +26,7 @@ const _ = require('lodash');
 const Publication = require('bookbrainz-data').Publication;
 const Creator = require('bookbrainz-data').Creator;
 const Edition = require('bookbrainz-data').Edition;
+const Editor = require('bookbrainz-data').Editor;
```
leftmostcat: Is this import unused?

PR #115, opened 2016-12-10
JSDoc additions - GCI

Arthelon:
### Problem
Lack of documentation makes maintaining the project harder and puts a higher bar of entry for new developers.

### Solution
Added JSDoc style annotations to a few server and client files.

### Areas of Impact
Annotations should not alter the behaviour of any existing code.

### DIFF
--- a/src/client/components/input/select2.js
+++ b/src/client/components/input/select2.js
@@ -116,7 +116,7 @@ class Select extends React.Component {
 	}
 
 	/**
-	 * Prevents firing of input change handlers.
+	 * Removes the input change handlers.
 	 */
 	disableOnChange() {
 		const select = $(this.target.getInputDOMNode());
@@ -127,7 +127,8 @@ class Select extends React.Component {
 	}
 
 	/**
-	 * Used by React to render the component.
+	 * Used by React to render the component: A select input that allows users
+	 * to select from a list of supplied options.
 	 *
 	 * @returns {object} - JSX to render.
 	 */
--- a/src/client/components/input/select2.js
+++ b/src/client/components/input/select2.js
@@ -16,6 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
 
 const React = require('react');
 const _assign = require('lodash.assign');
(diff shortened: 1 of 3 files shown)


src/client/components/input/select2.js
```diff
@@ -31,10 +32,21 @@ if (typeof window !== 'undefined') {
 }
 
 class Select extends React.Component {
+
```
leftmostcat: #nit: Extraneous whitespace. Generally, I'd start the block off with a line containing something. Not hugely important, though.

src/client/components/input/select2.js
```diff
@@ -31,10 +32,21 @@ if (typeof window !== 'undefined') {
...
+	/**
+	 * Invoked after component mount that initializes select2
```
leftmostcat: I'd suggest avoiding comments that describe what is calling a function; that's liable to change. Instead, describe what the function does.

src/server/routes/search.js
```diff
@@ -57,6 +61,10 @@ router.get('/', (req, res, next) => {
 		.catch(next);
 });
 
+/**
+ * Returns autocomplete results for a given user query. Can be further
```
leftmostcat: Given that you end one sentence with a period here, I'd recommend using them consistently at the end of all sentences. Consistent and will probably read better in generated documentation.

src/server/routes/search.js
```diff
@@ -66,6 +74,11 @@ router.get('/autocomplete', (req, res) => {
 	handler.sendPromiseResult(res, searchPromise);
 });
 
+/**
+ * Regenerates search index. Only accessible to authenticated users
```
leftmostcat: Instead of "only accessible to authenticated users", maybe "restricted to administrators"? Not just any authed user can do it.

src/client/components/input/select2.js
Arthelon: +1

src/client/components/input/select2.js
```diff
@@ -87,6 +115,9 @@ class Select extends React.Component {
 		mountElement.on('change', this.props.onChange);
 	}
 
+	/**
+	 * Prevents firing of input change handlers.
```
LordSputnik: I'd say this "Removes the input change handlers."

src/client/components/input/select2.js
```diff
@@ -95,6 +126,11 @@ class Select extends React.Component {
 		select.off('change');
 	}
 
+	/**
+	 * Used by React to render the component.
```
LordSputnik: Describe the component a bit here - what does it look like, and what does it do? (what Leftmost said - why/where would I use the component?)

src/client/components/pages/search.js
```diff
@@ -35,6 +44,11 @@ class SearchPage extends React.Component {
 		this.handleSearch = this.handleSearch.bind(this);
 	}
 
+	/**
+	 * Gets search results from user query.
```
LordSputnik: Search results from where? In what format? What does it do with them?

src/server/routes/search.js
```diff
@@ -37,6 +37,10 @@ const SearchPage = React.createFactory(
 
 const router = express.Router();
 
+/**
+ * Returns React markup for the search page that is rendered by the user's
```
LordSputnik: Don't like the use of "Returns" here - "Generates" would be better to avoid confusion (it doesn't actually *return* anything!)

src/server/routes/search.js
```diff
@@ -57,6 +61,10 @@ router.get('/', (req, res, next) => {
 		.catch(next);
 });
 
+/**
+ * Returns autocomplete results for a given user query. Can be further filtered
```
LordSputnik: Same thing with "Returns" here - maybe "Responds with"?

LordSputnik:
It looks like Leftmost's review comments have been addressed, so I'm going to merge this.

PR #116, opened 2016-12-11
area-fixes-Daniel-Hsing

Arthelon:
### Problem
Area entities are not being indexed; hence EntitySearch components included in #114 have no way of providing any area options to user. The Area entity is different from the other indexed entities so some special cases need to be made. There are also some issues with data retrieval and insertion in the EntitySearch component.

### Solution
Indexes the Area entity, patches data handling issues in EntitySearch, and adds special cases in search for dealing with Area entities. Also introduces a few styling touches such as disabling area links in the SearchPage results and a globe icon for area entities.

### Areas of Impact
Impacts the indexing and some search functions, Creator, Publisher and Profile edit forms, SearchResults component.

### DIFF
--- a/src/client/components/pages/parts/search-results.js
+++ b/src/client/components/pages/parts/search-results.js
@@ -33,20 +33,26 @@ function SearchResults(props) {
 		);
 	}
 
-	const results = props.results.map((result) => (
-		<tr key={result.bbid}>
-			<td>
-				<a href={`/${result.type.toLowerCase()}/${result.bbid}`}>
-					{result.defaultAlias ?
-						result.defaultAlias.name : '(unnamed)'
-					}
-				</a>
-			</td>
-			<td>
-				{result.type}
-			</td>
-		</tr>
-	));
+	const results = props.results.map((result) => {
+		// No redirect link for Area entity results
+		const name = result.defaultAlias ? result.defaultAlias.name :
+			'(unnamed)';
+		const alias = result.type === 'Area' ? name :
+			<a href={`/${result.type.toLowerCase()}/${result.bbid}`}>
+				{name}
+			</a>;
+
+		return (
+			<tr key={result.bbid}>
+				<td>
+					{alias}
+				</td>
+				<td>
+					{result.type}
+				</td>
+			</tr>
+		);
+	});
 
 	return (
 		<Table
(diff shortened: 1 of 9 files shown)


src/server/helpers/search.js
```diff
@@ -71,11 +72,26 @@ function _fetchEntityModelsForESResults(results) {
 
 	return Promise.map(results.hits, (hit) => {
 		const entityStub = hit._source;
-		const model = utils.getEntityModelByType(entityStub.type);
 
+		if (entityStub.type === 'Area') {
+			return Area.forge({bbid: entityStub.bbid})
```
LordSputnik: No need for the where clause here - just move the `{gid: entityStub.bbid}` into the forge function call instead of `{bbid: entityStub.bbid}`.

src/server/helpers/search.js
```diff
@@ -71,11 +72,26 @@ function _fetchEntityModelsForESResults(results) {
...
+					};
+					areaJSON.type = 'Area';
+					return areaJSON;
+				});
+		}
+		const model = utils.getEntityModelByType(entityStub.type);
 		return model.forge({bbid: entityStub.bbid})
 			.fetch({withRelated: ['defaultAlias']})
 			.then((entity) => entity.toJSON());
+
```
LordSputnik: Extra whitespace

src/server/helpers/search.js
```diff
@@ -129,6 +159,7 @@ search.indexEntity = (entity) =>
 		body: entity
 	});
 
+
```
LordSputnik: Extra whitespace

src/server/helpers/search.js
```diff
@@ -246,6 +277,14 @@ search.generateIndex = () => {
 					.map(search.indexEntity)
 			);
 		})
+		.then(() =>
+			 Promise.all(Area.forge()
```
LordSputnik: No need to surround this with `Promise.all`, I think - the map function should return a promise which resolves once all the elements in the collection have been processed.

src/client/components/pages/parts/search-results.js
```diff
@@ -33,20 +33,26 @@ function SearchResults(props) {
...
-			<td>
-				{result.type}
-			</td>
-		</tr>
-	));
+	const results = props.results.map((result) => {
+		// No redirect link for Area entity results
+		const name = result.defaultAlias ? result.defaultAlias.name :
+			'(unnamed)';
+		const alias = result.type === 'Area' ? name :
```
LordSputnik: Even better here - link to the area page on MusicBrainz - https://musicbrainz.org/area/<area.gid>

PR #117, opened 2016-12-14
Editor tab conversion - GCI

Arthelon:
### Problem
2 of the 3 editor tabs are currently being rendered server side using pug templates, whereas the last is being rendered on a different pug template. This should not be the case, given the similarities between all 3 pages.

### Solution
Rewrote the 2 pug page templates into React components. Utilized the existing Layout component to create a 'EditorContainer' component for the tabs to be nested under. Modified routes to render all 3 tab pages onto the same 'target.jade' template. 

### Areas of Impact
A lot of patching and refactoring was done in this PR. Areas affected include the build-script, the editors server route, the old achievements component (moved to folder with other tabs), template files, and some other files in the client folder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,10 +23,8 @@
   },
   "dependencies": {
     "babel-cli": "^6.18.0",
-    "babel-plugin-array-includes": "^2.0.3",
     "babel-plugin-transform-es2015-modules-commonjs": "^6.18.0",
     "babel-preset-es2015-node": "^6.1.0",
-    "babel-preset-react": "^6.11.1",
     "babelify": "^7.3.0",
     "bluebird": "^3.3.4",
     "body-parser": "^1.14.1",
--- a/src/client/components/layout.js
+++ b/src/client/components/layout.js
@@ -30,7 +30,6 @@ const injectChildElemsWithProps =
 const Nav = bootstrap.Nav;
 const Navbar = bootstrap.Navbar;
 const MenuItem = bootstrap.MenuItem;
-const FormGroup = bootstrap.FormGroup;
 
 const Footer = require('./footer');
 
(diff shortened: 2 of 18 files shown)


src/client/components/layout.js
```diff
@@ -145,25 +144,25 @@ class Layout extends React.Component {
 				}
 				{!(homepage || hideSearch) &&
 				<Navbar.Form pullRight
-					action="/search"
-					role="search"
+					 action="/search"
+					 role="search"
 				>
-					<FormGroup>
+					<div className="form-group">
```
LordSputnik: Could you explain a little bit more about this? Add a comment explaining why it's not good to use FormGroup here, so that we'll know why in a few years when we come back to this code.

src/server/routes/editor.js
```diff
@@ -325,18 +348,18 @@ router.get('/:id/achievements', (req, res, next) => {
...
-				props: {
-					editor: editorJSON,
-					achievement: achievementJSON
-				},
-				markup
+			const props = Object.assign({}, req.app.locals, res.locals, {
+				editor: editorJSON,
+				achievement: achievementJSON
+			});
+			res.render('target', {
```
LordSputnik: Hmm, generally I think I prefer having a `markup` variable holding the server-rendered markup, because it makes it clearer what's happening in the `res.render` function - e.g.:

        res.render('target', {markup, props})

package.json
```diff
@@ -23,10 +23,8 @@
   },
   "dependencies": {
     "babel-cli": "^6.18.0",
-    "babel-plugin-array-includes": "^2.0.3",
```
LordSputnik: Hmm, this doesn't matter for this PR, but it would have been better to remove the devDependencies here. The rule I use is - if it's needed to run the site, put it in dependencies, and if it's just needed to debug or test the site, put it in devDependencies. This is a new rule for me - I've only just started using it recently so not all the dependencies currently follow it (which is why we had some duplicates).

In this case, `babel-plugin-array-includes` and `babel-preset-react` are used by babel when compiling the site code when `npm run start` is called, so they aren't just used for debugging/testing.

This is just so that you know in future - please don't change this here - I'll do a cleanup of dependencies in the master branch after the PR is merged.

src/client/components/pages/parts/editor-achievements.js
```diff
@@ -141,8 +142,14 @@ class EditorAchievementTab extends React.Component {
 
 EditorAchievementTab.displayName = 'EditorAchievementTab';
 EditorAchievementTab.propTypes = {
-	achievement: React.PropTypes.object,
-	editor: React.PropTypes.object
+	achievement: React.PropTypes.shape({
+		model: React.PropTypes.array
+
```
LordSputnik: Stray whitespace here!

LordSputnik:
Also, I should've asked you this before, but could you also confirm that you license all the code you write for BookBrainz under the GPL-2.0+, so that we can use it safely in our project? (Just a yes here will do)

src/client/components/layout.js
Arthelon: Should've added a comment here. The usage of the react-bootstrap FormGroup component here was causing a DOM mutation.

src/server/routes/editor.js
Arthelon: +1

package.json
Arthelon: gotcha

Arthelon:
@LordSputnik Of course. I'll also update the license headers in files

LordSputnik:
One thing for the future: try to keep the first line of the commit message to about 80 characters or less, and then add more detail on subsequent lines if you need to (can use git gui or just don't close the quotation on the same line).

PR #118, opened 2016-12-14
Added area information to UI

Arthelon:
### Problem
Area information for Creator and Publisher entities and the editor are not being displayed on the site.

### Solution
Adds area attributes to entity pages and profile pages of editors. Also updated server + model code in https://github.com/bookbrainz/bookbrainz-data-js to include area changes in entity revisions.

### Areas of Impact
Minimal impact on Creator, Publisher, and Editor templates. Adds new cases to server-side revision code that is used to format changes for the client.

### DIFF
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -69,6 +69,16 @@ function formatCreatorChange(change) {
 		return baseFormatter.formatTypeChange(change, 'Creator Type');
 	}
 
+	if (_.isEqual(change.path, ['beginArea']) ||
+			_.isEqual(change.path, ['beginArea', 'name'])) {
+		return baseFormatter.formatAreaChange(change, 'Begin Area');
+	}
+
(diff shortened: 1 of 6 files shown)


src/server/routes/revision.js
```diff
@@ -69,6 +69,16 @@ function formatCreatorChange(change) {
 		return baseFormatter.formatTypeChange(change, 'Creator Type');
 	}
 
+	if (_.isEqual(change.path, ['beginArea']) ||
+			_.isEqual(change.path, ['beginArea', 'name'])) {
+		return baseFormatter.formatAreaChange(change, 'Begin Area');
+	}
+
+	if (_.isEqual(change.path, ['endArea'] ||
+			_.isEqual(change.path, ['beginArea']))) {
```
LordSputnik: Two issues here:
1. path in second line is wrong
2. parenthesis placement incorrect

PR #119, opened 2016-12-14
Merged near-identical template files

BayMinimum:
### Problem

Many template files have identities, and duplication is bad.

### Solution

Merged similar template files.

### Areas of Impact

Routing js files are also modified, so it will he safe.


templates/editor/editor-common.jade
```diff
@@ -0,0 +1,100 @@
+extends ../layout
```
LordSputnik: Please revert the merging of the editor template file. This is currently being dealt with in its own GCI task, since there is a lot less similarity here than in the other merged files.

templates/common.jade
```diff
@@ -0,0 +1,19 @@
+extends /templates/layout
+
+block body
+    if task == "edit"
+        .container#content
+         #relationshipForm
```
LordSputnik: This line needs indenting once more, so that the `relationshipForm` div is rendered inside the `content` div.

templates/entity/create/create-common.jade
```diff
@@ -0,0 +1,24 @@
...
+
+block body
+    .container#content
+        .row
+            h1=heading
+                small=' ' + subheading
+        .row
+            noscript
+                .alert.alert-danger(role='alert') This page will not function correctly without JavaScript! Please enable JavaScript to use this page.
+        if script == 'creator'
```
LordSputnik: Alternatively, to get rid of these if/else statements, change the ID of these form elements to "#entityForm", then update the corresponding `controllers/entity/` scripts to use this element.

LordSputnik:
Could you also confirm that you license all the code you write for BookBrainz under the GPL-2.0+, so that we can use it safely in our project? (Just a yes here will do)

BayMinimum:
Yes, I confirm that I license all the code I write for BookBrainz under the GPL-2.0+.

BayMinimum:
And fixed everything requested:D

src/server/routes/entity/edition.js
```diff
@@ -136,6 +136,7 @@ router.get('/create', auth.isAuthenticated, loadIdentifierTypes,
 				title: 'Add Edition',
 				heading: 'Create Edition',
 				subheading: 'Add a new Edition to BookBrainz',
+				script: 'edition',
```
LordSputnik: Also need to update the template name from `entity/create/<entity>` to `entity/create/create-common` for all entity create routes.

src/server/routes/entity/edition.js
```diff
@@ -166,6 +167,7 @@ router.get('/:bbid/edit', auth.isAuthenticated, loadIdentifierTypes,
 			title: 'Edit Edition',
 			heading: 'Edit Edition',
 			subheading: 'Edit an existing Edition in BookBrainz',
+			script: 'edition',
```
LordSputnik: Also need to update the template name from `entity/create/<entity>` to `entity/create/create-common` for all entity edit routes.

templates/common.jade
```diff
@@ -0,0 +1,19 @@
+extends /templates/layout
+
+block body
+    if task == "edit"
+        .container#content
+            #relationshipForm
+    else if task == "delete"
+        h1 Delete Entity
```
LordSputnik: Missing the container here:

    .row.margin-top-2
        #deletion-form!= markup

templates/common.jade
```diff
@@ -0,0 +1,19 @@
+extends /templates/layout
```
LordSputnik: Use a relative path here - absolute doesn't work:

    extends layout

BayMinimum:
Oops...fixed it!:D

PR #120, opened 2016-12-18
Entity revisions conversion - GCI

Arthelon:
### Problem
Entity revision display is being rendered with pug instead of React.

### Solution
Created a React component for displaying entity revisions. Updated server code to render the newly created component instead of the old template. Also included 'revision.notes' in the relations being fetched to the client.

### Areas of Impact
Updates the displayRevisions render function and adds a new component to the client folder.

### DIFF
--- a/src/client/components/pages/entity-revisions.js
+++ b/src/client/components/pages/entity-revisions.js
@@ -78,7 +78,11 @@ class EntityRevisions extends React.Component {
 							key={`${revision.revision.author.id}${revision.id}`}
 						>
 							{header}
-							{revision.revision.note}
+							{revision.revision.notes.length > 0 &&
+								<p className="list-group-item-text">
+									{revision.revision.notes[0].content}
+								</p>
+							}
 						</ListGroupItem>
 					);
 				})}
--- a/src/client/components/pages/entity-revisions.js
+++ b/src/client/components/pages/entity-revisions.js
@@ -29,7 +29,7 @@ class EntityRevisions extends React.Component {
 	constructor(props) {
 		super(props);
 		this.renderHeader = this.renderHeader.bind(this);
-		this.renderRevisions = this.renderRevisions.bind(this);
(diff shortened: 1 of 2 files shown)


src/client/components/pages/entity-revisions.js
```diff
@@ -0,0 +1,112 @@
...
+							<small className="pull-right">
+							{`${revision.revision.author.name}, ${dateLabel}`}
+							</small>
+							{`r${revision.id}`}
+						</h4>
+					);
+					return (
+						<ListGroupItem
+							href={`/revision/${revision.id}`}
+							key={`${revision.revision.author.id}${revision.id}`}
```
Arthelon: Using the index as the key might be too primitive. Don't want to run the risk of messing with React's reconciliation algo

src/client/components/pages/entity-revisions.js
```diff
@@ -0,0 +1,112 @@
...
+					<hr/>
+				</Col>
+			</Row>
+		);
+	}
+
+	renderRevisions() {
+		const {revisions} = this.props;
+
+		return (
```
Arthelon: sure, that sounds good.

src/client/components/pages/entity-revisions.js
```diff
@@ -0,0 +1,109 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+const React = require('React');
```
LordSputnik: Module name is lowercase - `require('react')`

Arthelon:
+1

PR #121, opened 2016-12-20
Revision notes input - GCI

Arthelon:
### Problem
Editors are currently unable to make additional comments on an existing revision.

### Solution
Adds new input on the revision page for editors to add notes. New server endpoint was also added to receive the note data and create new Note records.

### Areas of Impact
Rendering code for the original RevisionPage component was altered, new helper function added into the entity routes module, new endpoint added to revisions route module.


src/server/routes/revision.js
```diff
@@ -237,14 +238,25 @@ router.get('/:id', (req, res, next) => {
...
-			const props = {revision: revision.toJSON(), diffs};
-			res.render('page', {
+			const props = Object.assign({}, req.app.locals, res.locals, {
+				revision: revision.toJSON(),
 				title: 'RevisionPage',
-				markup: ReactDOMServer.renderToString(RevisionPage(props))
+				diffs
 			});
+			const markup = ReactDOMServer.renderToString(
+				<Layout {...props}>
```
leftmostcat: There's some duplication here of what the controller is doing. Is there any way around that? It'd be nice if the controller could take the appropriate props and either render to string or into the DOM, but I'm not sure if that's feasible. @Arthelon, @LordSputnik, feedback appreciated.

Arthelon: Not to my knowledge. The client side script is necessary for the React hooks and the execution of other Javascript.

Arthelon: hmm I could try writing a helper func that can be imported by both server and client. The overall benefit could be quite marginal though.

leftmostcat: May be something for future work. No change needed in this PR.

src/client/components/pages/revision.js
```diff
@@ -105,9 +112,29 @@ class RevisionPage extends React.Component {
...
+	handleSubmit(event) {
+		event.preventDefault();
+		const data = {
+			note: this.noteInput.getValue()
+		};
+		console.log(data);
+		request.post(`/revision/${this.props.revision.id}/note`)
+			.send(data).promise()
+			.then(() => {
+				location.reload();
```
leftmostcat: Is this `reload()` copied from other components? I'd rather we not have to leave the page and instead just update inline, but I'm guessing other components just reload, so I won't ask that this change here. (This is more a note to myself.)

src/client/components/pages/revision.js
```diff
@@ -166,24 +193,47 @@ class RevisionPage extends React.Component {
...
+
+					<h3>Revision Notes</h3>
+					{revisionNotes}
+					{user &&
+						<form
+							className="margin-top-2"
+							onSubmit={this.handleSubmit}
+						>
+							<Input
+								label="Additional Revision Notes"
```
leftmostcat: I'd call this something like "Add Note", which corresponds to the user interacting with the page, rather than just a display of further information.

src/server/routes/entity/entity.js
```diff
@@ -166,9 +166,11 @@ module.exports.displayRevisions = (req, res, next, RevisionModel) => {
 
 function _createNote(content, editor, revision, transacting) {
 	if (content) {
+		const revisionId = typeof revision === 'string' ?
```
leftmostcat: I'd rather we not overload the meaning of this param. It should be possible to do something like the following in `addNoteToRevision()` below:
```javascript
const revision = Revision.forge({id: req.params.id});
_createNote(req.body.note, editorJSON, revision, transacting);
```
Generally, having to check the type of an argument means we're doing something wrong.

src/server/routes/revision.js
```diff
@@ -237,14 +238,25 @@ router.get('/:id', (req, res, next) => {
...
+					<RevisionPage />
```
leftmostcat: Forgive me if I missed earlier discussion on this, but is there a reason `diffs` and `revision` aren't set on `RevisionPage` here (as opposed to the controller)?

src/client/components/pages/revision.js
Arthelon: I'm pretty sure the other components have reloads. This preferably should be done with an XHR though since the user would be guaranteed to have JS enabled if they made a comment. Consider making this a task on GCI? :)

src/server/routes/revision.js
Arthelon: No reason. It'll work the way it is but those fields should actually be set on RevisionPage to make things more explicit. Thanks for the catch.

src/client/components/pages/revision.js
```diff
@@ -19,11 +19,18 @@
  */
 
 const React = require('react');
+const bootstrap = require('react-bootstrap');
```
Arthelon: +1

src/client/components/pages/revision.js
```diff
@@ -105,9 +112,29 @@ class RevisionPage extends React.Component {
...
+		super(props);
+		this.handleSubmit = this.handleSubmit.bind(this);
+	}
+
+	handleSubmit(event) {
+		event.preventDefault();
+		const data = {
+			note: this.noteInput.getValue()
+		};
+		console.log(data);
```
Arthelon: yeah, that shouldn't have been committed. mb

src/server/routes/revision.js
Arthelon: As you say, putting the props directly onto the RevisionPage may be more explicit but separating the props slightly overcomplicates the injection of props into the client code (it's probably faster because of this too!). Do you think it's worth making that change here?

leftmostcat: I think it's worth making the change. As I understand it, the difference in the props between this rendering and the rendering in the controller will make react unable to correctly link the two and lose us most of the benefit of isomorphic react, as it will have to re-render the page from scratch.

Arthelon: The props in the controller should be the same though?

LordSputnik: Yeah this is fine since the Layout props will propagate down to the RevisionPage due to the way the Layout component handles its children. I agree we should come back to that though (as mentioned in IRC) - if we can give those props to the child directly we should do.

Arthelon:
Made a few changes towards improving the passing of props. Wasn't able to test it fully (still getting that user error) so do scrutinize the code. In the code I'm keeping all the props in a centralized object for clarity and also for increased ease in passing props down to client controllers.

Arthelon:
Next step: Write a helper function that checks the module type and calls the appropriate render function based on the environment to help cut away JSX duplication (client and server JSX is identical atm).

leftmostcat:
This pattern is looking a lot better. Thanks for your hard work. I'll do some testing and merge this if it's all working.

leftmostcat:
In testing, I've found a couple issues:
- `user` prop is unset on `RevisionPage`; ensure it's passed explicitly to any pages that need it
  - Additionally, `user` isn't defined under `propTypes` for `RevisionPage`
- Submitted revision notes are not added more often than not; inconsistent, no errors on either console

leftmostcat:
Revision submission issue is not solved by the intervening commits; this is now an open issue in master.

LordSputnik:
What issue is that? I tested note submission and the three I tried adding seemed to work.

Arthelon:
Wasn't able to replicate this issue on my laptop either.

LordSputnik:
I can't see anything in the code that might result in intermittent submission. All of the stuff with promises seems to be done correctly and the client request code is good. @Leftmost perhaps it's an environment issue? We could do some further testing on test.bb.org.

PR #122, opened 2016-12-21
Introduce es2017 async/await

leftmostcat:
### Problem

Promise chains can become awkward and introduce difficulties in both structuring and following code.

### Solution

Make use of es2017 async/await syntax to hide the grittier parts of Promise syntax.

### Areas of Impact

Main page, user auth, babel compilation.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -81,7 +81,6 @@ rules:
     - 2
     - allow-null
   no-alert: 2
-  guard-for-in: 2
   no-caller: 2
   no-div-regex: 2
   no-else-return: 2
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 ---
 parserOptions:
-  ecmaVersion: 7
+  ecmaVersion: 8
   ecmaFeatures:
     jsx: true,
     modules: false
@@ -117,7 +117,6 @@ rules:
   no-self-compare: 2
   no-sequences: 2
   no-throw-literal: 2
-  no-unused-expressions: 2
   no-unused-labels: 2
   no-useless-call: 2
   no-useless-concat: 2
@@ -275,7 +274,6 @@ rules:
   wrap-regex: 2
 
   # ECMAScript 6
-  arrow-parens: 2
   arrow-spacing: 2
   constructor-super: 2
   generator-star-spacing:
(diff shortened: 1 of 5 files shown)


.eslintrc
```diff
@@ -81,7 +81,6 @@ rules:
     - 2
     - allow-null
   no-alert: 2
-  guard-for-in: 2
```
leftmostcat: Removed for convenience; I'd be willing to put this back, but it creates unnecessary checks when we know the contents of the object we're being passed.

.eslintrc
```diff
@@ -117,7 +116,6 @@ rules:
   no-self-compare: 2
   no-sequences: 2
   no-throw-literal: 2
-  no-unused-expressions: 2
```
leftmostcat: Gives us back an error when we have a bare `await foo()`, even though this is actually a function call.

.eslintrc
```diff
@@ -275,7 +273,6 @@ rules:
   wrap-regex: 2
 
   # ECMAScript 6
-  arrow-parens: 2
```
leftmostcat: Gives false positives for async arrow functions. See https://github.com/eslint/eslint/issues/3354

src/server/helpers/auth.js
```diff
@@ -65,22 +62,25 @@ auth.init = (app) => {
...
+					const linkedUser = await _linkMBAccount(req.user, profile);
+
+					// Logged in, associate
+					return done(null, linkedUser.toJSON());
+				}
+
+				// Not logged in, authenticate
+				const fetchedUser = await _getAccountByMBUserId(profile);
+
+				await _updateCachedMBName(fetchedUser, profile);
```
leftmostcat: Unsure here; does the save mutate `fetchedUser` in any way, or are we fine holding onto the previous reference?

src/server/helpers/auth.js
```diff
@@ -107,15 +107,11 @@ auth.isAuthenticated = (req, res, next) => {
 };
 
 auth.isAuthenticatedForHandler = (req, res, next) => {
-	new Promise((resolve) => {
-		if (req.isAuthenticated()) {
-			resolve();
-		}
+	if (req.isAuthenticated()) {
```
leftmostcat: Not really async/await related, but I'm mad at myself for how I wrote this before.

PR #123, opened 2016-12-26
Convert search to async/await and fix indexing

leftmostcat:
### Problem

Search index generation is broken.

### Solution

Don't try to delete an index that doesn't exist and catch the error; error message is not stable across versions.

### Areas of Impact

Search.


package.json
```diff
@@ -35,7 +35,7 @@
     "catw": "^1.0.1",
     "connect-redis": "^3.0.1",
     "debug": "^2.2.0",
-    "elasticsearch": "^10.1.3",
+    "elasticsearch": "^12.1.3",
```
leftmostcat: Was unable to get search working with ES 5.x and older versions of the JS library.

Arthelon:
+1 The version bump fixed indexing for me as well. I noticed that Promise rejections are not being handled properly in generateIndex. It might be good to add a few try/catch blocks in the async function or to add a root level .catch handler where generateIndex is being invoked.

src/server/helpers/search.js
```diff
@@ -224,65 +223,71 @@ search.generateIndex = () => {
...
+
+	const areaCollection = await Area.forge()
+		// countries only
+		.where({type: 1})
+		.fetchAll();
+
+	const areas = areaCollection.toJSON();
+
+	for (const area of areas) {
+		await search.indexArea(area);
```
LordSputnik: Won't this request that each area is indexed *in sequence*, rather than in parallel, which is the existing behaviour?

LordSputnik: I think we need to do this here:

    const areaCollection = await Area.forge()
        .where({type: 1})
        .fetchAll();
    
    const areas = areaCollections.toJSON();
    const indexedAreas = areas.map((area) => search.indexArea(area));
    await Promise.all(indexedAreas);

See https://pouchdb.com/2015/03/05/taming-the-async-beast-with-es7.html

src/server/helpers/search.js
```diff
@@ -224,65 +223,71 @@ search.generateIndex = () => {
...
+				withRelated: baseRelations.concat(behavior.relations)
+			});
+
+		const entities = collection.toJSON();
+
+		for (const entity of entities) {
+			await search.indexEntity(entity);
+		}
+	}
+
```
Arthelon: Same issue that @LordSputnik brought up.

Here's an idea for a solution. It still blocks the subsequent area indexing though.
```
const behaviorPromise = entityBehaviors.map((behavior) =>
    behavior.model.forge().fetchAll({
        withRelated: baseRelations.concat(behavior.relations)
    })
);
const entityList = await Promise.all(behaviorPromise);
const indexedEntities = entityList.map((entities) =>
    Promise.all(entities.toJSON().map((entityObj) => 
        search.indexEntity(entityObj)
    ))
)
await Promise.all(indexedEntities);
    
```

LordSputnik: I think that's OK - the previous code blocked area indexing too.

PR #124, opened 2016-12-29
Add JSdocs comments for Creator form methods

dpmittal:
### Problem

### Solution

### Areas of Impact

### DIFF
--- a/src/client/components/forms/creator.js
+++ b/src/client/components/forms/creator.js
@@ -16,6 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
 
 const Icon = require('react-fontawesome');
 const React = require('react');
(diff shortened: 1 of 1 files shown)


src/client/components/forms/creator.js
```diff
@@ -55,14 +70,25 @@ class CreatorForm extends React.Component {
...
+	/**
+	* Handler which directs the viewer to the next tab
+	*/
 	handleNextClick() {
 		this.handleTabSelect(this.state.tab + 1);
 	}
 
+	/**
+	* Handler which collects the creator information edited by the author
+	*	and sends it to the database to be revised.
```
LordSputnik: "sends it to the server to be processed" would be better here.

src/client/components/forms/creator.js
```diff
@@ -29,7 +30,16 @@ const CreatorData = require('./parts/creator-data');
...
 
+/**
+* This is an extended class of the React Component to
+* create and edit a creator
+*/
 class CreatorForm extends React.Component {
+	/**
+	* @constructor
+	* @param {object} props - Object of properties passed in the constructor.
+	* Binds the class methods to their respective data
```
LordSputnik: Please put this description above the `@constructor` and `@param` lines

src/client/components/forms/creator.js
```diff
@@ -111,6 +137,12 @@ class CreatorForm extends React.Component {
 			});
 	}
 
+	/**
+	* Renders the document and allows the author to submit the data
+	* Includes a loading spinner if there is delay in data display due to
+	* connectivity.
+	* @return {HTML} a HTML document which displays a creatoreditor
```
LordSputnik: Actually returns a React Element - let's use the `ReactElement` type to indicate this from now on.

dpmittal:
@LordSputnik I have made the changes :D

PR #125, opened 2016-12-31
Profile editor extension - GCI

Arthelon:
### Problem
Current profile editor form does not expose all user-related field, i.e. gender, birth date and display name.

### Solution
Adds new form inputs for the above three fields and updates server handler for profile edits. Also moved most of the rendering to React.

### Areas of Impact
Code in the editor route used for rendering profile form and receiving profile edits. Updated the profile react component. Updated the client side controller to use the Layout component.

GOTCHAS:
- Title field in the original profile editor form did not include a default title value, so this was left out until further clarification.
- Date field is always decremented by 1 day server side. This behaviour seems like it existed in master before this change though.

### DIFF
--- a/src/client/components/forms/profile.js
+++ b/src/client/components/forms/profile.js
@@ -27,6 +27,8 @@
 const React = require('react');
 const request = require('superagent-bluebird-promise');
 
+const Grid = require('react-bootstrap').Grid;
+const Row = require('react-bootstrap').Row;
 const Button = require('react-bootstrap').Button;
 const Input = require('react-bootstrap').Input;
 const Col = require('react-bootstrap').Col;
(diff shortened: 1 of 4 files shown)


src/client/components/pages/parts/editor-profile.js
```diff
@@ -39,9 +39,14 @@ class EditorProfileTab extends React.Component {
...
 		const {user, editor} = this.props;
-		const {cachedMetabrainzName, metabrainzUserId} = editor;
+		const {cachedMetabrainzName,
+			metabrainzUserId,
+			name,
+			gender,
+			birthDate} = editor;
 		const createdAtDate = (new Date(editor.createdAt)).toUTCString();
 		const lastActiveDate = (new Date(editor.activeAt)).toUTCString();
+		const birthday = (new Date(birthDate)).toUTCString();
```
LordSputnik: We should format this as YYYY-MM-DD - but since date/time display needs an overhaul across the site, you don't have to worry about this. If you want to anyway, please add Moment.js as a dependency then require that and use it to format.

Arthelon:
@LordSputnik With regards to the titles, I just wanted to know if there was a particular reason the default value was being excluded from the original profile editor form.

Arthelon:
Also horizontally aligned all form inputs along the same axis. It looked a bit weird having them all centered.

src/client/components/forms/profile.js
```diff
@@ -67,66 +87,168 @@ class ProfileForm extends React.Component {
...
+							className="form-horizontal"
+							onSubmit={this.handleSubmit}
 						>
-							Update!
-						</Button>
-					</div>
-				</div>
-			</form>
+							{loadingElement}
+							<Row>
```
LordSputnik: Hmm, let's remove the `<Row>` and `<Col>` elements surrounding the inputs, so that they take up the full width of the form. This may not look the best, but it'll be consistent with the entity editing PR, and then we can adjust both of them in the same ways to look good.

src/client/components/forms/profile.js
```diff
@@ -67,66 +87,168 @@ class ProfileForm extends React.Component {
...
+									/>
+								</Col>
+							</Row>
+							<div className="form-group">
+								<Col
+									md={4}
+									mdOffset={4}
+								>
+									<Button
+										block
```
LordSputnik: Also, once the `<Col>` around this button is removed, can you remove the `block` attribute here, and add a `text-center` class to the surrounding form group div?

LordSputnik:
One final issue: when no titles are available, and the form is submitted, the submission function tries to access this.title which is undefined - just need to check for this with something like:

    const title = this.title && this.title.getValue();

PR #126, opened 2016-12-31
Add JSdocs comments

dpmittal:
### Problem

### Solution

### Areas of Impact

### DIFF
--- a/src/client/components/pages/about.js
+++ b/src/client/components/pages/about.js
@@ -15,11 +15,16 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+* Renders the document and displays the 'About' page.
+* @returns {ReactElement} a HTML document which displays the About page
+*/
 function AboutPage() {
 	'use strict';
 
@@ -110,7 +115,7 @@ function AboutPage() {
 			<p>
 				The core components of BookBrainz are the site, the
 				webservice and the database. All of these components are
-				running on MetaBrainz's Sakura server, which also runs&nbsp;
+				running on MetaBrainz&rsquo;s Sakura server, which also runs&nbsp;
 				<a href="https://critiquebrainz.org/">
 					CritiqueBrainz
 				</a>.
--- a/src/client/components/pages/about.js
+++ b/src/client/components/pages/about.js
@@ -15,7 +15,6 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
@@ -23,7 +22,8 @@ const PageHeader = require('react-bootstrap').PageHeader;
 
 /**
 * Renders the document and displays the 'About' page.
-* @returns {ReactElement} a HTML document which displays the About page
+* @returns {ReactElement} a HTML document which displays the About
+* page
 */
 function AboutPage() {
 	'use strict';
@@ -115,7 +115,8 @@ function AboutPage() {
 			<p>
 				The core components of BookBrainz are the site, the
 				webservice and the database. All of these components are
-				running on MetaBrainz&rsquo;s Sakura server, which also runs&nbsp;
+				running on MetaBrainz&rsquo;s Sakura server, which also runs
+        &nbsp;
 				<a href="https://critiquebrainz.org/">
 					CritiqueBrainz
 				</a>.
--- a/src/client/components/pages/about.js
+++ b/src/client/components/pages/about.js
@@ -115,7 +115,7 @@ function AboutPage() {
 			<p>
 				The core components of BookBrainz are the site, the
 				webservice and the database. All of these components are
-				running on MetaBrainz&rsquo;s Sakura server, which also runs
+				running on MetaBrainz's Sakura server, which also runs
         &nbsp;
 				<a href="https://critiquebrainz.org/">
 					CritiqueBrainz
--- a/src/client/components/pages/contribute.js
+++ b/src/client/components/pages/contribute.js
@@ -15,15 +15,15 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
 /**
-* Renders the the document and display the 'Contribute' Page.
-* @returns {ReactElement} a HTML document which displays the contribute page
+* Renders the the document and displays the 'Contribute' page.
+* @returns {ReactElement} a HTML document which displays the
+* Contribute page
 */
 function ContributePage() {
 	'use strict';
--- a/src/client/components/pages/contribute.js
+++ b/src/client/components/pages/contribute.js
@@ -15,11 +15,16 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+* Renders the the document and display the 'Contribute' Page.
+* @returns {ReactElement} a HTML document which displays the contribute page
+*/
 function ContributePage() {
 	'use strict';
 
--- a/src/client/components/pages/develop.js
+++ b/src/client/components/pages/develop.js
@@ -15,11 +15,16 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+* Renders the document and display the 'Develop' Page.
+* @returns {ReactElement} a HTML document to display the Develop Page
+*/
 function DevelopPage() {
 	'use strict';
 
--- a/src/client/components/pages/develop.js
+++ b/src/client/components/pages/develop.js
@@ -15,15 +15,14 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
 /**
-* Renders the document and display the 'Develop' Page.
-* @returns {ReactElement} a HTML document to display the Develop Page
+* Renders the document and displays the 'Develop' page.
+* @returns {ReactElement} a HTML document to display the Develop page
 */
 function DevelopPage() {
 	'use strict';
--- a/src/client/components/pages/entity-revisions.js
+++ b/src/client/components/pages/entity-revisions.js
@@ -33,7 +33,7 @@ class EntityRevisions extends React.Component {
 /**
 * Binds the class methods to their respective data.
 * @constructor
-* @param {object} props Properties passed to the component
+* @param {object} props - Properties passed to the component
 */
 	constructor(props) {
 		super(props);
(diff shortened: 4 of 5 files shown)


src/client/components/pages/about.js
```diff
@@ -110,7 +115,7 @@ function AboutPage() {
 			<p>
 				The core components of BookBrainz are the site, the
 				webservice and the database. All of these components are
-				running on MetaBrainz's Sakura server, which also runs&nbsp;
+				running on MetaBrainz&#39;s Sakura server, which also runs&nbsp;
```
LordSputnik: Good change, but it would be better to use &rsquo; - more typographically correct and easier to remember the HTML character code.

src/client/components/pages/about.js
```diff
@@ -15,11 +15,16 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
```
LordSputnik: Let's update the rules globally for this, since we seem to be making exceptions everywhere for it. It seems like the right thing to be doing.

src/client/components/pages/about.js
```diff
@@ -15,11 +15,16 @@
...
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+* Renders the document and display the 'About' Page.
```
LordSputnik: "displays" rather than "display" and also lowercase "page".

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +26,28 @@ const Col = bootstrap.Col;
 const ListGroup = bootstrap.ListGroup;
 const ListGroupItem = bootstrap.ListGroupItem;
 
+/**
+* This is an extended class of the React Component to
```
LordSputnik: "This class is derived from the React Component base class and renders..." would be better, I think.

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +26,28 @@ const Col = bootstrap.Col;
...
+* render the 'Entity Revisions' page.
+*/
 class EntityRevisions extends React.Component {
 
+/**
+* Binds the class methods to their respective data.
+* @constructor
+* @param {object} props - Object of properties passed in the constructor
```
LordSputnik: "the properties passed to the component" would be better here

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +26,28 @@ const Col = bootstrap.Col;
...
+*/
 	constructor(props) {
 		super(props);
 		this.renderHeader = this.renderHeader.bind(this);
 		this.renderRevision = this.renderRevision.bind(this);
 	}
 
+/**
+* Renders the Header of the page consisting of Revision ID and
+* entity information in the form of a table.
```
LordSputnik: This isn't quite correct - there is no table involved - only the entity name and disambiguation comment are rendered here.

src/client/components/pages/entity-revisions.js
```diff
@@ -54,6 +70,12 @@ class EntityRevisions extends React.Component {
 		);
 	}
 
+/**
+* Renders the data related to Revision such as 'author' and 'date'.
+* It also displays 'Revision Notes' column.
```
LordSputnik: It actually displays the first revision note, which should be a summary of the changes made in the revision

src/client/components/pages/entity-revisions.js
```diff
@@ -54,6 +70,12 @@ class EntityRevisions extends React.Component {
...
+* @param {object} revision object is passed for it's data to be accessed
```
LordSputnik: "the revision to be represented by the rendered component" would be better here

src/client/components/pages/entity-revisions.js
```diff
@@ -83,6 +105,11 @@ class EntityRevisions extends React.Component {
 		);
 	}
 
+/**
+* Renders the 'Revision History' column of the EntityRevisions page
```
LordSputnik: I'd have something like this here: "Renders the EntityRevisions page, which is a list of all the revisions made to an entity, along with information about the author and the first revision note for each revision"

dpmittal:
@LordSputnik I have made the suggested changes :D

src/client/components/pages/about.js
LordSputnik: Leaving this as it was is fine, but it seems that GitHub changed what I meant to say - that using `&rsquo;` is better (rather than `'` or `&#39;`).

src/client/components/pages/contribute.js
```diff
@@ -15,11 +15,16 @@
...
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+* Renders the the document and display the 'Contribute' Page.
```
LordSputnik: This should also be "displays"

src/client/components/pages/develop.js
```diff
@@ -15,11 +15,16 @@
...
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: "error" */
 
 const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+* Renders the document and display the 'Develop' Page.
```
LordSputnik: This should also be "displays"

src/client/components/pages/contribute.js
```diff
@@ -15,11 +15,16 @@
...
+* @returns {ReactElement} a HTML document which displays the contribute page
```
LordSputnik: "the Contribute page" - you've capitalised elsewhere, so it makes sense to be consistent

src/client/components/pages/about.js
dpmittal: I have made the change :D

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +26,28 @@ const Col = bootstrap.Col;
 const ListGroup = bootstrap.ListGroup;
 const ListGroupItem = bootstrap.ListGroupItem;
 
+/**
+* The class is derived from React Component base class and
```
LordSputnik: "from *the* React Component..." makes more sense here

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +26,28 @@ const Col = bootstrap.Col;
...
+* renders the 'Entity Revisions' page.
+*/
 class EntityRevisions extends React.Component {
 
+/**
+* Binds the class methods to their respective data.
+* @constructor
+* @param {object} props - Object of properties passed to the component
```
LordSputnik: No need for "Object of" here

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +26,28 @@ const Col = bootstrap.Col;
...
 	constructor(props) {
 		super(props);
 		this.renderHeader = this.renderHeader.bind(this);
 		this.renderRevision = this.renderRevision.bind(this);
 	}
 
+/**
+* Renders the Header of the page consisting of Entity name and
+* disambiguation comment.
+* @returns {ReactElement} a HTML document which is a a part of Revision page
```
LordSputnik: Repeated "a" in this comment. Also it would make more sense if you wrote "is a part of the Revision page"

src/client/components/pages/entity-revisions.js
```diff
@@ -54,6 +70,14 @@ class EntityRevisions extends React.Component {
 		);
 	}
 
+/**
+* Renders the data related to Revision such as 'author' and 'date'.
+* It also displays 'Revision Notes' column which displays the first revison note
```
LordSputnik: No columns involved here - it renders the first revision note for each revision inside a list group item. The function is used in a loop in `render()` to render all of the revisions inside a list group.

src/client/components/pages/entity-revisions.js
```diff
@@ -54,6 +70,14 @@ class EntityRevisions extends React.Component {
...
+* which is a summary of the changes made in the revision.
+* @param {object} revision object is passed for the revison to be represented
```
LordSputnik: No need for "object is passed" - that's implied by `revision` being a parameter for the function.

src/client/components/pages/entity-revisions.js
```diff
@@ -83,6 +107,12 @@ class EntityRevisions extends React.Component {
 		);
 	}
 
+/**
+* Renders the EntityRevisions page, which is a list of all the revisions made
+* to an entity, along with information about the author and the
+* first revision note for each revision.
+* @returns {ReactElement} a HTML document which is a part of the Revision page
```
LordSputnik: "a HTML document which displays the Revision page" - this renders the whole thing, not just a part

src/client/components/pages/about.js
```diff
@@ -15,11 +15,16 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: "error" */
```
LordSputnik: What I meant here was it would be better to add the `requiresReturn: false` option to the `.eslintrc` file in the root directory of the repository, to save us from having to add a comment at the top of each file.

src/client/components/pages/about.js
```diff
@@ -110,7 +115,8 @@ function AboutPage() {
 			<p>
 				The core components of BookBrainz are the site, the
 				webservice and the database. All of these components are
-				running on MetaBrainz's Sakura server, which also runs&nbsp;
```
Freso: Why are you changing this line?

src/client/components/pages/entity-revisions.js
```diff
@@ -54,6 +69,14 @@ class EntityRevisions extends React.Component {
 		);
 	}
 
+/**
+* Renders the data related to Revision such as 'author' and 'date'.
+* It also displays the first revison note which is a summary of the changes
+* made in the revision.
+* @param {object} revision for the revison to be represented  by the
```
Freso: Two spaces between "represented" and "by".

src/client/components/pages/entity-revisions.js
```diff
@@ -16,6 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+
```
Freso: It is usually better to keep code style changes (such as whitespace) in separate commits (or even completely separate PRs), so they can be reviewed outside of "actual" code changes.

src/client/components/pages/entity-revisions.js
```diff
@@ -24,14 +25,28 @@ const Col = bootstrap.Col;
...
+/**
+* The class is derived from the React Component base class and
+* renders the 'Entity Revisions' page.
+*/
 class EntityRevisions extends React.Component {
 
+/**
+* Binds the class methods to their respective data.
+* @constructor
+* @param {object} props - Properties passed to the component
```
Freso: This `@param` is formatted differently than the one for `renderRevision`. Is this intentional? Did you purposefully uppercase "Properties"?

src/client/components/pages/about.js
dpmittal: It improved the readability of the code. @LordSputnik  has approved these changes you can see his 3rd comment from top for reference :)

src/client/components/pages/entity-revisions.js
dpmittal: This `@param` is formatted correctly but the `renderRevsion` so I'm changing that now. Properties hasn't been purposely uppercased. Should I lowercase it?

src/client/components/pages/entity-revisions.js
dpmittal: I'm changing it now :) it's a silly mistake

src/client/components/pages/entity-revisions.js
dpmittal: It was just one line addition so i thought it to include it in this PR? Should I remove it and send a separate PR?

LordSputnik: Shouldn't need to add this line at all - it's just left over from the eslint comment you added earlier.

src/client/components/pages/about.js
LordSputnik: I don't mind this change - not part of the changes mentioned in the PR, so generally it would be better to keep it separate (don't worry for now), but it's an improvement nonetheless.

src/client/components/pages/entity-revisions.js
LordSputnik: http://usejsdoc.org/tags-param.html uses uppercase `@param` descriptions, so let's follow that (although it will vary from project to project).

src/client/components/pages/entity-revisions.js
LordSputnik: I'd also remove the "for" here, and have the description start with "the revision..."

dpmittal:
@LordSputnik Made the changes :)

LordSputnik:

PR #128, opened 2017-01-01
Added JSDoc comments in about.js

SrijanJha28:
Added JSDoc comments in about.js in the function AboutPage

### DIFF
--- a/src/client/components/pages/about.js
+++ b/src/client/components/pages/about.js
@@ -20,6 +20,13 @@ const React = require('react');
 
 const PageHeader = require('react-bootstrap').PageHeader;
 
+/**
+*Description of the function AboutPage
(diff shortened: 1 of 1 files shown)


LordSputnik:
Unfortunately, comments have already been added to the about.js file by another GCI student in https://github.com/bookbrainz/bookbrainz-site/pull/126 - please choose another of the pages to add comments to (one that isn't in #126).

Also, some general feedback about this submission - you should aim to describe what a function does, and how it interacts with the rest of the system, rather than describing the content of the site. It may take some time to look at the code base and work this out so that you have a detailed enough comment.

Also, if you find that the task takes you much less than the three hour estimated time, you might consider commenting a few more files, which will be a more impressive achievement.

Thanks for your work so far!

PR #130, opened 2017-01-05
New error component - GCI

Arthelon:
### Problem
Current error component is being rendered solely with Pug, looks horrible, and does not display any useful information.

### Solution
Modifies server code to send down more detailed responses with information taken from the req object, creates an error page react component, adds a bit more styling to the original page.

### Areas of Impact
Makes non-breaking updates to all Error constructors to include the req express object and detailed message field. Modifies 404 handler to render down page with React. 

### NOTE
PR isn't complete. A detailed message has only been created for the 404 status code. Would like to get some feedback on this approach before continuing further.

### DIFF
--- a/src/client/components/pages/error.js
+++ b/src/client/components/pages/error.js
@@ -40,10 +40,10 @@ class ErrorPage extends React.Component {
 				<Row className="margin-bottom-6">
 					<h1>{error.status}</h1>
 				</Row>
-				<Row className="margin-top-6 margin-bottom-2">
-					<b className="lead">
-						{error.message}
-					</b>
+				<Row className="margin-top-6 margin-bottom-1">
+					<p className="lead">
+						<b>{error.message}</b>
+					</p>
 				</Row>
 				<div>
 					{detailedMessage &&
--- a/src/client/components/pages/error.js
+++ b/src/client/components/pages/error.js
@@ -17,6 +17,8 @@
  */
 const React = require('react');
 const Button = require('react-bootstrap').Button;
+const Grid = require('react-bootstrap').Grid;
+const Row = require('react-bootstrap').Row;
 
 /**
  * Links to different pages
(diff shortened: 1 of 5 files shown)


src/server/helpers/error.js
```diff
@@ -19,14 +19,21 @@
...
 const status = require('http-status');
+const React = require('react');
+const ReactDOMServer = require('react-dom/server');
+const propHelpers = require('./props');
+const Layout = require('../../client/containers/Layout');
+const ErrorPage = require('../../client/components/pages/error');
 
 class SiteError extends Error {
-	constructor(message) {
+	constructor(message, req) {
```
LordSputnik: Hmm, I feel like we should only provide `req` as a parameter to error classes that are route-specific. So, for example, add a constructor to those classes which takes `message` and `req` parameters, calls `super(message)` and then creates the detailed message.

This'll allow us to keep using SiteError for things like errors during the app startup (outside of requests).

src/client/components/pages/error.js
```diff
@@ -0,0 +1,87 @@
...
+		const {error} = this.props;
+		let detailedMessage = error.detailedMessage;
+		if (typeof detailedMessage === 'string') {
+			detailedMessage = [detailedMessage];
+		}
+
+		return (
+			<div
+				style={{
+					display: 'flex',
```
LordSputnik: Does this need to be a flex box? That won't work in IE 9, which is one of our supported browsers: https://bb-user-guide.readthedocs.io/en/latest/reference/browser-support/

src/client/components/pages/error.js
```diff
@@ -0,0 +1,87 @@
...
+					flexDirection: 'column',
+					alignItems: 'center',
+					justifyContent: 'center'
+				}}
+			>
+				<h1>{error.status}</h1>
+				<b
+					style={{
+						marginTop: '1.5em',
```
LordSputnik: Hmm, use the margin-top-1 and margin-bottom-1 classes here rather than inline styling (there will be a margin-top-1d5 or equivalent soon, but not added in the master branch yet).

src/server/helpers/error.js
Arthelon: +1

src/client/components/pages/error.js
```diff
@@ -0,0 +1,81 @@
...
+		/* to-do: Adjust margins for error status title and message once image
+		   is added in
+		*/
+		return (
+			<Grid className="text-center">
+				<Row className="margin-bottom-6">
+					<h1>{error.status}</h1>
+				</Row>
+				<Row className="margin-top-6 margin-bottom-2">
+					<b style={{fontSize: '1.2em'}}>
```
LordSputnik: This might give the styling you want without needing to specify it directly:

    <p className="lead">
        <b>{error.message}</b>
    </p>

src/client/components/pages/error.js
```diff
@@ -0,0 +1,81 @@
...
+						{error.message}
+					</b>
+				</Row>
+				<div>
+					{Boolean(detailedMessage) &&
```
LordSputnik: I don't *think* the Boolean cast is needed here. `x && y` should always evaluate the operands as Boolean values, and return x if x is falsey, y otherwise.

src/client/components/pages/error.js
```diff
@@ -41,12 +41,12 @@ class ErrorPage extends React.Component {
 					<h1>{error.status}</h1>
 				</Row>
 				<Row className="margin-top-6 margin-bottom-2">
-					<b style={{fontSize: '1.2em'}}>
+					<b className="lead">
```
LordSputnik: Hmm, either just `<p className="lead">` or `<p className="lead"><b>` here - the lead class overrides the font weight styling on the <b> element as it is.

PR #132, opened 2017-01-07
Add JSdoc comments to Input Component files

dpmittal:
### Problem

### Solution

### Areas of Impact

### DIFF
--- a/src/client/components/input/drag-and-drop-image.js
+++ b/src/client/components/input/drag-and-drop-image.js
@@ -2,8 +2,8 @@ const React = require('react');
 const Input = require('react-bootstrap').Input;
 
 /**
-* This class is derived from the React Component base class to render an image
-* which supports drag and drop functionality.
+* This class is derived from the React Component base class and renders
+* an image which supports drag and drop functionality.
 */
 class DragAndDropImage extends React.Component {
 	/**
@@ -16,8 +16,9 @@ class DragAndDropImage extends React.Component {
 	}
 	/**
 	* Transfers the data of the achievement badge component properties to the
-	* DragAndDrop event, which is used for displaying achievement data on the
-	* profile editor page.
+	* DragAndDrop event, which in turn transfers the data on handleDrop to that
+	* of the achievement badge which will be showcased on editor's
+	* public profile.
 	* @param {object} ev - Passed in the function to be initialized with data
 	* onDragStart.
 	*/
@@ -30,9 +31,9 @@ class DragAndDropImage extends React.Component {
 		ev.dataTransfer.setData('text', JSON.stringify(data));
 	}
 	/**
-	* Renders an image of the achievement badges which supports
-	* DragAndDrop functionality.
-	* @returns {ReactElement} a HTML image supporting DragAndDrop functionality.
+	* Renders an image of a particular achievement badge, which can be dragged
+	* to set the user's publicly showcased achievements
+	* @returns {ReactElement} - The rendered image element.
 	*/
 	render() {
 		return (
--- a/src/client/components/input/drag-and-drop-image.js
+++ b/src/client/components/input/drag-and-drop-image.js
@@ -2,8 +2,8 @@ const React = require('react');
 const Input = require('react-bootstrap').Input;
 
 /**
-* This class is derived from the React Component base class to render the
-* drag and drop image functionality.
+* This class is derived from the React Component base class to render an image
+* which supports drag and drop functionality.
 */
 class DragAndDropImage extends React.Component {
 	/**
@@ -15,8 +15,9 @@ class DragAndDropImage extends React.Component {
 		this.handleDragStart = this.handleDragStart.bind(this);
 	}
 	/**
-	* Binds the data of the achievement badge on drag start to the data of the
-	* of the achievements displayed on the profile of the editor.
+	* Transfers the data of the achievement badge component properties to the
+	* DragAndDrop event, which is used for displaying achievement data on the
+	* profile editor page.
 	* @param {object} ev - Passed in the function to be initialized with data
 	* onDragStart.
 	*/
@@ -29,8 +30,9 @@ class DragAndDropImage extends React.Component {
 		ev.dataTransfer.setData('text', JSON.stringify(data));
 	}
 	/**
-	* Renders the DragAndDrop functionality of achievement badges.
-	* @returns {ReactElement} a HTML DragAndDrop functionality.
+	* Renders an image of the achievement badges which supports
+	* DragAndDrop functionality.
+	* @returns {ReactElement} a HTML image supporting DragAndDrop functionality.
 	*/
 	render() {
 		return (
--- a/src/client/components/input/drag-and-drop-image.js
+++ b/src/client/components/input/drag-and-drop-image.js
@@ -1,11 +1,25 @@
 const React = require('react');
 const Input = require('react-bootstrap').Input;
 
+/**
+* This class is derived from the React Component base class to render the
+* drag and drop image functionality.
+*/
 class DragAndDropImage extends React.Component {
+	/**
+	* Binds the class methods to their respective data.
+	* @constructor
+	*/
 	constructor() {
 		super();
 		this.handleDragStart = this.handleDragStart.bind(this);
 	}
+	/**
+	* Binds the data of the achievement badge on drag start to the data of the
+	* of the achievements displayed on the profile of the editor.
+	* @param {object} ev - Passed in the function to be initialized with data
+	* onDragStart.
+	*/
 	handleDragStart(ev) {
 		const data = {
 			id: this.props.achievementId,
@@ -14,6 +28,10 @@ class DragAndDropImage extends React.Component {
 		};
 		ev.dataTransfer.setData('text', JSON.stringify(data));
 	}
+	/**
+	* Renders the DragAndDrop functionality of achievement badges.
+	* @returns {ReactElement} a HTML DragAndDrop functionality.
+	*/
 	render() {
 		return (
 			<img
--- a/src/client/components/input/partial-date.js
+++ b/src/client/components/input/partial-date.js
@@ -81,7 +81,7 @@ class PartialDate extends React.Component {
 	/**
 	* An event handler, which is triggered whenever the value in the date field
 	* changes and then, it validates the PartialDate value and updates
-	* the components state with the new value and validation state.
+	* the component's state with the new value and validation state.
 	* It also triggers the onChange function if the validation state is true.
 	*/
 	handleChange() {
--- a/src/client/components/input/partial-date.js
+++ b/src/client/components/input/partial-date.js
@@ -27,16 +27,16 @@ const yRegex = /^\d{4}$/;
 
 /**
 * This class is derived from the React Component base class
-* to render the partial date of any of the formats "YYYY", "YYYY-MM",
-* "YYYY-MM-DD" which allows the editor to be vague where information is not
+* to render a partial date (any of the formats "YYYY", "YYYY-MM",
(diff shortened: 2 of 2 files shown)


src/client/components/input/partial-date.js
```diff
@@ -25,7 +25,17 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
 const ymRegex = /^\d{4}-\d{2}$/;
 const yRegex = /^\d{4}$/;
 
+/**
+* This class is derived from the React Component base class
+* to render the date field and check validity of the value filled in it.
```
LordSputnik: This actually renders a partial date - that is, it can render and validate any of these three formats: "YYYY", "YYYY-MM", "YYYY-MM-DD" - this allows the editor to be vague where information isn't know.

src/client/components/input/partial-date.js
```diff
@@ -25,7 +25,17 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
...
+*/
 class PartialDate extends React.Component {
+	/**
+	* Function to check validity of the date filled in the field.
+	* @param {string} value - Date filled in field is passed in the validator.
+	* @returns {boolean} - Returns true either if no data is filled in the field
+	* or is the date filled in the field is valid.
```
LordSputnik: or *if*

src/client/components/input/partial-date.js
```diff
@@ -53,10 +68,18 @@ class PartialDate extends React.Component {
...
+	* To get the value filled in the form field.
+	* @returns {string} - Value filled in the form field.
+	*/
 	getValue() {
 		return this.state.value;
 	}
 
+	/**
+	* Track the changes in the date field of the form and validate the value in
+	* the field after each change.
```
LordSputnik: Also important to mention that it updates the component state

src/client/components/input/partial-date.js
```diff
@@ -74,10 +97,20 @@ class PartialDate extends React.Component {
 		}
 	}
 
+	/**
+	* Calls the date validator function of the 'PartialDate' class.
```
LordSputnik: What value is passed in to the validator function?

src/client/components/input/partial-date.js
```diff
@@ -74,10 +97,20 @@ class PartialDate extends React.Component {
...
+	* @returns {boolean} - The result as obtained from the date validator.
+	*/
 	valid() {
 		return PartialDate.validate(this.input.getValue().trim());
 	}
 
+	/**
+	* Checks that whether the editor has succesfully filled the date field in
```
LordSputnik: It does do this, but that's not the purpose of this function. It exists to provide the correct validation CSS class to the field input.

src/client/components/input/partial-date.js
```diff
@@ -88,6 +121,10 @@ class PartialDate extends React.Component {
 		return validationClass;
 	}
 
+	/**
+	* Renders the date input field in the form.
```
LordSputnik: Not necessarily in a form. Just "Renders the PartialDate input field" would work here.

src/client/components/input/partial-date.js
```diff
@@ -88,6 +121,10 @@ class PartialDate extends React.Component {
...
+	* @returns {ReactElement} - A HTMl field in the form.
```
LordSputnik: "the rendered input field element" would be better here

src/client/components/input/drag-and-drop-image.js
```diff
@@ -14,6 +28,10 @@ class DragAndDropImage extends React.Component {
 		};
 		ev.dataTransfer.setData('text', JSON.stringify(data));
 	}
+	/**
+	* Renders the DragAndDrop functionality of achievement badges.
+	* @returns {ReactElement} a HTML DragAndDrop functionality.
```
LordSputnik: "Functionality" is a process, so it can't really be rendered. Could you rewrite this in a better way? What is actually being rendered here?

src/client/components/input/drag-and-drop-image.js
```diff
@@ -1,11 +1,25 @@
...
+	/**
+	* Binds the class methods to their respective data.
+	* @constructor
+	*/
 	constructor() {
 		super();
 		this.handleDragStart = this.handleDragStart.bind(this);
 	}
+	/**
+	* Binds the data of the achievement badge on drag start to the data of the
```
LordSputnik: Where does this data end up? Take a look at the other drag and drop components to see if you can find out, and then add some relevant information here. I'd also reword this comment to talk about transfer of data from the component properties to the event.

src/client/components/input/drag-and-drop-image.js
```diff
@@ -1,11 +1,25 @@
 const React = require('react');
 const Input = require('react-bootstrap').Input;
 
+/**
+* This class is derived from the React Component base class to render the
+* drag and drop image functionality.
```
LordSputnik: "to render an image which supports drag and drop functionality" sounds better here

dpmittal:
@LordSputnik I have made the requested changes

src/client/components/input/partial-date.js
```diff
@@ -25,7 +25,19 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
 const ymRegex = /^\d{4}-\d{2}$/;
 const yRegex = /^\d{4}$/;
 
+/**
+* This class is derived from the React Component base class
+* to render the partial date of any of the formats "YYYY", "YYYY-MM",
```
LordSputnik: This makes more sense if you put it as follows:

    ...and renders a partial date (any of the formats "YYYY",
    "YYYY-MM", "YYYY-MM-DD"). This allows...

src/client/components/input/partial-date.js
```diff
@@ -25,7 +25,19 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
...
+* "YYYY-MM-DD" which allows the editor to be vague where information is not
+* known. It also checks validity of the value filled in the date field.
+*/
 class PartialDate extends React.Component {
+	/**
+	* Function to check validity of the date filled in the field.
```
LordSputnik: "... validity of a partial date" - this is a static function, so it's not really linked to the field at all.

src/client/components/input/partial-date.js
```diff
@@ -25,7 +25,19 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
...
+	* @param {string} value - Date filled in field is passed in the validator.
```
LordSputnik: No need to mention that it's passed in - that's obvious from the context. Also not linked to the field. So a better comment here might be "The partial date to be validated"

src/client/components/input/partial-date.js
```diff
@@ -25,7 +25,19 @@ const ymdRegex = /^\d{4}-\d{2}-\d{2}$/;
...
+	* @returns {boolean} - Returns true either if no data is filled in the field
```
LordSputnik: No need to repeat "Returns" here - just "True if the date is valid or if no date is present" is good

src/client/components/input/partial-date.js
```diff
@@ -53,10 +70,18 @@ class PartialDate extends React.Component {
...
+	/**
+	* To get the value filled in the form field.
+	* @returns {string} - Value filled in the form field.
+	*/
 	getValue() {
 		return this.state.value;
 	}
 
+	/**
+	* Track the changes in the date field of the form and validate the value in
```
LordSputnik: Hmm, this doesn't really track anything, strictly speaking. It's "An event handler, which is triggered whenever the value in the date field changes". And it "validates the value... and updates the components state with the new value and validation state". It also calls the component's `onChange` handler, if this is set.

src/client/components/input/partial-date.js
```diff
@@ -74,10 +99,20 @@ class PartialDate extends React.Component {
...
+	* @returns {boolean} - The result as obtained from the date validator.
+	*/
 	valid() {
 		return PartialDate.validate(this.input.getValue().trim());
 	}
 
+	/**
+  * Applies correct validationClass to the CSS of the field input. based on
+	* results obtained from the valid() function.
+	* @returns {string} validationClass - State of the date field in the form.
```
LordSputnik: `The validation class to be applied to the date field`

src/client/components/input/partial-date.js
```diff
@@ -74,10 +99,20 @@ class PartialDate extends React.Component {
...
+	* Calls the date validator function of the 'PartialDate' class and the value
+	* filled in the date field is passed for validation.
+	* @returns {boolean} - The result as obtained from the date validator.
+	*/
 	valid() {
 		return PartialDate.validate(this.input.getValue().trim());
 	}
 
+	/**
+  * Applies correct validationClass to the CSS of the field input. based on
```
LordSputnik: `input, based`

src/client/components/input/drag-and-drop-image.js
```diff
@@ -1,11 +1,26 @@
 const React = require('react');
 const Input = require('react-bootstrap').Input;
 
+/**
+* This class is derived from the React Component base class to render an image
```
LordSputnik: `and renders` rather than `to render`

src/client/components/input/drag-and-drop-image.js
```diff
@@ -1,11 +1,26 @@
...
+	/**
+	* Binds the class methods to their respective data.
+	* @constructor
+	*/
 	constructor() {
 		super();
 		this.handleDragStart = this.handleDragStart.bind(this);
 	}
+	/**
+	* Transfers the data of the achievement badge component properties to the
```
LordSputnik: This is better, but still not specific enough. Take a look at https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/input/drag-and-drop.js#L28 in particular - that function is where this data ends up. Then come back to this comment and describe where the data goes, and what it's used for and how this happens.

src/client/components/input/drag-and-drop-image.js
```diff
@@ -14,6 +29,11 @@ class DragAndDropImage extends React.Component {
 		};
 		ev.dataTransfer.setData('text', JSON.stringify(data));
 	}
+	/**
+	* Renders an image of the achievement badges which supports
```
LordSputnik: Better! How about `Renders an image of a particular achievement badge, which can be dragged to set the user's publicly showcased achievements`?

src/client/components/input/drag-and-drop-image.js
```diff
@@ -14,6 +29,11 @@ class DragAndDropImage extends React.Component {
...
+	* DragAndDrop functionality.
+	* @returns {ReactElement} a HTML image supporting DragAndDrop functionality.
```
LordSputnik: `the rendered image element` works better here (more consistent with other React `@returns` comments, and avoids repeating the part about DragAndDrop functionality).

src/client/components/input/drag-and-drop-image.js
dpmittal: Rather than rephrasing the whole comment I added a little more context as to where the data ends up, though I'm not sure that it is the change which you intended.

src/client/components/input/partial-date.js
```diff
@@ -79,8 +79,10 @@ class PartialDate extends React.Component {
 	}
 
 	/**
-	* Track the changes in the date field of the form and validate the value in
-	* the field after each change and it updates the component state.
+	* An event handler, which is triggered whenever the value in the date field
+	* changes and then, it validates the PartialDate value and updates
+	* the components state with the new value and validation state.
```
LordSputnik: `component state` or `component's state`

dpmittal:
@LordSputnik made the changes :)

PR #133, opened 2017-01-08
Consistent dates for existing React components, entity templates conversion - GCI

Arthelon:
### Problem
Current date formatting with client side JS is affected by browser locale, resulting in inconsistent dates.

### Solution
Introduces a formatDate helper function, formats all dates in existing React components with the helper function.

### Areas of Impact
Client React Components, helpers/utils module.

Next step after this PR is to convert templates using dates (i.e. the entity template) to React so that date formatting can be applied non-obstrusively.

### DIFF
--- a/src/client/components/forms/parts/creator-data.js
+++ b/src/client/components/forms/parts/creator-data.js
@@ -27,6 +27,7 @@ const Identifiers = require('./identifier-list');
 const PartialDate = require('../../input/partial-date');
 const Select = require('../../input/select2');
 const SearchSelect = require('../../input/entity-search');
+const formatDate = require('../../../helpers/utils').formatDate;
 
 const validators = require('../../../helpers/react-validators');
 const injectDefaultAliasName =
@@ -90,8 +91,8 @@ class CreatorData extends React.Component {
 			if (prefillData.endArea) {
 				initialEndArea = prefillData.endArea;
 			}
-			initialBeginDate = prefillData.beginDate;
-			initialEndDate = prefillData.endDate;
+			initialBeginDate = formatDate(new Date(prefillData.beginDate));
+			initialEndDate = formatDate(new Date(prefillData.endDate));
 			initialGender = prefillData.gender ?
 				prefillData.gender.id : null;
 			initialCreatorType = prefillData.creatorType ?
--- a/src/client/components/forms/parts/creator-data.js
+++ b/src/client/components/forms/parts/creator-data.js
@@ -27,7 +27,6 @@ const Identifiers = require('./identifier-list');
 const PartialDate = require('../../input/partial-date');
 const Select = require('../../input/select2');
 const SearchSelect = require('../../input/entity-search');
-const formatDate = require('../../../helpers/utils').formatDate;
 
 const validators = require('../../../helpers/react-validators');
 const injectDefaultAliasName =
@@ -91,8 +90,8 @@ class CreatorData extends React.Component {
 			if (prefillData.endArea) {
 				initialEndArea = prefillData.endArea;
 			}
-			initialBeginDate = formatDate(new Date(prefillData.beginDate));
-			initialEndDate = formatDate(new Date(prefillData.endDate));
+			initialBeginDate = prefillData.beginDate;
+			initialEndDate = prefillData.endDate;
 			initialGender = prefillData.gender ?
 				prefillData.gender.id : null;
 			initialCreatorType = prefillData.creatorType ?
--- a/src/client/components/forms/parts/edition-data.js
+++ b/src/client/components/forms/parts/edition-data.js
@@ -20,6 +20,7 @@
 const Icon = require('react-fontawesome');
 const React = require('react');
 const _assign = require('lodash.assign');
+const formatDate = require('../../../helpers/utils').formatDate;
 
 const Input = require('react-bootstrap').Input;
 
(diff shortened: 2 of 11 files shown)


src/client/components/pages/entity-revisions.js
```diff
@@ -78,9 +79,11 @@ class EntityRevisions extends React.Component {
...
 	renderRevision(revision) {
 		const createdDate = new Date(revision.revision.createdAt);
-		const dateLabel = (Date.now() - createdDate < 86400000) ?
-			createdDate.toLocaleTimeString() :
-			createdDate.toLocaleDateString();
+		let includeTime = false;
+		if (Date.now() - createdDate.getTime() < 86400000) {
+			includeTime = true;
+		}
+		const dateLabel = formatDate(createdDate, includeTime);
```
LordSputnik: `const includeTime = Boolean(Date.now() - createdDate.getTime() < 86400000);` is simpler here :)

LordSputnik: In fact it would be better to have this as a function and stick it in the helpers module since it's reused elsewhere.

Arthelon:
Added code for the entity template conversion. Also converted the edition entity page to start with. If the pattern used works well I can continue to do the same for the rest of the entity templates.

LordSputnik:
Hmm, could you undo the merge here? I'd prefer to merge everything up to 29fd7ff first, and then the Pug template work.

LordSputnik:
But generally the pattern you've used for the entity components looks OK - I should note that we're already planning to redesign the entity components, so it's up to you whether you want to convert them here first. It might be useful if you do, so that we can find solutions for any potential unforeseen problems before the rewrite.

src/client/components/forms/parts/publisher-data.js
```diff
@@ -76,14 +77,15 @@ class PublisherData extends React.Component {
...
+		console.log(this.props.publisher);
 
 		const prefillData = this.props.publisher;
 		if (prefillData) {
 			if (prefillData.area) {
 				initialArea = prefillData.area;
 			}
-			initialBeginDate = prefillData.beginDate;
-			initialEndDate = prefillData.endDate;
+			initialBeginDate = formatDate(new Date(prefillData.beginDate));
```
LordSputnik: Hmm can't do this here - these are partial dates, which can't be converted to dates (can be "YYYY") for example. They don't need formatting though, since they're already formatted strings.

src/client/components/forms/parts/publisher-data.js
```diff
@@ -76,14 +77,15 @@ class PublisherData extends React.Component {
 		let initialDisambiguation = null;
 		let initialAnnotation = null;
 		let initialIdentifiers = [];
+		console.log(this.props.publisher);
```
LordSputnik: Stray console log

src/client/components/forms/parts/edition-data.js
```diff
@@ -109,7 +110,9 @@ class EditionData extends React.Component {
 			if (prefillData.releaseEventSet &&
 				prefillData.releaseEventSet.releaseEvents) {
 				initialReleaseDate =
-					prefillData.releaseEventSet.releaseEvents[0].date;
+					formatDate(new Date(
+						prefillData.releaseEventSet.releaseEvents[0].date
```
LordSputnik: Also a partial date - already formatted/can't be formatted.

src/client/components/forms/parts/creator-data.js
```diff
@@ -90,8 +91,8 @@ class CreatorData extends React.Component {
 			if (prefillData.endArea) {
 				initialEndArea = prefillData.endArea;
 			}
-			initialBeginDate = prefillData.beginDate;
-			initialEndDate = prefillData.endDate;
+			initialBeginDate = formatDate(new Date(prefillData.beginDate));
```
LordSputnik: Also a partial date - already formatted/can't be formatted.

src/client/helpers/utils.js
```diff
@@ -34,4 +34,24 @@ function injectDefaultAliasName(instance) {
 	return instance;
 }
 
+function formatDate(date, includeTime) {
+	'use strict';
+	if (!date || !(Object.prototype.toString.call(date) === '[object Date]') ||
```
leftmostcat: I'm not clear on what the second condition here is trying to accomplish. Is this going for `date instanceof Date`?

src/client/helpers/utils.js
```diff
@@ -34,4 +34,24 @@ function injectDefaultAliasName(instance) {
...
+		return null;
+	}
+	const formatter = moment(date);
+	if (includeTime) {
+		return formatter.format('YYYY-MM-DD HH:mm:ss');
+	}
+	return formatter.format('YYYY-MM-DD');
+}
+
+function isWithinDay(date) {
```
leftmostcat: This might be just a little clearer as `isWithinDayOfPresent`.

leftmostcat:
Looking good. Just a couple small things.

src/client/helpers/utils.js
Arthelon: Yup, just something I've gotten used to doing to avoid cross-frame issues.

src/client/helpers/utils.js
Arthelon: I'll use isWithinDayFromNow

src/client/helpers/utils.js
LordSputnik: Hmm, could you explain this a bit more?

Arthelon: as brought up by @leftmost, this condition is just a more reliable way of checking the object prototype since date objects do not have the same prototype across different frames. Granted, this scenario is unlikely to pop up. I'll add in a comment explaining purpose of this condition.

PR #134, opened 2017-01-09
Fixes typo on 'Contributing' page, also adds parameter for extra function to extend PartialDate's validation functionality.

inondle:
### Problem
Firstly, there was a missing space in the 'contributing' page so I added one in.
Second problem was that the 'brithdate' field of the registration and profile forms allowed users to enter birthdays in the future.

### Solution
The PR adds a parameter to `PartialDate.validate()` that takes a function and, if defined, also validates the value with that function. The forms that use `PartialDate` as a field for taking birthdays now pass in an extra function to check that the date put into the field is in the past.

### Areas of Impact
This modifies the validation functions of `PartialDate`, `PartialDate` can now be rendered with the `customValidation` field which takes in a callback for extra validation.

### DIFF
--- a/src/client/components/forms/profile.js
+++ b/src/client/components/forms/profile.js
@@ -33,6 +33,8 @@ const PartialDate = require('../input/partial-date');
 const validators = require('../../helpers/react-validators');
 const injectDefaultAliasName =
 	require('../../helpers/utils').injectDefaultAliasName;
+const isValidUserBirthday =
+	require('../../helpers/utils').isValidUserBirthday;
 
 class ProfileForm extends React.Component {
 	constructor(props) {
@@ -183,6 +185,7 @@ class ProfileForm extends React.Component {
 								ref={(ref) => this.gender = ref}
 							/>
 							<PartialDate
+								customValidator={isValidUserBirthday}
 								defaultValue={initialBirthDate}
 								label="Birth Date"
 								placeholder="YYYY-MM-DD"
--- a/src/client/components/forms/registration-details.js
+++ b/src/client/components/forms/registration-details.js
@@ -29,6 +29,8 @@ const PartialDate = require('../input/partial-date');
 
 const LoadingSpinner = require('../loading-spinner');
 const validators = require('../../helpers/react-validators');
+const isValidUserBirthday =
+	require('../../helpers/utils').isValidUserBirthday;
 
 class RegistrationForm extends React.Component {
 	constructor(props) {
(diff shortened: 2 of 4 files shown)


inondle:
Seems the CI doesn't like the constructor. I'll update it rn.

LordSputnik:
Rather than creating a new class here, it'd be simpler to update the original class to allow the user to set a custom validation to be run in addition to the standard validation. Then we can just reuse the same component and just pass in a different validation strategy.

src/client/components/forms/profile.js
```diff
@@ -123,6 +123,10 @@ class ProfileForm extends React.Component {
 			initialBirthDate = `${year}-${month}-${day}`;
 		}
 
+		function birthdayValidation(value) {
```
leftmostcat: I'd suggest `isValidUserBirthday` or the like. That makes it clear that it returns a bool and exactly what its purpose is.

src/client/components/input/partial-date.js
```diff
@@ -109,6 +118,7 @@ class PartialDate extends React.Component {
 
 PartialDate.displayName = 'PartialDate';
 PartialDate.propTypes = {
+	customValidation: React.PropTypes.func,
```
leftmostcat: Maybe `customValidator`, as it's active functionality.

src/client/components/input/partial-date.js
```diff
@@ -38,7 +38,12 @@ class PartialDate extends React.Component {
 		);
 		const validValue = !isNaN(Date.parse(value));
 
-		return validSyntax && validValue;
+		let cv = true;
```
leftmostcat: `cv` isn't a very clear name. I'd prefer something that indicates at a glance what this represents (a boolean indicating whether the date passed the custom validator).

src/client/components/forms/registration-details.js
```diff
@@ -104,6 +104,10 @@ class RegistrationForm extends React.Component {
 			width: '100%'
 		};
 
+		function birthdayValidation(value) {
```
leftmostcat: Instead of repeating here, this would be best split into a separate helper file.

leftmostcat:
Some naming suggestions and one small refactor, but otherwise looks reasonable to me.

Androbin:
Any updates on this yet?

LordSputnik:
Closing, as I'm not longer able to merge this (there is a technical issue with the patch GitHub is giving me). The work looks good, though, so if you're able to resubmit, I'm happy to review.

Androbin:
@inondle deleted his fork. I recovered, rebased and resubmitted as #206. Commit authorship preserved.

PR #135, opened 2017-01-12
Entity template conversion

Arthelon:
### Problem
Date formatting across the site is inconsistent and entity pages are still being rendered with Pug.

### Solution
Converts all entity templates to React, updates server rendering code, includes date formatting.

### Areas of Impact
The displayEntity method in the entity module was altered to render React components. Previous entity templates were also removed. Additional utility files and entity React components were created.


src/client/components/pages/entity/creator.js
```diff
@@ -0,0 +1,53 @@
...
+class CreatorPage extends React.Component {
+
+	static get iconName() {
+		return 'user';
+	}
+
+	static attributes(entity) {
+		return (
+			<div>
+				{showEntityType(entity.creatorType)}
```
LordSputnik: My first thought is that we should have an AttributeList component, which takes an Object property, and renders the keys and values as `<dd>` and `<dt>` elements respectively.

src/client/containers/entity.js
```diff
@@ -0,0 +1,283 @@
...
+							</small>
+						}
+					</h1>
+					<hr/>
+				</Col>
+			</Row>
+		);
+	}
+
+	renderContent() {
```
LordSputnik: This function would be better split into parts.

src/client/components/pages/entity/creator.js
Arthelon: agreed

src/client/components/pages/entity/edition.js
```diff
@@ -0,0 +1,99 @@
...
+		>
+			<p>
+				{entity.publication ?
+					<a href={`/publication/${entity.publication.bbid}`}>
+						<FontAwesome name="external-link"/>
+						{' See all other editions'}
+					</a> :
+					<span className="bg-danger">
+							Publication unset - please add one if you see this!
+						</span>
```
LordSputnik: Slight mis-indentation here (at least in the GitHub display).

src/client/helpers/entity.js
```diff
@@ -0,0 +1,127 @@
...
+function showEntityEditions(entity) {
+	'use strict';
+	return (
+		<div>
+			<h2>
+				Editions
+				<Button
+					bsStyle="success"
+					className="pull-right"
+					href={`/edition/create?publication=${entity.bbid}`}
```
LordSputnik: This link won't work if you call showEntityEditions and pass in a publisher edition - I'd suggest doing something equivalent to: `/edition/create?${entity.type.toLowerCase()}=${entity.bbid}`

src/client/helpers/entity.js
```diff
@@ -0,0 +1,128 @@
...
+function getTypeAttribute(entityType) {
+	'use strict';
+	return {title: 'Type', data: extractAttribute(entityType, 'label')};
+}
+
+function getDateAttributes(entity) {
+	'use strict';
+	const attributes = [{
+		title: 'Begin Date',
+		data: extractAttribute(formatDate(new Date(entity.beginDate)))
```
LordSputnik: Just noticed, if `entity.beginDate` is null, the formatted date will be 1970-01-01 - need to add a condition to prevent this.

LordSputnik: Actually, this is a partial date again, so don't do any formatting!

src/client/helpers/entity.js
```diff
@@ -0,0 +1,128 @@
...
+	}];
+	if (entity.ended) {
+		attributes.push({
+			title: 'End Date',
+			data: extractAttribute(formatDate(new Date(entity.endDate)))
```
LordSputnik: Also a partial date, no formatting needed.

src/client/components/pages/entity/edition.js
```diff
@@ -0,0 +1,99 @@
...
+					href={`/publisher/${publisher.bbid}`}
+					key={publisher.bbid}
+				>
+					{publisher.defaultAlias.name}
+				</a>
+			) : '?';
+	const releaseDate = (entity.releaseEventSet &&
+	entity.releaseEventSet.releaseEvents &&
+	entity.releaseEventSet.releaseEvents.length) ?
+		formatDate(new Date(entity.releaseEventSet.releaseEvents[0].date)) :
```
LordSputnik: This is a partial date, so no formatting needed.

src/client/helpers/entity.js
```diff
@@ -0,0 +1,128 @@
...
+							<td>
+								<a href={`/edition/${edition.bbid}`}>
+									{editionName}
+								</a>
+								<span className="text-muted">
+									{editionComment}
+								</span>
+							</td>
+							<td>
+								{formatDate(new Date(releaseEventDate))}
```
LordSputnik: This will be a partial date, so don't format.

src/client/components/pages/entity/publication.js
```diff
@@ -0,0 +1,55 @@
...
+		<EntityPage
+			attributes={attributes}
+			iconName="th-list"
+			{...extractEntityProps(props)}
+		>
+			{showEntityEditions(entity)}
+		</EntityPage>
+	);
+}
+PublicationPage.getAttributes = (entity) => ([getTypeAttribute(entity)]);
```
LordSputnik: This needs to be `getTypeAttribute(entity.publicationType)`

src/client/components/pages/entity/edition.js
```diff
@@ -0,0 +1,99 @@
...
+	const dimensions = `${width}×${height}×${depth} mm`;
+	return [
+		{title: 'Status', data: editionStatus},
+		{title: 'Format', data: editionFormat},
+		getLanguageAttribute(entity),
+		{title: 'Publishers', data: publishers},
+		{title: 'Release Date', data: releaseDate},
+		{title: 'Page Count', data: pageCount},
+		{title: 'Weight', data: weight},
+		{title: 'Dimensions', data: dimensions}
```
LordSputnik: Should be `'Dimensions (W×H×D)'`

src/client/components/pages/entity/edition.js
```diff
@@ -0,0 +1,99 @@
...
+					{publisher.defaultAlias.name}
+				</a>
+			) : '?';
+	const releaseDate = (entity.releaseEventSet &&
+	entity.releaseEventSet.releaseEvents &&
+	entity.releaseEventSet.releaseEvents.length) ?
+		entity.releaseEventSet.releaseEvents[0].date :
+		'?';
+	const pageCount = extractAttribute(entity.pages);
+	const weight = extractAttribute(entity.weight);
```
LordSputnik: Should be `${extractAttribute(entity.weight)} g`

PR #137, opened 2017-01-19
Fix bug tracker link

anshuman73:
### Problem

Currently, "Report a Bug" button in the footer doesn't link to MeB's bug tracker.

### Solution

The fix was mostly simple, adding a link to the Jira Bug Tracker.

### Areas of Impact

All webpages should now link to the bug tracker


Arthelon:
Some pages are currently being rendered primarily with React. This change will need to be added to the Footer React component in src/client/components/footer.js

anshuman73:
Ahh, didn't realise that. will fix up in a sec.

PR #139, opened 2017-01-22
Add documentation to PublisherEditor component

rioc0719:
### Problem
Adding documentation to files in the src/client/components/forms/ directory

### Solution
Adds documentation to the publisher.js file.

### Areas of Impact
No behavior has been changed.

### DIFF
--- a/src/client/components/forms/publisher.js
+++ b/src/client/components/forms/publisher.js
@@ -37,7 +37,7 @@ const RevisionNote = require('./parts/revision-note');
  */
 
 class PublisherForm extends React.Component {
-	
+
 	/**
 	 * Binds class methods to their respective data.
 	 * @param {object} props - Properties passed into the constructor
--- a/src/client/components/forms/publisher.js
+++ b/src/client/components/forms/publisher.js
@@ -16,6 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
 
 const Icon = require('react-fontawesome');
 const React = require('react');
@@ -29,7 +30,18 @@ const LoadingSpinner = require('../loading-spinner');
 const PublisherData = require('./parts/publisher-data');
 const RevisionNote = require('./parts/revision-note');
 
+/**
+ * React component to define a form to create and edit
+ * a Publisher.
+ * @extends React.Component
+ */
+
 class PublisherForm extends React.Component {
+	
+	/**
+	 * Binds class methods to their respective data.
+	 * @param {object} props - Properties passed into the constructor
+	 */
 	constructor(props) {
 		super(props);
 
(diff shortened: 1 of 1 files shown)


rioc0719:
I have a couple questions about differences I saw between publisher.js and work.js here:
* Should the handleBackClick and handleNextClick methods have an event.preventDefault() call?
* Should handleSubmit() have a `const self = this` line before the POST request (around line 120)?  

Thanks!

LordSputnik:
1. I don't think so - these are buttons or divs rather than anchors, so I don't think there's any associated behavior we want to suppress.
2. I think it's OK as it is - I don't see anywhere that `this` is going to be redefined here.

src/client/components/forms/publisher.js
```diff
@@ -29,7 +30,18 @@ const LoadingSpinner = require('../loading-spinner');
...
+/**
+ * React component to define a form to create and edit
+ * a Publisher.
+ * @extends React.Component
+ */
+
 class PublisherForm extends React.Component {
+
+	/**
+	 * Binds class methods to their respective data.
```
LordSputnik: I'm not too keen on this description. I think this would be better:

	/**
	 * Constructs a new instance of the PublisherForm class.
	 * This constructor calls the constructor of the React.Component class,
	 * initializes the component state and binds class methods to the class
	 * instance.
	 * @constructor
	 *
	 * @props {object} props - The properties of the component
	 **/

src/client/components/forms/publisher.js
```diff
@@ -55,21 +72,32 @@ class PublisherForm extends React.Component {
...
+	/**
+	 * Handler to direct the user to the next tab.
+	 */
 	handleNextClick() {
 		this.handleTabSelect(this.state.tab + 1);
 	}
 
+	/**
+	 * Handler to send publisher information in the form to the server.
+	 * @param {Event} evt - used to prevent event propagation
```
LordSputnik: `@param {Event} evt - The submission event that triggered the handler` would be better, I think.

src/client/components/forms/publisher.js
rioc0719: The JSDoc website says [here](http://usejsdoc.org/howto-es2015-classes.html) that `@constructor` doesn't need to be used with ES6 classes. Do you still want the tag included?

src/client/components/forms/publisher.js
```diff
@@ -47,6 +59,11 @@ class PublisherForm extends React.Component {
 		this.handleSubmit = this.handleSubmit.bind(this);
 	}
 
+	/**
+	 * Changes to the tab selected by the user, which has the number
+	 * in 'tab', and checks the validity of the alias and other data.
+	 * @param {number} tab - the number of the selected tab.
```
Arthelon: I think you can exclude the latter 2 portions of this description. Avoid mentioning behavior that is not included in this handler.

src/client/components/forms/publisher.js
LordSputnik: Ah yes I remember now, that's fine then.

PR #141, opened 2017-06-10
Upgrade to NPM version 5

Arthelon:
### Problem
Current npm installations are non-deterministic (can result in different dependency versions) and the script for copying dependencies into the project source folder is attached to the wrong npm script hook.

### Solution
- Adds a package-lock.json file to ensure deterministic installs with the npm client. 
- Replaces `npm prepublish` with `npm prepublishOnly` and `npm postinstall` (prevents case where node_modules is empty). 
- Improved logical separation of dependencies by moving build dependencies into the devDependencies tree

### Areas of Impact
- package.json
- package-lock.json

Should still be compatible with older npm clients


LordSputnik:
Why move babel to devDependencies when it's required for running the site in production?

Arthelon:
Babel should only be required for running the production build. This separation of dependencies made sense to me.

LordSputnik:
Hmm, I'd like to keep it/make it so that `npm install --production` installs everything needed to build and run the application in the production environment. And then have anything that's *just* for development or testing in devDependencies.

PR #143, opened 2017-10-01
Linting cleanup

leftmostcat:
### Problem

Lots of linting warnings and errors.

### Solution

Partial sweep fixing lint.

### Areas of Impact

The only functional change is in displaying recent edits.


src/client/components/forms/profile.js
```diff
@@ -98,7 +98,7 @@ class ProfileForm extends React.Component {
 			name: gender.name
 		}));
 		const titleOptions = this.state.titles.map((unlock) => {
-			const title = unlock.title;
+			const {title} = unlock.title;
```
LordSputnik: Slight issue here - should be `const {title} = unlock;`

PR #145, opened 2017-11-26
Fixes Edit Page error

agbilotia1998:
### Problem
Fixes BB-226 (Edit option redirecting to page throwing http error 500).

### Solution
Added condition to check if Publisher Set  is undefined.

### Areas of Impact
Changes src/server/routes/entity/edition.js


LordSputnik:
Hi,

I can't see what this code changes functionally...

    _.isUndefined(edition.publisherSet)

...is already covered by...

    edition.publisherSet

... since an undefined value is falsey.

Did you experience this bug yourself? If not, it may already have been fixed since it was reported.

Thanks for contributing though :+1:

LordSputnik:
OK, I just saw that it was you that made the JIRA ticket, so I'm investigating this now.

agbilotia1998:
@LordSputnik edition.publisherSet was false that's why it remained undefined and error is thrown when editionSection object  is created.Do you require something else here?

LordSputnik:
The intent of the current code is:

1  Check whether edition.publisherSet is truthy (not null/undefined)
2. If it is, check whether edition.publisherSet.publishers is not empty
3. If there are publishers, get the first publisher. If there aren't, return null.

However, due to operator precedence, what was actually happening was:

1. Check that both edition.publisherSet is truthy and edition.publisherSet.publishers is not empty
2. If both conditions above are true, return null
3. If they're both false, try accessing the first publisher (which will always be undefined)

I've fixed this in f669abd61cb4dccb756163ca9d8a1c1386de16f7

PR #148, opened 2017-11-29
fix: mobile page width now scales with device width

falafel72:
### Problem
Webpage did not adjust size responsively for mobile. 

### Solution
Added meta tag to scale mobile page width to the device width.


Arthelon:
Looks good, could you also add `initial-scale=1` to the content value of the meta tag? This way the zoom level on page load will be constant. (separate the `width` and `initial-scale` key-value pairs with commas)

falafel72:
Just added, should be ok now.

Arthelon:
I'll approve your task once you submit it for review again. @LordSputnik this looks ready to merge.

PR #153, opened 2017-12-04
refactor(tests): Eliminate some code duplication

Androbin:
### Problem
Reduce the amount of code that needs to be maintained for testing.

### Solution
Factor out common assertion pattern applied to `achievementPromise`.

### Areas of Impact
This affects most `.js` files directly in the `test/` directory and doesn't change any behaviour.

### DIFF
--- a/test/test-creator-creator.js
+++ b/test/test-creator-creator.js
@@ -17,6 +17,7 @@
  */
 
 import * as testData from '../data/test-data.js';
+import {expectAchievementIds, expectAchievementIdsNested} from './common';
 import Promise from 'bluebird';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
@@ -55,14 +56,11 @@ export default function tests() {
 					edit.creatorCreator['Creator Creator I']
 				);
 
-			return Promise.all([
-				expect(achievementPromise).to.eventually.have
-					.property('editorId',
-						testData.editorAttribs.id),
-				expect(achievementPromise).to.eventually.have
-					.property('achievementId',
-						testData.creatorCreatorIAttribs.id)
-			]);
+			return expectAchievementIds(
+				achievementPromise,
+				testData.editorAttribs.id,
+				testData.creatorCreatorIAttribs.id
+			);
 		}
 	);
 
@@ -82,14 +80,11 @@ export default function tests() {
 					edit.creatorCreator['Creator Creator II']
 				);
 
-			return Promise.all([
-				expect(achievementPromise).to.eventually.have
-					.property('editorId',
-						testData.editorAttribs.id),
-				expect(achievementPromise).to.eventually.have
-					.property('achievementId',
-						testData.creatorCreatorIIAttribs.id)
-			]);
+			return expectAchievementIds(
+				achievementPromise,
+				testData.editorAttribs.id,
+				testData.creatorCreatorIIAttribs.id
+			);
 		});
 
 	it('III should be given to someone with 100 creator creations',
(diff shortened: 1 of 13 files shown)


test/common-util.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+
+
+chai.use(chaiAsPromised);
+const {expect} = chai;
+
+export function expectAchievementProp(achievement, editor, key, value) {
```
leftmostcat: It's a little strange to have this factored out of `expectAchievementId()`, as it doesn't seem to be used anywhere else.

test/common-util.js
```diff
@@ -0,0 +1,37 @@
...
+	return Promise.all([
+		expect(achievement).to.eventually.have
+			.property('editorId', editor),
+		expect(achievement).to.eventually.have
+			.property(key, value)
+	]);
+}
+
+export function expectAchievementId(achievement, editor, id) {
```
leftmostcat: Maybe call this `expectAchievementIds()`, since it's checking two different IDs.

test/common-util.js
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (C) 2016  Max Prettyjohns
```
leftmostcat: Your name should probably go here, since you're doing the refactor.

test/common-util.js
Arthelon: Agreed. I'd also change the argument names to `promise`, `editorId`, and `achievementId` for additional clarity.

test/test-creator-creator.js
```diff
@@ -16,6 +16,7 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+import * as commonUtil from './common-util';
```
Arthelon: Doesn't make much difference performance-wise but it's a good practice to import only what you need so I'd refrain from importing the entire `common-util` module namespace since you're only using a single method. Use a named import statement instead.

test/common-util.js
Androbin: I was thinking about factoring out the `.nested` blocks as well and wanted to keep things consistently clean and reusable.

test/common-util.js
Androbin: Okay

Androbin:
I implemented the requested changes and eliminated one more function.

Androbin:
Let me know if further changes are desired.

LordSputnik:
This looks good to me, as far as I can see you've addressed all of Leftmost's and Artherlon's requests. So I'm going to approve now.

This is a bit of an extension, if you're interested in doing a follow-up task. If you are interested, let me know here and I'll write it up and create it in the GCI site...

I'm wondering if we can expand the scope of this change. Pretty much all of these achievement tests follow this pattern:

1. Rewire some properties of the achievement module to allow for easier testing
2. Create an editor
3. Call a function that's expected to create an achievement
4. Check that the resulting achievement object has a key with the achievement name, with values for editor ID and achievement ID
5. Sometimes check that the resulting achievement object has a key with the badge name, with values for editor ID and title ID

 The only things that really change here are:

1. The properties of Achievement to rewire
2. The function to be called to generate the achievement
3. The achievement and badge names

So we should be able to write a function that takes those as parameters, and either

a) returns a function that'll test a particular achievement
b) test the achivement

Androbin:
@LordSputnik I'd be happy to implement those changes.

Androbin:
It might be a good idea to create a standalone issue ticket for this.

Androbin:
I am already working on it. Just keep me updated on the task.

Androbin:
The various tests actually differ (file-wise) in unpredictable ways. But I can certainly de-duplicate each files tests separately and unify them later.

PR #154, opened 2017-12-05
GCI: Remove A Jade File (favicon) (TheJYKoder)

TheJYKoder:


templates/target.js
```diff
@@ -0,0 +1,44 @@
+export default ({markup, script, page, props}) => {
+  return `
+    <!DOCTYPE html>
+    <html>
+      <head>
+    	<title>BookBrainz – The Open Book Database</title>
```
Arthelon: Interpolate a [ternary expression](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) so that the contents of the `<title>` tags will become `YOUR_TITLE – The Open Book Database` if the `title` variable exists. Otherwise, the title will remain `BookBrainz – The Open Book Database`

templates/target.js
```diff
@@ -0,0 +1,44 @@
+export default ({markup, script, page, props}) => {
```
Arthelon: You're forgetting the `title` function argument.

templates/target.js
```diff
@@ -0,0 +1,44 @@
...
+        <meta name="application-name" content="BookBrainz">
+        <meta name="msapplication-TileColor" content="#da532c">
+        <meta name="msapplication-TileImage" content="/images/icons/mstile-144x144.png">
+        <meta name="theme-color" content="#754e37">
+      </head>
+      <body>
+    	<div id="target">${markup}</div>
+    	<script src="/js/bundle.js"></script>
+    	<script type="application/json" id="page">${page}</script>
+    	<script type="application/json" id="props">${props}</script>
```
Arthelon: Same issue as the title tags. Add ternary expressions here so that these scripts are only added if certain conditions are fulfilled (refer to template.jade to see the conditions).

Arthelon:
@LordSputnik this can be closed. Another student has completed the task.

PR #157, opened 2017-12-07
Fix(Unsorted Editions):Sorted Editions according to date.

agbilotia1998:
bb publications
Fixes BB-228(Sort Editions)

Added SortByDate function in order to sort the editions according to release date.

### DIFF
--- a/src/server/routes/entity/publisher.js
+++ b/src/server/routes/entity/publisher.js
@@ -55,6 +55,7 @@ function _setPublisherTitle(res) {
 	);
 }
 
+
 router.get('/:bbid', middleware.loadEntityRelationships, (req, res, next) => {
 	// Fetch editions
 	const {Publisher} = req.app.locals.orm;
@@ -69,6 +70,7 @@ router.get('/:bbid', middleware.loadEntityRelationships, (req, res, next) => {
 		.then((editions) => {
 			res.locals.entity.editions = editions.toJSON();
 			_setPublisherTitle(res);
+			res.locals.entity.editions.sort(entityRoutes.compareEntitiesByDate);
 			entityRoutes.displayEntity(req, res);
 		})
 		.catch(next);
--- a/src/server/routes/entity/entity.js
+++ b/src/server/routes/entity/entity.js
@@ -954,13 +954,33 @@ export function areaToOption(area) {
 	if (!area) {
 		return null;
 	}
-
 	const {id} = area;
-
 	return {
 		disambiguation: area.comment,
 		id,
 		text: area.name,
 		type: 'area'
 	};
+}
+
+export function compareEntitiesByDate(a, b) {
+	const aDate = _.get(a, 'releaseEventSet.releaseEvents[0].date', null);
+	const bDate = _.get(b, 'releaseEventSet.releaseEvents[0].date', null);
+	if (_.isNull(aDate)) {
+		/*
+		 * return a positive value,
+		 * so that non-null dates always come before null dates.
+		 */
+		return 1;
+	}
+
+	if (_.isNull(bDate)) {
+		/*
+		 * return a negative value,
+		 * so that non-null dates always come before null dates.
+		 */
+		return -1;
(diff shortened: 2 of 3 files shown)


agbilotia1998:
@LordSputnik please review.

LordSputnik:
Hi @agbilotia1998, thanks for your contribution. This change is going to get applied to all entities, even if they don't have editions. Please could you move it to the routes where editions need to be sorted (publisher and publication display)? Please also follow [these guidelines](http://karma-runner.github.io/1.0/dev/git-commit-msg.html) for commit messages, and check your code with `npm run test`.

agbilotia1998:
@LordSputnik done please review.

src/server/routes/entity/publisher.js
```diff
@@ -58,6 +58,16 @@ function _setPublisherTitle(res) {
 	);
 }
 
+/**
+ * @param {object} a The first object.
+ * @param {object} b The second object.
+ * @returns {number} The sum of the two numbers.
```
LordSputnik: Nope - this is the difference. Also, you haven't mentioned the numbers earlier in the comment. Please format like this:

    /**
     * description goes here
     *
     * @param ...
     * @param ...
     * @returns ...
     **/

LordSputnik: Also, might be worth mentioning that the objects being passed in represent editions - something like "@param a The first edition to compare."

src/server/routes/entity/publisher.js
```diff
@@ -58,6 +58,16 @@ function _setPublisherTitle(res) {
...
+ */
+function SortByDate(a, b) {
```
LordSputnik: Please use camelCase for function names - make sure you're running "npm run lint" prior to comitting.

agbilotia1998: @LordSputnik isn't it already in CamelCase?

agbilotia1998:
@LordSputnik please review.

src/server/routes/entity/publisher.js
Arthelon: The camelCase standard used in JavaScript is to have an initial lower case letter instead of an upper case letter.

agbilotia1998:
@LordSputnik please review.

src/server/routes/entity/publisher.js
```diff
@@ -58,6 +58,17 @@ function _setPublisherTitle(res) {
...
 }
 
+/**
+ * Returns positive negative or zero value according to a and b.
+ * @param {object} a The first edition to compare.
+ * @param {object} b The second edition to compare.
+ * @returns {number} .
+ */
+function sortByDate(a, b) {
+	return new Date(a.releaseEventSet.releaseEvents[0].date).getTime() -
```
LordSputnik: You need to check whether a.releaseEventSet and b.releaseEventSet are null, and then whether releaseEvents has any elements. I'd suggest using:

    function compateEntitiesByDate(a, b) {
        const aDate = _.get(a, 'releaseEventSet.releaseEvents[0].date', null);
        const bDate = _.get(b, 'releaseEventSet.releaseEvents[0].date', null);
        if (_.isNull(aDate)) {
            // return a positive value, so that non-null dates always come before null dates.
            return 1; 
        }

        if (_.isNull(bDate)) {
            // return a negative value, so that non-null dates always come before null dates.
            return -1; 
        }

        return new Date(aDate).getTime() - new Date(bDate).getTime();
    }

Also, please place this in a common file (maybe `entity/entity.js`), and then import it to `entity/publisher.js` and `entity/publication.js`.

agbilotia1998:
@LordSputnik done please review.

LordSputnik:
This looks pretty good now. Please revert the changes to package-lock.json - then I should be able to merge this. Thanks!

agbilotia1998:
@LordSputnik done please review!

agbilotia1998:
@LordSputnik please review this PR.

PR #158, opened 2017-12-08
refactor(tests): Rewrite achievement tests

Androbin:
### Problem
Reduce the amount of code that needs to be maintained for testing.

### Solution
Factor out common assertion patterns

### Areas of Impact
This affects most `.js` files directly in the `test/` directory and doesn't change any behaviour.

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -76,3 +76,11 @@ export function rewireTypeCreation(Achievement, name, threshold) {
 			)
 	});
 }
+
+export function testAchievement(rewiring, generator, expectations) {
+	return () => {
+		rewiring();
+		const promise = generator();
+		return expectations(promise);
+	};
+}
--- a/test/common.js
+++ b/test/common.js
@@ -50,3 +50,16 @@ export function expectIdsNested(name, prop, rev) {
 				testData[`${prop}Attribs`].id)
 	]);
 }
+
+export function generateProcessEdit(Achievement, orm, func, name, rev) {
+	return () => testData.createEditor()
+		.then((editor) => Achievement.processEdit(orm, editor.id))
+		.then((edit) => edit[func][`${name} ${rev}`]);
+}
+
+export function generateProcessEditNamed(Achievement, orm, func, name) {
+	return () => new orm.Editor({name: testData.editorAttribs.name})
+		.fetch()
+		.then((editor) => Achievement.processEdit(orm, editor.id))
+		.then((edit) => edit[func][name]);
+}
--- a/test/common.js
+++ b/test/common.js
@@ -63,3 +63,16 @@ export function generateProcessEditNamed(Achievement, orm, func, name) {
 		.then((editor) => Achievement.processEdit(orm, editor.id))
 		.then((edit) => edit[func][name]);
 }
+
+export function rewireAchievement(Achievement, rewiring) {
+	return () => Achievement.__set__(rewiring);
+}
+
+export function rewireTypeCreation(Achievement, name, threshold) {
+	return rewireAchievement(Achievement, {
+		getTypeCreation:
+			testData.typeCreationHelper(
+				`${name}_revision`, threshold
+			)
+	});
+}
--- a/test/common.js
+++ b/test/common.js
@@ -16,6 +16,8 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+/* eslint import/namespace: ['error', { allowComputed: true }] */
+import * as testData from '../data/test-data.js';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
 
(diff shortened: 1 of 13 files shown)


Androbin:
@LordSputnik I am doing my best to modularize this as best as I can.

test/common.js
```diff
@@ -45,3 +59,39 @@ export function expectAchievementIdsNested(promise, name,
...
+	return () => rewireAchievement({
+		getTypeCreation:
+			testData.typeCreationHelper(
+				`${name}_revision`, threshold
+			)
+	});
+}
+
+export function testAchievement(rewiring, generator, expectations) {
+	return () => {
```
LordSputnik: This probably shouldn't return a function. I'd expect it to just be:

    testAchievement(...) {
        rewiring();
        const promise = generator();
        expectations(promise);
    }

test/common.js
```diff
@@ -45,3 +59,39 @@ export function expectAchievementIdsNested(promise, name,
...
+		)
+		.then((edit) => edit[func][name]);
+}
+
+export function rewireAchievement(rewiring) {
+	return () => Achievement.__set__(rewiring);
+}
+
+export function rewireTypeCreation(name, threshold) {
+	return () => rewireAchievement({
```
LordSputnik: This shouldn't return a function.

test/common.js
```diff
@@ -45,3 +59,39 @@ export function expectAchievementIdsNested(promise, name,
...
+	return () => new Editor({name: testData.editorAttribs.name})
+		.fetch()
+		.then((editor) =>
+			Achievement.processEdit(orm, editor.id)
+		)
+		.then((edit) => edit[func][name]);
+}
+
+export function rewireAchievement(rewiring) {
+	return () => Achievement.__set__(rewiring);
```
LordSputnik: This shouldn't return a function. Also, better to not use globals, so I'd say pass in the `Achievement` object and then change it within the function (and go back to calling `const Achievement = rewire(...)` in each Achievement module.

There may be unintended side effects to having a shared Achievement variable (not being cleaned up properly after each test, for example).

test/common.js
```diff
@@ -45,3 +59,39 @@ export function expectAchievementIdsNested(promise, name,
 			.property(`${name}.titleId`, titleId)
 	]);
 }
+
+export function generateProcessEdit(func, name, rev) {
+	return () => testData.createEditor()
```
LordSputnik: Shouldn't return a function.

test/common.js
```diff
@@ -45,3 +59,39 @@ export function expectAchievementIdsNested(promise, name,
...
+		.then((editor) => Achievement.processEdit(orm, editor.id))
+		.then((edit) => edit[func][`${name} ${rev}`]);
+}
+
+export function generateProcessEditNamed(func, name) {
+	return () => new Editor({name: testData.editorAttribs.name})
```
LordSputnik: Shouldn't return a function.

Androbin:
Since all those functions in `common.js` are to be passed as a function, I thought it would be less boilerplate after all.

test/common.js
Androbin: What about changing this to use the `__with__` method?

test/common.js
Androbin: > So, we'd like you to write a function that takes those as parameters, and either
> a) returns a function that'll test a particular achievement
> b) test the achievement itself

LordSputnik: OK, so if you want to take that approach, you'll need to do something like:

    const test1 = testAchievement(a, b, c); // at this point test1 is a function taking no parameters generated by testAchievement
    it('...', test1);

Androbin: Shall I disassemble the parameters in the same way?

Androbin:
@LordSputnik Travis gives error messages I don't understand.

Androbin:
@LordSputnik Can you tell what this error message means?
`Unhandled rejection error: duplicate key value violates unique constraint "editor_type_pkey"`

Androbin:
Some quick research suggests that the database got "out of sync".

Androbin:
@LordSputnik I have problems dealing with this CI error:
`AssertionError: expected false to have property 'editorId'`
I suspect that some value isn't passed properly all the way.

Androbin:
@LordSputnik Failing `master` is blocking the CI tests on this PR.
PS. Yeah, I'm sorry, it fails to run locally.

Androbin:
I am experimenting a bit with passing deps to common instead of importing from there.

LordSputnik:
Hi @Androbin,

I think it would probably be quicker to start from scratch and make this change in small steps. Transform the code in one way per commit, check the tests work and then make the next step. If the tests don't run successfully locally with the current master, let me know and we can try to fix it.

Androbin:
@LordSputnik Okay, I will split into more granular commits.

Androbin:
I have parked the original patch set on [branch refactor](https://github.com/Androbin/bookbrainz-site/tree/refactor).

Androbin:
Status Quo:
- `common.js` looks nearly like before.
- tests are split into parts again

Yet To Do:
- de-duplicate file-wise
- refactor more customizations

Androbin:
Also, I'm thinking about passing `Achievement` and/or `orm` to `testAchievement`, not earlier.

Androbin:
@LordSputnik Turns out that tests fail if the rewiring is reverted after each test.

Androbin:
@LordSputnik Any objections?

PR #160, opened 2017-12-09
Fix(Duplicate Alias):Added check if Alias is duplicated

agbilotia1998:
Fixes: Duplicate Aliases problem.

Error is thrown at time of form submission if alias is duplicated.

A check function is added to check if added alias is already present in AliasSet.

### DIFF
--- a/src/server/helpers/entityRouteUtils.js
+++ b/src/server/helpers/entityRouteUtils.js
@@ -180,7 +180,18 @@ export function makeEntityCreateOrEditHandler(
 			const err = new error.FormSubmissionError();
 			error.sendErrorAsJSON(res, err);
 		}
-
+		if (req.body.aliasEditor !== null) {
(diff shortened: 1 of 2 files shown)


agbilotia1998:
@LordSputnik @leftmostcat please review.

LordSputnik:
Hi Ayush, we currently don't have a lot of time to review contributions to BookBrainz because we're mentoring for Google Code In. But we do appreciate the work you're doing and will look at it as soon as possible. Thanks!

LordSputnik:
Just took a quick look. I don't think we want this check to cause an error, because there are many valid cases where an entity will have the same alias as another entity. It would be better to warn the user that the entity they're adding might already be in the database.

To do that, I think the best thing to do would be to update the "NameSection" component to do an async request to the server search with the name the user entered.

agbilotia1998:
Okay I will do that.

PR #161, opened 2017-12-11
fix(registration): fix CI error from 0930e7d

Androbin:
### Problem
Make the CI tests pass.

### Solution
Revert changes mistakenly not reverted in 0930e7d.

### Areas of Impact
Only `src/client/components/forms/registration-details.js` is affected.
No behaviour changes.


LordSputnik:
Looks good, thanks!

PR #162, opened 2017-12-12
update init function to address requestTimeOut Failure on reindexing

akhilesh26:
### Problem
When reindexing the data in elasticsearch it is giving error request timeout.

### Solution
Increase the value of requestTimeOut.

### Areas of Impact
When we build the project and we have to reindex the data in the elasticsearch. The default requestTimeOut is 30000ms which is become short for indexing current dumped database.


LordSputnik:
Think this is no longer needed with the new bulk indexing code.

PR #165, opened 2017-12-15
Add eslint precommit hook

naiveaiguy:
### Problem

Linter errors aren't being caught when committing.

### Solution

Adds a pre-commit hook that runs `eslint` on all staged `*.js` files, and does
not allow you to commit until errors are fixed.

### Areas of Impact

Only the `package.json` (and of course the `package-lock.json`) is affected.

By the way, the branch is named `precommit-eshansingh` because my GCI username is Eshan Singh, as it's forced to be my real name, but my GH username is different.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,14 @@
     "test-cov": "nyc --reporter=text npm run test",
     "test-ci": "nyc --reporter=lcovonly --reporter=text-summary npm run test",
     "jsdoc": "node_modules/.bin/jsdoc -r src",
-    "flow": "flow"
+    "flow": "flow",
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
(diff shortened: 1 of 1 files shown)


naiveaiguy:
Just realized, some weird version bump stuff happened when I did an `npm install` to check this. Should I remove the `package-lock.json` file? From what it looks like it's automatically generated anyway.

Arthelon:
Looks good to me. Try adding a [.gitattributes](https://git-scm.com/docs/gitattributes) file and flagging `package-lock.json` as a binary to resolve the conflict you're getting.

Arthelon:
@LordSputnik

LordSputnik:
Thanks @naiveaiguy ! :+1:

PR #166, opened 2017-12-16
GCI: Replace BookBrainz jade template (oswinso)

oswinso:


src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+		  <link rel="stylesheet" href='/stylesheets/react-virtualized.css' />
+		  <link rel="stylesheet"
+		    href='/stylesheets/react-virtualized-select.css' />
+		  <meta name='viewport' content='//react-virtualized.css' />
+		  ${favicon}
+		</head>
+
+		<body>
+		  <div id="target">${markup}</div>
+		  <script src='/js/bundle.js' />
```
Arthelon: Both the starting and ending tags are [required for the script element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script).

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+		  ${page ?
```
Arthelon: It'd be clearer to use a logical `&&` here instead of a ternary expression since the falsy condition is empty. Do the same for the other ternary expression below.

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+		    `<script id='page' type='application/json' >${page}</script>` : ''}
+		  ${props && script ?
+		    `<script id='props' type='application/json'> ${props}</script>
+			<script src='${script} />` : ''}
```
Arthelon: Same issue as the previous script tag. Also try to keep indentation consistent.

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+	  <!DOCTYPE html>
+	  <html>
+		<head>
+		  <title>${title ? `'${title} – BookBrainz'` :
+		     'BookBrainz – The Open Book Database'}</title>
+		  <link rel="stylesheet" href='/stylesheets/style.css' />
+		  <link rel="stylesheet" href='/stylesheets/react-virtualized.css' />
+		  <link rel="stylesheet"
+		    href='/stylesheets/react-virtualized-select.css' />
+		  <meta name='viewport' content='//react-virtualized.css' />
```
Arthelon: Please fix the `content` value for this meta tag. Your use of quotations is also inconsistent, please stick to single-quotes since it is enforced in the rest of the project.

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
+import favicon from './favicon';
+
+
+export default ({title, markup, page, props, script}) => `
+	  <!DOCTYPE html>
+	  <html>
+		<head>
+		  <title>${title ? `'${title} – BookBrainz'` :
```
Arthelon: Please remove the single quotes nested inside the inner template string.

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+		  ${favicon}
+		</head>
+
+		<body>
+		  <div id="target">${markup}</div>
+		  <script src='/js/bundle.js' />
+		  ${page ?
+		    `<script id='page' type='application/json' >${page}</script>` : ''}
+		  ${props && script ?
+		    `<script id='props' type='application/json'> ${props}</script>
```
Arthelon: unnecessary whitespace

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+		    href='/stylesheets/react-virtualized-select.css' />
+		  <meta name='viewport' content='//react-virtualized.css' />
+		  ${favicon}
+		</head>
+
+		<body>
+		  <div id="target">${markup}</div>
+		  <script src='/js/bundle.js' />
+		  ${page ?
+		    `<script id='page' type='application/json' >${page}</script>` : ''}
```
Arthelon: unnecessary whitespace after `type` attribute

src/server/templates/target.js
```diff
@@ -0,0 +1,28 @@
...
+	  <html>
+		<head>
+		  <title>${title ? `'${title} – BookBrainz'` :
+		     'BookBrainz – The Open Book Database'}</title>
+		  <link rel="stylesheet" href='/stylesheets/style.css' />
+		  <link rel="stylesheet" href='/stylesheets/react-virtualized.css' />
+		  <link rel="stylesheet"
+		    href='/stylesheets/react-virtualized-select.css' />
+		  <meta name='viewport' content='//react-virtualized.css' />
+		  ${favicon}
```
Arthelon: `favicon` is imported as a function that returns a template string but it is never invoked. I recommend combining `favicon.js` into `target.js` and adding it as a constant for the time being.

oswinso:
@Arthelon Should I be using the custom express template engine here for the GCI task?

oswinso:
@Arthelon Styling errors fixed with d07bf14

src/server/templates/favicon.js
```diff
@@ -0,0 +1,40 @@
+export default () => `
```
Arthelon: This file is unused and can be deleted.

src/server/templates/target.js
```diff
@@ -0,0 +1,71 @@
+const favicon = `
+	<link rel="apple-touch-icon" sizes="57x57"
```
Arthelon: replace all double-quotes here with single quotes as per the previous review.

src/server/templates/target.js
```diff
@@ -0,0 +1,71 @@
...
+	title,
+	markup,
+	page,
+	props,
+	script
+}) => `
+	<!DOCTYPE html>
+	<html>
+		<head>
+		<title>${title ? `${title} – BookBrainz` :
```
Arthelon: indent the following 3 lines one level outwards.

src/server/templates/target.js
```diff
@@ -0,0 +1,71 @@
...
+		  <link rel='stylesheet'
+		    href='/stylesheets/react-virtualized-select.css' />
+		  <meta name='viewport' content='width=device-width, initial-scale=1' />
+		  ${favicon}
+		</head>
+
+		<body>
+			<div id='target'>${markup}</div>
+			<script src='/js/bundle.js'></script>
+			${page && script ?
```
Arthelon: The conditional logic here doesn't match the jade template. It should look something more like this:

Arthelon: The conditional logic here doesn't match the jade template. It should look something more like this (you shouldn't need to use a ternary expression):

```
${page && ...}
${props && script && ...}
```

oswinso:
@Arthelon Fixed again with d0de250

Sorry...

Arthelon:
Looks good!

PR #167, opened 2017-12-16
Dockerize

leftmostcat:
### Problem

There is no configuration for deployment with docker.

### Solution

Create configuration for deployment with docker.

### Areas of Impact

Configuration loading, newly-created docker assets.


Dockerfile
```diff
@@ -0,0 +1,48 @@
+FROM metabrainz/consul-template-base:v0.18.5-1
+
+ARG BUILD_DEPS=" \
```
mwiencek: The reason I originally separated these from `RUN_DEPS` was to uninstall them from the image after they weren't needed anymore, e.g.: https://github.com/metabrainz/CAA-indexer/blob/a8a27df1b1e2f8961b2e0cdae63231b552dd22e2/Dockerfile#L44-L44 though it's not really necessary, just reduces the image size a bit.

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+ARG RUN_DEPS=" \
+    nodejs"
+
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+RUN apt-get update
+RUN apt-get install --no-install-suggests --no-install-recommends -y \
+        $BUILD_DEPS \
+        $RUN_DEPS
+RUN rm -rf /var/lib/apt/lists/*
```
mwiencek: You should put the `apt-get update` and `apt-get install` in the same `RUN` statement, because each `RUN` creates an intermediary (cached) image layer. So if you change `$RUN_DEPS`, it will use the old, cached `apt-get update` layer, which may be outdated at that point.

mwiencek: I would also combine the `rm -rf /var/lib/apt/lists/*` into that same `RUN` statement too, since it's also good to reduce unnecessary image layers where they don't need to be cached separately (say because they are expensive).

docker/bookbrainz
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cd /home/bookbrainz/bookbrainz-site
+export NODE_ENV="production"
+node ./lib/server/app.js
```
mwiencek: `exec node`? Otherwise signals will go to the shell

docker/consul-template-bookbrainz.conf
```diff
@@ -0,0 +1,11 @@
+template {
+    source = "/home/bookbrainz/bookbrainz-site/config.json.ctmpl"
+    destination = "/home/bookbrainz/bookbrainz-site/config/config.json"
+}
+
+exec {
+    command = "/usr/local/bin/bookbrainz"
+    reload_signal = ""
```
mwiencek: Setting this to nil means it'll restart the process every time the template changes, causing a small amount of downtime. Which you might be aware of, but just in case :)

Dockerfile
leftmostcat: Ahh, yep. Meant to remove this; I'll address this.

Dockerfile
leftmostcat: Good to know, will address.

docker/consul-template-bookbrainz.conf
leftmostcat: If there's a change, that's the only way we can get up-to-date. I'd like to be able to respond to SIGHUP, but that will take some work.

.dockerignore
```diff
@@ -0,0 +1,22 @@
...
+.editorconfig
+.eslintignore
+.eslintrc.js
+.flowconfig
+.gitignore
+.gitmodules
+.tern-project
+.travis-yml
+LICENSE
+package-lock.json
```
mwiencek: I think you'll want this copied into the image, otherwise `npm install` will get a slightly different dependency tree every time you build the image

docker/consul-template-bookbrainz.conf
```diff
@@ -0,0 +1,10 @@
+template {
+    source = "/home/bookbrainz/bookbrainz-site/config.json.ctmpl"
+    destination = "/home/bookbrainz/bookbrainz-site/config/config.json"
+}
+
+exec {
+    command = "/usr/local/bin/bookbrainz"
```
mwiencek: It'd be good to run node as non-root, e.g. the bookbrainz user, if you can. The `chpst` command is an easy way to drop privileges. Though you might have to adjust the permissions of the generated `config.json` after it's written too (since consul-template runs & writes stuff as root). See https://github.com/metabrainz/coverart_redirect/blob/master/docker/prod/consul-template-redirect.conf#L4-L8

src/server/helpers/config.js
```diff
@@ -16,8 +16,19 @@
...
-const env = process.env.NODE_ENV || 'development';
-const config = require(`../../../config/${env}`);
+const env = process.env.NODE_ENV || 'development'; // eslint-disable-line no-process-env,max-len
+
+let configFileBasename = 'config';
+if (env === 'test') {
+	configFileBasename = 'test';
+}
+
+const configContents =
```
LordSputnik: There must be a way that we can do this better... Some sort of asynchronous caching config loader module...

Something that we can use in places where config is consumed, like this:

    const config = await new Config();
    initLibrary(config.data.here);

LordSputnik:
This looks good to me 👍

PR #168, opened 2017-12-17
Use bulk search operations during reindexing

leftmostcat:
### Problem

Reindexing can flood elasticsearch's queue, causing the reindexing operations to halt.

### Solution

Use bulk operations to batch up indexing operations.

### Areas of Impact

Reindexing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -73,6 +73,7 @@ const bestPracticesRules = {
 	'accessor-pairs': ERROR,
 	'array-callback-return': ERROR,
 	'block-scoped-var': ERROR,
+	'class-methods-use-this': TRANSITION_IGNORE,
 	complexity: ERROR,
 	'consistent-return': ERROR,
 	curly: ERROR,
@@ -161,6 +162,7 @@ const strictModeRules = {
 };
 
 const variablesRules = {
+	'init-declarations': TRANSITION_IGNORE,
 	'no-catch-shadow': ERROR,
 	'no-label-var': ERROR,
 	'no-shadow': ERROR,
@@ -281,9 +283,18 @@ const stylisticIssuesRules = {
 			tabWidth: 4
 		}
 	],
+	'max-lines': TRANSITION_IGNORE,
 	'max-nested-callbacks': [
 		ERROR,
 		5
+	],
+	'max-params': [
+		TRANSITION_IGNORE,
+		4
+	],
+	'max-statements': [
+		TRANSITION_IGNORE,
+		15
 	],
 	'multiline-comment-style': ERROR,
 	'new-parens': ERROR,
@@ -298,6 +309,7 @@ const stylisticIssuesRules = {
 	],
 	'no-multiple-empty-lines': ERROR,
 	'no-nested-ternary': ERROR,
+	'no-new-object': TRANSITION_IGNORE,
 	'no-trailing-spaces': ERROR,
 	'no-unneeded-ternary': ERROR,
 	'no-whitespace-before-property': ERROR,
@@ -411,6 +423,7 @@ const reactRules = {
 	'react/boolean-prop-naming': ERROR,
 	'react/button-has-type': ERROR,
 	'react/default-props-match-prop-types': ERROR,
+	'react/forbid-component-props': TRANSITION_IGNORE,
 	'react/forbid-foreign-prop-types': ERROR,
 	'react/jsx-boolean-value': ERROR,
 	'react/jsx-closing-bracket-location': [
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -73,7 +73,6 @@ const bestPracticesRules = {
 	'accessor-pairs': ERROR,
 	'array-callback-return': ERROR,
 	'block-scoped-var': ERROR,
-	'class-methods-use-this': TRANSITION_WARNING,
```
leftmostcat: There are a bunch of places we don't do this and I'm not sure we need to enforce it.

.eslintrc.js
```diff
@@ -162,14 +161,13 @@ const strictModeRules = {
 };
 
 const variablesRules = {
-	'init-declarations': TRANSITION_WARNING,
```
leftmostcat: This floods us with warnings and doesn't always make sense. There are a number of places we declare a variable and then if/else to determine its value.

.eslintrc.js
```diff
@@ -162,14 +161,13 @@ const strictModeRules = {
...
 	'no-catch-shadow': ERROR,
 	'no-label-var': ERROR,
 	'no-shadow': ERROR,
 	'no-shadow-restricted-names': ERROR,
 	'no-undef-init': ERROR,
 	'no-undefined': ERROR,
-	'no-unused-vars': ERROR,
```
leftmostcat: Error seems too severe for something that's not really an issue except for code cleanliness.

.eslintrc.js
```diff
@@ -283,19 +281,10 @@ const stylisticIssuesRules = {
 			tabWidth: 4
 		}
 	],
-	'max-lines': TRANSITION_WARNING,
```
leftmostcat: This seems mostly aspirational; it can easily be added back if it's ever addressed, but I'm not sure it always makes sense.

.eslintrc.js
```diff
@@ -283,19 +281,10 @@ const stylisticIssuesRules = {
...
 	'max-nested-callbacks': [
 		ERROR,
 		5
 	],
-	'max-params': [
```
leftmostcat: Both of these are aspirational and don't necessarily always make sense.

.eslintrc.js
```diff
@@ -309,7 +298,6 @@ const stylisticIssuesRules = {
 	],
 	'no-multiple-empty-lines': ERROR,
 	'no-nested-ternary': ERROR,
-	'no-new-object': ERROR,
```
leftmostcat: I can see why we'd prefer to avoid it, but arrow functions returning new objects make this tricky.

.eslintrc.js
```diff
@@ -387,7 +375,7 @@ const ecmaScript6Rules = {
 	'no-var': ERROR,
 	'object-shorthand': ERROR,
 	'prefer-arrow-callback': ERROR,
-	'prefer-const': ERROR,
```
leftmostcat: This seems overly-severe when it's not really an issue.

.eslintrc.js
```diff
@@ -423,7 +411,6 @@ const reactRules = {
 	'react/boolean-prop-naming': ERROR,
 	'react/button-has-type': ERROR,
 	'react/default-props-match-prop-types': ERROR,
-	'react/forbid-component-props': TRANSITION_WARNING,
```
leftmostcat: This is definitely aspirational and floods the warnings. If we want to address this, we should turn this option on locally to work on it, but in the meantime it's just noise.

.eslintrc.js
```diff
@@ -482,7 +469,6 @@ const reactRules = {
 		}
 	],
 	'react/no-redundant-should-component-update': ERROR,
-	'react/no-set-state': TRANSITION_WARNING,
```
leftmostcat: Noise until we want to actually address it.

test/.eslintrc.js
```diff
@@ -5,5 +5,8 @@
 module.exports = {
 	env: {
 		mocha: true
+	},
+	rules: {
+		'no-unused-expressions': 0
```
leftmostcat: Without ignoring these, we get a ton of noise from tests.

src/server/helpers/search.js
```diff
@@ -63,6 +63,46 @@ function _searchForEntities(orm, dslQuery) {
 		.then((results) => _fetchEntityModelsForESResults(orm, results));
 }
 
+async function _bulkIndexEntities(entities) {
+	if (entities.length === 0) {
+		return;
+	}
+
+	const bulkOperations = entities.reduce((accumulator, entity) => {
```
LordSputnik: Isn't this also equivalent to `map` (reduce and push)?

leftmostcat: Since I'm pushing two elements per loop, I'm not sure I can use map.

PR #169, opened 2017-12-22
Filter identifier types by entity type

leftmostcat:
### Problem

We have provision in the database for making identifier types specific to entity type, but this is unused. As such, it is possible to set the ISBN of a Creator, which does not make sense.

### Solution

Before passing identifier types to the editor component, filter them by entity type. To allow users to edit existing identifiers which have been mistakenly added to the wrong entity type, the types of any existing identifiers are added on the edit page.

### Areas of Impact

Entity creation and editing.

**NOTE**: The following script should be run to migrate existing data to use appropriate entity types: https://gist.github.com/leftmostcat/b2d91955769831887b0b28f0f13db40c

### DIFF
--- a/src/server/helpers/utils.js
+++ b/src/server/helpers/utils.js
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2015  Ben Ockmore
- *               2015  Sean Burke
+ * Copyright (C) 2015       Ben Ockmore
+ *               2015-2017  Sean Burke
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


src/server/helpers/utils.js
```diff
@@ -174,8 +175,7 @@ export function incrementEditorEditCountById(orm, id, transacting) {
  * @returns {Promise} a promise which will be fulfilled when the operation to
  *                    truncate tables completes
  */
-export function truncateTables(Bookshelf, tables) {
-	return Promise.each(
```
LordSputnik: What's the reason for this change? I have a feeling "each" is needed for its sequential execution of promises, since the tables are often passed in in an order that minimises dependencies.

PR #170, opened 2017-12-23
Show no-session alert

leftmostcat:
### Problem

There's no indication on the site when redis isn't up.

### Solution

Add scaffolding to show alerts on the site, use this to display a message when we can't find a session.

### Areas of Impact

Page display.

### DIFF
--- a/src/client/components/pages/contribute.js
+++ b/src/client/components/pages/contribute.js
@@ -58,7 +58,7 @@ function ContributePage() {
 					BookBrainz style guidelines
 				</a> when adding data.
 				Unfortunately the aforementioned are still work in progress
-				and should they not cover some cases, please refer to the
+				and should they not cover some cases, please refer to the&nbsp;
 				<a href="http://musicbrainz.org/doc/Style">
 					MusicBrainz ones
 				</a>, adapting as necessary.
--- a/src/client/containers/layout.js
+++ b/src/client/containers/layout.js
@@ -191,14 +191,17 @@ class Layout extends React.Component {
 			</div>
 		);
 
-		const alerts = [];
-		for (const alert of this.props.alerts) {
-			alerts.push(
-				<Alert bsStyle={alert.level} className="text-center">
+		const alerts = this.props.alerts.map(
+			(alert, idx) => (
+				<Alert
+					bsStyle={alert.level}
+					className="text-center"
+					key={idx}
+				>
 					<p>{alert.message}</p>
 				</Alert>
-			);
-		}
+			)
+		);
 
 		return (
 			<div>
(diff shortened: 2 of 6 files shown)


src/client/components/pages/contribute.js
```diff
@@ -58,7 +58,7 @@ function ContributePage() {
 					BookBrainz style guidelines
 				</a> when adding data.
 				Unfortunately the aforementioned are still work in progress
-				and should they not cover some cases, please refer to the
+				and should they not cover some cases, please refer to the&nbsp;
```
leftmostcat: Opportunistic fix to the Contributing page I found while testing.

src/server/helpers/props.js
```diff
@@ -27,5 +27,5 @@ export function escapeProps(props) {
 }
 
 export function generateProps(req, res, props) {
-	return Object.assign({}, req.app.locals, res.locals, props);
+	return Object.assign({}, res.locals, props);
```
leftmostcat: I took this opportunity to pull out `req.app.locals`. There's a ton of information in there that's not relevant to the user and I don't want to risk anything sensitive being added.

Arthelon: have you tested this out on all routes? I believe there were some request variables that were being used.

leftmostcat: I haven't tested it on all routes, but as far as I can tell, the only things which were being pulled from there were site revision and git repo URL. I've moved both of these into `res.locals`.

src/client/containers/layout.js
```diff
@@ -191,6 +191,15 @@ class Layout extends React.Component {
 			</div>
 		);
 
+		const alerts = [];
```
LordSputnik: Can we use:

    const alerts = this.props.alerts.map(...)

I think that's probably a bit cleaner here.

src/server/app.js
```diff
@@ -112,6 +112,14 @@ app.use((req, res, next) => {
 	res.locals.repositoryUrl = repositoryUrl;
 	res.locals.alerts = [];
 
+	if (!req.session) {
+		res.locals.alerts.push({
+			level: 'danger',
+			message: 'We are currently experiencing technical difficulties; ' +
+				'logins will not work until this is resolved.'
```
LordSputnik: Maybe "signing in will not work until this is resolved." makes more sense?

src/client/containers/layout.js
leftmostcat: Good catch.

src/server/app.js
leftmostcat: Will address.

PR #171, opened 2017-12-23
Build maintenance

leftmostcat:
### Problem

`babel-preset-es201?` are deprecated and don't give us very much control over how we transpile.

### Solution

Use the replacement for the yearly presets, `babel-preset-env`.

### Areas of Impact

Build.


.babelrc
```diff
@@ -6,16 +6,52 @@
...
-	  "flow",
-	  "es2016",
-	  "es2017",
-	  "react"
-  ]
+    "flow",
+    "react",
+    ["env", {
+      "targets": {
+        "node": "current"
```
leftmostcat: I've had babel default to transpiling all source to the current version of node. This should decrease the amount of unnecessary work that's happening.

.babelrc
```diff
@@ -6,16 +6,52 @@
...
+      }
+    }]
+  ],
+  "env": {
+    "browser": {
```
leftmostcat: I've created a babel environment specifically for the browser in place of having a separate config for the client directory. As long as `BABEL_ENV` is set to `browser` when building the site code (which I've done in the client JS build scripts), we'll build for our target browsers.

package.json
```diff
@@ -33,50 +33,30 @@
     ]
   },
   "engines": {
-    "node": ">=6.0"
+    "node": ">=8.0"
```
leftmostcat: Update to latest LTS.

package.json
```diff
@@ -104,20 +84,35 @@
     "validator": "^9.1.2"
   },
   "devDependencies": {
+    "babel-cli": "^6.18.0",
```
leftmostcat: I shifted a bunch of build-time dependencies to `devDependencies`. This will allow us to cut down on the size of our docker image. I've done my best to move over everything that's only needed for build and nothing that would be needed at runtime.

package.json
LordSputnik: Also need to update `.travis.yml` to change the minimum node version for testing.

.babelrc
```diff
@@ -6,16 +6,52 @@
...
+          "transform-runtime", {
+            "polyfill": false,
+            "regenerator": true
+          }
+        ]
+      ],
+      "presets": [
+        "flow",
+        "react",
+        ["env", {
```
LordSputnik: Hmm, I think it'd be good to build some margin of error in here. I'd particularly like to make sure the site keeps working for CatCat, so let's see what version of FF he's using and try to keep code compatible with that (even if our official minimum is higher).

leftmostcat: I thought about this, and I believe that IE 11 provides us any margin for error we'd need, though it'd be good to get some testing.

LordSputnik: CatCat confirmed he's on FF 40, so let's use that instead of 44.

PR #173, opened 2017-12-27
[WIP] Duplication refactoring

naiveaiguy:
**WARNING**: In a perhaps futile effort to keep the history clean, I am performing `--amend` and history-rewriting commits every now and then when I spot that I've done something unbearably stupid. 

Work is ongoing. Currently have a `npm` task called `dupreport` that simply runs `jsinspect`. Will fix as many issues raised by it as I can, then fill out the template.

See current `dupreport` here: (*EDIT*: I gave up trying to mantain the Markdown, it's the raw output now)

<a href="https://gist.github.com/naiveaiguy/1eb8a12f2aa75876a1729d6f0bff1c10">gist.github.com/naiveaiguy/1eb8a12f2aa75876a1729d6f0bff1c100</a>

- [ ] sign commits. Am pushing from VM which does not have my GPG keys loaded. Will sign the commits from my non-VM environment later.

PR #174, opened 2017-12-29
Various duplication refactors

naiveaiguy:
_**NOTE**: This PR is #173, but all at once, with signed commits, and as such no coveralls spam. Will push less often next time 🙇_

### Problem

Various instances of duplication. See current `dupreport`:

https://gist.github.com/naiveaiguy/1eb8a12f2aa75876a1729d6f0bff1c10

Originally 23 matches were there. Now there are only 12. 

### Solution

Extracts methods that abstract out the functionality of these duplicates. I've tried as much as possible to keep each commit sensible and working, so any of these can be reverted individually relatively cleanly if the need arises. 

### Areas of Impact

* package.json - `dupreport` script
* entity routes on server - unchanged behaviour - various extracted utils
* entity page on client - unchanged behaviour - `EntityDetail` component - *EDIT*: changed name to `EntityLinks`
* `editor` - unchanged behaviour - small extractions
	I actually want  to do a total refactor of this code because there are some larger problems with 	
    it, but I'll do that after GCI as it's a pretty big undertaking.
* `validators/common` - unchanged behaviour - small extraction

### DIFF
--- a/src/client/components/pages/entities/details.js
+++ b/src/client/components/pages/entities/details.js
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2017  Eshan Singh
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+
+import EntityIdentifiers from './identifiers';
+import EntityRelationships from './relationships';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+const {Col, Row} = bootstrap;
+
+function EntityDetails({entity, identifierTypes, urlPrefix}) {
+	return (
+		<Row>
+			<Col md={8}>
+				<EntityRelationships
+					entityUrl={urlPrefix}
+					relationships={entity.relationships}
+				/>
+			</Col>
+			<Col md={4}>
+				<EntityIdentifiers
+					identifierSet={entity.identifierSet}
+					identifierTypes={identifierTypes}
+				/>
+			</Col>
+		</Row>
+	);
+}
+EntityDetails.displayName = 'EntityDetails';
+EntityDetails.propTypes = {
+	entity: PropTypes.object.isRequired,
+	identifierTypes: PropTypes.array,
+	urlPrefix: PropTypes.string.isRequired
+};
+EntityDetails.defaultProps = {
+	identifierTypes: []
+};
+
+export default EntityDetails;
--- a/src/client/entity-editor/validators/common.js
+++ b/src/client/entity-editor/validators/common.js
@@ -67,7 +67,7 @@ export function validateAlias(value: any): boolean {
 }
 
 export const validateAliases = _.partial(
-	validateMultiple, _, validateAlias);
+	validateMultiple, _.partial.placeholder, validateAlias);
 
 export type IdentifierType = {
 	id: number,
@@ -124,7 +124,8 @@ export function validateIdentifier(
 }
 
 export const validateIdentifiers = _.partial(
-	validateMultiple, _, validateIdentifier, _);
+	validateMultiple, _.partial.placeholder,
+	validateIdentifier, _.partial.placeholder);
 
 export function validateNameSectionName(value: any): boolean {
 	return validateRequiredString(value);
--- a/src/client/entity-editor/validators/common.js
+++ b/src/client/entity-editor/validators/common.js
@@ -25,6 +25,22 @@ import {Iterable} from 'immutable';
 import _ from 'lodash';
 
 
+export function validateMultiple(
+	values: any[],
+	validationFunction: (value: any, ...rest: any[]) => boolean,
+	additionalArgs?: any): boolean {
+	let every = (object, predicate) => _.every(object, predicate);
+	if (Iterable.isIterable(values)) {
+		every = (object, predicate) => object.every(predicate);
+	}
+	else if (!_.isObject(values)) {
+		return false;
+	}
+
+	return every(values, (value) =>
+		validationFunction(value, additionalArgs));
+}
+
 export function validateAliasName(value: any): boolean {
 	return validateRequiredString(value);
 }
@@ -50,17 +66,8 @@ export function validateAlias(value: any): boolean {
 	);
 }
 
-export function validateAliases(values: any): boolean {
-	let every = (object, predicate) => _.every(object, predicate);
-	if (Iterable.isIterable(values)) {
-		every = (object, predicate) => object.every(predicate);
-	}
-	else if (!_.isObject(values)) {
-		return false;
-	}
-
-	return every(values, (value) => validateAlias(value));
-}
+export const validateAliases = _.partial(
+	validateMultiple, _, validateAlias);
 
 export type IdentifierType = {
 	id: number,
@@ -116,19 +123,8 @@ export function validateIdentifier(
 	);
 }
 
-export function validateIdentifiers(
-	values: any, types?: ?Array<IdentifierType>
-): boolean {
-	let every = (object, predicate) => _.every(object, predicate);
-	if (Iterable.isIterable(values)) {
-		every = (object, predicate) => object.every(predicate);
-	}
-	else if (!_.isObject(values)) {
-		return false;
-	}
-
-	return every(values, (value) => validateIdentifier(value, types));
-}
+export const validateIdentifiers = _.partial(
+	validateMultiple, _, validateIdentifier, _);
 
 export function validateNameSectionName(value: any): boolean {
 	return validateRequiredString(value);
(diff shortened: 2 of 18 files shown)


src/server/routes/entity/creator.js
```diff
@@ -232,18 +229,14 @@ router.get(
 			creator
 		);
 
-		const props = generateProps(req, res, {
-			creator,
```
LordSputnik: Looks like `creator` may be missing from the new set of props.

src/server/routes/entity/creator.js
```diff
@@ -95,17 +96,13 @@ router.get(
...
-			languageOptions: res.locals.languages,
-			requiresJS: true,
-			subheading: 'Add a new Creator to BookBrainz',
-			submissionUrl: '/creator/create/handler'
-		});
+		const props = generateEntityProps(
+			'creator', 'create', req, res, {
+				genderOptions: res.locals.genders,
+				identifierTypes: filteredIdentifierTypes,
+				submissionUrl: '/creator/create/handler'
```
LordSputnik: Maybe we can set a default for this too - I'd suggest passing in a BBID or null value to the generateEntityProps function, and then using that to determine whether it's a create or edit action. Then you can also use the BBID for the submission URL generation (if not null).

src/server/routes/entity/creator.js
```diff
@@ -95,17 +96,13 @@ router.get(
...
-			initialState: {},
-			languageOptions: res.locals.languages,
-			requiresJS: true,
-			subheading: 'Add a new Creator to BookBrainz',
-			submissionUrl: '/creator/create/handler'
-		});
+		const props = generateEntityProps(
+			'creator', 'create', req, res, {
+				genderOptions: res.locals.genders,
+				identifierTypes: filteredIdentifierTypes,
```
LordSputnik: Could we do identifier type filtering in the generateEntityProps function too? Seems like we have all the info there we need.

src/server/helpers/entityRouteUtils.js
```diff
@@ -53,3 +64,33 @@ export function generateEntityProps(
...
+				<EntityEditor
+					validate={getValidator(props.entityType)}
+					{...propHelpers.extractChildProps(rest)}
+				>
+					<EntitySection/>
+				</EntityEditor>
+			</Provider>
+		</Layout>
+	);
+	props.initialState = store.getState();
```
LordSputnik: It might be a good idea to document that props gets modified within this function. Might be better to return updated props as well as markup to make this function cleaner.

src/server/routes/entity/creator.js
```diff
@@ -238,32 +210,9 @@ router.get(
 			}
 		);
 
-		const {initialState, ...rest} = props;
-
 		const rootReducer = createRootReducer(props.entityType);
```
LordSputnik: Can probably move `createRootReducer` into `entityEditorMarkup`.

src/server/helpers/entityRouteUtils.js
```diff
@@ -94,3 +96,58 @@ export function entityEditorMarkup(
...
+		req: express.request,
+		res: express.response) {
+		const validate = getValidator(entityType);
+		if (!validate(req.body)) {
+			const err = new error.FormSubmissionError();
+			error.sendErrorAsJSON(res, err);
+		}
+
+		req.body = transformNewForm(req.body);
+		return entityRoutes.createEntity(
```
LordSputnik: Looks like this is only creating, and not allowing for editing.

src/client/components/pages/entities/details.js
```diff
@@ -0,0 +1,57 @@
...
+
+import EntityIdentifiers from './identifiers';
+import EntityRelationships from './relationships';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+const {Col, Row} = bootstrap;
+
+function EntityDetails({entity, identifierTypes, urlPrefix}) {
```
LordSputnik: Could we rename this to `EntityLinks`?

src/server/routes/editor.js
```diff
@@ -240,29 +242,7 @@ router.get('/:id/revisions', (req, res, next) => {
...
-						if (unlock !== null) {
-							editorJSON.title =
-								unlock.relations.title.attributes;
-						}
-						return editorJSON;
-					});
-			}
-			return editorTitleJSON;
-		})
+		.then(getEditorTitleJSON, TitleUnlock)
```
LordSputnik: Editor is being passed in here, not JSON - need to convert, I think.

src/server/routes/editor.js
```diff
@@ -155,28 +178,7 @@ router.get('/:id', (req, res, next) => {
...
```
LordSputnik: What is the second parameter here for? The promise spec suggests `then` takes two functions.

src/client/entity-editor/validators/common.js
```diff
@@ -50,17 +66,8 @@ export function validateAlias(value: any): boolean {
...
-		every = (object, predicate) => object.every(predicate);
-	}
-	else if (!_.isObject(values)) {
-		return false;
-	}
-
-	return every(values, (value) => validateAlias(value));
-}
+export const validateAliases = _.partial(
+	validateMultiple, _, validateAlias);
```
LordSputnik: Might be best practice to use `_.partial.placeholder` in case this code gets copied for anything on the client side.

src/client/components/pages/entities/details.js
naiveaiguy: Sure.

src/client/entity-editor/validators/common.js
naiveaiguy: Yeah, this is a good idea anyway for the clarity.

src/server/helpers/entityRouteUtils.js
naiveaiguy: I'll document it, but I don't agree that returning `props` would make it cleaner. The value is unrelated to markup, and it'll make calling code messier without any counter-benefit.

src/server/helpers/entityRouteUtils.js
naiveaiguy: Oops! Will fix.

src/server/routes/editor.js
naiveaiguy: This was stupid of me - I wanted to pass it to `getEditorTitleJSON`.

src/server/routes/editor.js
naiveaiguy: Same as above.

src/server/routes/entity/creator.js
naiveaiguy: Yeah, you're right.

src/server/routes/entity/creator.js
naiveaiguy: I can use `res.locals.entity` to determine if this is a create or edit action and use it accordingly.

src/server/routes/entity/creator.js
naiveaiguy: Yeah, the above solves this problem as well.

src/server/routes/entity/creator.js
naiveaiguy: Sure.

src/server/helpers/entityRouteUtils.js
naiveaiguy: On second thought, this would allow me to declare `props` as the result of `entityEditorMarkup`. Will do this.

src/server/helpers/entityRouteUtils.js
```diff
@@ -0,0 +1,187 @@
...
+	(entity) => new Object()): Object {
+	const entityName = _.capitalize(entityType);
+	const {entity} = res.locals;
+
+	const filteredIdentifierTypes = utils.filterIdentifierTypesByEntity(
+		res.locals.identifierTypes,
+		entity ? entity : entityName
+	);
+
+	const submissionUrl = entity ?
```
LordSputnik: could you do 

    const isEdit = Boolean(entity);

And then simplify the heading, subheading and submissionURL generation...

    const heading = isEdit ? `Edit ${entityName}` : `Create ${entityName}`;
    const subheading = isEdit ? `Edit an existing ${entityName} on BookBrainz` : `Add a new ${entityName} to BookBrainz`;
    const submissionUrl = isEdit ? `/${entityType}/${entity.bbid}/edit/handler` : `/${entityType}/create/handler`;

Then there's no need to pass in the action here.

src/server/routes/editor.js
```diff
@@ -240,29 +248,7 @@ router.get('/:id/revisions', (req, res, next) => {
 				}
 			}
 		})
-		.then((editor) => {
```
LordSputnik: Need to convert editor to JSON here.

src/server/helpers/entityRouteUtils.js
naiveaiguy: Sure.

src/server/routes/editor.js
naiveaiguy: Ah, you're right. Fixed.

src/server/helpers/entityRouteUtils.js
```diff
@@ -0,0 +1,189 @@
...
+	entityType: string,
+	req: express.request, res: express.response,
+	additionalProps: Object,
+	initialStateCallback: (entity: ?Object) => Object =
+	(entity) => new Object()): Object {
+	const entityName = _.capitalize(entityType);
+	const {entity} = res.locals;
+	const isEdit = Boolean(entity);
+
+	const filteredIdentifierTypes = utils.filterIdentifierTypesByEntity(
```
LordSputnik: filterIdentifierTypesByEntity only takes an Object as its second argument. Previously, both `utils.filterIdentifierTypesByEntity` and `utils.filterIdentifierTypesByEntityType` were used here, so I think we need to make the function dependent on `isEdit`.

naiveaiguy: Man, I should pay more attention to these things 😅 Will do so in the future.

naiveaiguy: Fixed.

PR #175, opened 2017-12-30
Improve search indexing

leftmostcat:
### Problem

Bulk indexing was failing to index many entities.

### Solution

Bump the bulk operation size to try to reduce the chances of silent indexing failure. Only index the master revision of entities for which the master revision includes entity data.

### Areas of Impact

Search indexing.

### DIFF
--- a/config/config.json.example
+++ b/config/config.json.example
@@ -27,9 +27,9 @@
 		}
 	},
 	"search": {
-	    "host": "localhost:9200",
-	    "apiVersion": "5.5",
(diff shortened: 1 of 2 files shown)


leftmostcat:
Surprisingly, if there's an error during bulk indexing, no error is raised. See https://github.com/elastic/elasticsearch-js/issues/620.

LordSputnik:
Can we do error detection in the "then" handler based on the response from elasticsearch?

PR #177, opened 2018-01-07
Add documentation and tests for renderRelationship

naiveaiguy:
### Problem
There's no documentation or tests for `renderRelationship`. 

### Solution
Add documentation and tests. Maybe not the biggest issue, but I though it was useful to have some anyway. It was also good to get used to `mocha` and `chai`. Will make some more substantial tests after GCI.

### Areas of Impact
No behaviour changes, only new tests and docs.

### DIFF
--- a/src/server/helpers/render.js
+++ b/src/server/helpers/render.js
@@ -52,6 +52,10 @@ type Relationship = {
 
 /**
  * Returns the markup corresponding to the given entity relationship.
+ * This markup is defined by a Handlebars template in
+ * relationship.type.displayTemplate.
+ * This function is used, for example, to render
+ * relationships on entity display pages.
  * @param {Relationship} relationship - Relationship object.
  * @returns {string} - Rendered HTML string.
  */
--- a/src/server/helpers/render.js
+++ b/src/server/helpers/render.js
@@ -23,7 +23,7 @@ import Handlebars from 'handlebars';
 
 
 type EntityInRelationship = {
-	bbid: number | string,
+	bbid: string,
 	defaultAlias?: {name: string},
 	type: string
 };
@@ -36,7 +36,7 @@ type Relationship = {
 
 /**
  * @typedef {Object} EntityInRelationship
- * @property {number|string} bbid
+ * @property {string} bbid
  * @property {?Object} defaultAlias
  * @property {string} defaultAlias.name
  * @property {string} type
--- a/src/server/helpers/render.js
+++ b/src/server/helpers/render.js
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2015  Ben Ockmore
  *               2015  Sean Burke
+ *               2018  Eshan Singh
  *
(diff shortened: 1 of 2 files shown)


test/test-render-relationship.js
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (C) 2016  Eshan Singh
```
Arthelon: *2018

naiveaiguy: Oops, sorry! Fixed.

src/server/helpers/render.js
```diff
@@ -16,12 +16,46 @@
...
  */
+// @flow
 
 import * as utils from './utils';
 import Handlebars from 'handlebars';
 
 
-function renderRelationship(relationship) {
+type EntityInRelationship = {
+	bbid: number | string,
```
Arthelon: I noticed that you're using integers for the relationship tests but all BBIDs should be strings.

naiveaiguy: Fixed, you're right.

src/server/helpers/render.js
```diff
@@ -16,12 +16,46 @@
...
+/**
+ * @typedef {Object} Relationship
+ * @property {EntityInRelationship} source
+ * @property {EntityInRelationship} target
+ * @property {Object} type
+ * @property {string} type.displayTemplate
+ */
+
+/**
+ * Returns the markup corresponding to the given entity relationship.
```
LordSputnik: Would be good to mention what this is used for (rendering relationships on entity display pages), and also what defines the markup (the display template).

test/test-render-relationship.js
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+import {expect} from 'chai';
+import renderRelationship from '../src/server/helpers/render.js';
+
+
+const relationshipTests = {
```
LordSputnik: It'd be good to see some more "bad" data tested here. What if one of the provided entities is null? What if a number is provided instead of a string? What if the display template is null, or a blank string? I don't know what happens in these circumstances right now, so we should make sure nothing bad happens by testing any likely failures.

naiveaiguy: All of these tests (except for empty template) would be covered by flow type annotations. I'll add them anyway for the moment. What currently happens:

1) one of the entities is null - `TypeError: cannot read property 'defautAlias' of null`
2) number is provided instead of a string - it works in the same exact way
3) `displayTemplate` is null - Handlebars complains `Error: You must pass a string or Handlebars AST to Handlebars.compile. You passed null`
4) `displayTemplate is the empty string - it returns the empty string as well

LordSputnik:
Just need to add yourself to the copyright header of the helpers.js file now, then I'll merge.

PR #180, opened 2018-01-16
Fix(Area field of editor not getting edited):Area field of editor get…

agbilotia1998:
…ting edited.

The areaId now is been sent as data to server instead of null.
Fixes #BB-234

### Solution
The areaId is in area constant itself instead of area.id,therefore corrected it.
area


agbilotia1998:
@LordSputnik please review.

PR #181, opened 2018-01-17
BB-241: fix(deletion): Add missing spaces in remove entity screen.

shivam-tripathi:
### Problem
BB-241: Space error in remove entity screen
https://tickets.metabrainz.org/projects/BB/issues/BB-241?filter=allopenissues

### Solution
Insert spaces at relevant positions.

### Areas of Impact
src/client/components/forms/deletion.js
Remove entity screen.

### Final result
commit

### DIFF
--- a/src/client/components/forms/deletion.js
+++ b/src/client/components/forms/deletion.js
@@ -111,9 +111,9 @@ class EntityDeletionForm extends React.Component {
 								footer={footerComponent}
 								header={headerComponent}
 							>
-								If you’re sure that {entity.type} {entityName}
-								should be deleted, please enter a revision note
-								below and confirm deletion.
+								If you’re sure that {entity.type} {' '}
+								{entityName} should be deleted, please enter a
+								revision note below and confirm deletion.
 
 								<CustomInput
 									ref={(ref) => this.note = ref}


LordSputnik:
Hi, not sure what this is trying to fix. Looks like there are already spaces in the relevant places?

shivam-tripathi:
Hi, I have update my commit description - the screen shot might have caused some confusion. Full details of the bug can be found at [https://tickets.metabrainz.org/projects/BB/issues/BB-241?filter=allopenissues](https://tickets.metabrainz.org/projects/BB/issues/BB-241?filter=allopenissues).

For more reference, I am attaching a screenshot from the bb site: 
err

src/client/components/forms/deletion.js
```diff
@@ -111,9 +111,9 @@ class EntityDeletionForm extends React.Component {
 								footer={footerComponent}
 								header={headerComponent}
 							>
-								If you’re sure that {entity.type} {entityName}
```
LordSputnik: So this gets transpiled to:

    'If you\u2019re sure that ',
    entity.type,
    ' ',
    entityName,
    'should be deleted, please enter a revision note below and confirm deletion.',

Given that, I think whitespace between `{} {}` is kept, but when a `{}` is at the end of the line, the leading whitespace on the line following that is ignored. So we just need to add a `{' '}` at any point where a variable substitution is currently at the end of the line (as is the case on line 114 here).

shivam-tripathi: This is very strange, I didn't look the transpiled code. The output on running with changes doesn't introduce double spaces - there is one space alright. 
Regarding insertion, this can be a good place: https://reactjs.org/blog/2014/02/20/react-v0.9.html#jsx-whitespace.

LordSputnik: Hmm I don't understand why there are no double spaces here. It might be best to make it `that {entity.type}{' '}` to avoid unwanted changes in the future.

MonkeyDo: @shivam-tripathi I have an alternate proposition, adding a non-breaking space before should:
`If you’re sure that {entity.type} {entityName}
&nbsp;should be deleted, please enter a revision note
below and confirm deletion.`

MonkeyDo:
Problem solved in (and closing in favor of) PR #243

PR #182, opened 2018-01-26
BB-237 fix(validateDates): Accept dates in YYYY-MM-DD, YYYY-MM and YYYY formats. Also ensure the begin date is less than the end date.

shivam-tripathi:
### Problem

Dates presently are being accepted in  ISO 8601 format which allows for dates being entered in formats different than ones mentioned above.

### Solution
The PR utilises moment, already a dependency in the project to verify date to the exact format, and does away with the validator.isDate method.

### Areas of Impact
src/client/entity-editor/validators/base.js


LordSputnik:
We also need to accept YYYY and YYYY-MM.

shivam-tripathi:
@LordSputnik, How does it look now? :)

src/client/entity-editor/validators/base.js
```diff
@@ -20,9 +20,12 @@
 
 import {Iterable} from 'immutable';
 import _ from 'lodash';
+import moment from 'moment';
 import validator from 'validator';
 
 
+const dateFormats = ['YYYY-MM-DD', 'YYYY-MM', 'YYYY'];
```
LordSputnik: A constant called VALID_DATE_FORMATS might be better here (if the linter allows that name).

shivam-tripathi:
Normally, this PR should have failed tests on validate*SectionendDate, but that didn't happen as the test script in the package.json wasn't running all the tests (more precisely, those in the subdirectories). The PR #193 solves this, and I will be adding fixes to the broken tests.

shivam-tripathi:
@LordSputnik Good to go from my side. The test coverage would increase once #193 is merged.

MonkeyDo:
Thanks @shivam-tripathi ! :)
This took a while, thanks for your patience

PR #185, opened 2018-02-26
Remove unnecessary call of _.snakeCase on a literal

akhilesh26:
### Problem

There is use of _.snakeCase on literal(editorId).

### Solution

Remove use of _.snakeCase and use editor_id directly.

### Areas of Impact

src/server/helpers/achievement.js
 src/server/routes/editor.js


Arthelon:
Nice catch!

akhilesh26:
@LordSputnik thanks for the review. I have updated the commit message as suggested. :)

PR #188, opened 2018-03-12
Add feature to warn users if an entity name already exists 

shivam-tripathi:
### Problem
Presently, there is no warning if a user enters an entity with a name already existing in the database. While it may be the case that two entities with same name exists, it may also lead to user mistakenly adding an entity which may already be present in the database.

### Solution
It adds raises a warning if the user enters an entity with a name which already exists in the database.

### Areas of Impact
src/client/entity-editor/name-section
src/client/entity-editor/commons
src/server/routes 
src/server/helpers


MonkeyDo:
Closing in favour of #208 (and #195)

PR #189, opened 2018-03-12
fix(client): remove logger from client side.

akhilesh26:
Remove log.debug from revision page to resolve
error: can not read property `readable` of undefined.

Fixes #245

### Problem

Issue [BB-245](https://tickets.metabrainz.org/projects/BB/issues/BB-245?filter=allopenissues) 

### Solution

Remove log.debug from revision page in the client side.

### Areas of Impact

src/client/components/pages/revision.js


LordSputnik:
Could you add a "TODO" comment saying something like "add proper error handling"? That way we know to come back to it later and implement something to give a visible error message to the user.

akhilesh26:
I have added the `TODO` statement.

PR #190, opened 2018-03-19
feat(handleDelete): Upon deletion, delete the entity from ElasticSearch also

shivam-tripathi:
### Problem
Presently upon deleting an entity, it still turn up in search as 'unnamed'

### Solution
This is happening as the entity has not been deleted in the elasticSearch index. In this PR I add code to to delete the entity from the ES index also.

### Areas of Impact
src/server/helpers/search
src/server/routes/entity


shivam-tripathi:
It is to be noted here that this fixes the case when user searches the entity in the ES (the recent activity still shows unnamed entries). I'll open a separate PR for that.
@LordSputnik: This means that as ES is already indexed everytime BB fires up, all the previously deleted entities still exist in the ES index. A solution to this is to delete the present index, and reindex the ES (done automatically) next time BB site fires up.

MonkeyDo:
I've tested this locally, and everything was working.
Travis test for Node LTS is failing at the moment, even on commits that were passing fine. Passing on Node 8.
Not sure what that issue is yet, but I know it's not code related, so will ignore it and merge anyway.

PR #192, opened 2018-03-28
fix(deleteEntity): Update the revision parent upon deletion

shivam-tripathi:
Upon deleting an entity, update the revision_parent table.

Fix #BB-251


MonkeyDo:
Updated an good to merge, thanks @shivam-tripathi !

PR #194, opened 2018-04-05
feat(delete): Upon deletion of an entity, redirect to home page

shivam-tripathi:
### Problem
Upon deleting an entity, the user lands up on the new revised entity page which is unnamed and contains no relevant information. Infact, the user still has access to the delete entity button - which means s/he can delete a previously deleted entity. 

### Solution
After deleting an entity, the user is redirected to the home page, and thus has no access to the previously deleted entity (once #190 is merged, the user cannot access it via search too). In future we can perhaps develop a way by which we can provide access to the previously deleted entities displaying the parent_revision data to undo the deletion.

### Areas of Impact
Deletion form.

### DIFF
--- a/src/client/components/forms/deletion.js
+++ b/src/client/components/forms/deletion.js
@@ -49,7 +49,7 @@ class EntityDeletionForm extends React.Component {
 		request.post(this.deleteUrl)
 			.send()
 			.then(() => {
-				window.location.href = this.entityUrl;
+				window.location.href = '/';
 			})
 			.catch((res) => {
 				const {error} = res.body;


LordSputnik:
I'm not sure about this. The idea of going back to the entity page is to provide confirmation that the entity has been deleted (the user should see a "This entity has been deleted" box). We should just disable the delete and edit buttons, I think.

shivam-tripathi:
That makes sense. Alongside disabling the buttons, we can probably instead of displaying everything as null, show alias name as "This ${entity} has been deleted". Later we can add a button to undo this action, and can extend the above message as "To undo this, please click here/this button" or something like that.

MonkeyDo:
I'll close this, as there are good suggestions for how to proceed instead.

PR #196, opened 2018-05-20
/delete-user/:metabrainzUserID endpoint for remote editor deletion

LordSputnik:
### Problem
Users can request deletion on MB, but currently an account deleted on MB remains active on BB.

### Solution
This PR introduces an endpoint, `/delete-user/:metabrainzUserID`, which can be pinged by MusicBrainz whenever a user is deleted there. See https://tickets.metabrainz.org/browse/MBS-9680

### Areas of Impact
Authentication. Also, need to add a regular expression check to prevent users setting their name to "Deleted User #\d+", otherwise there might be conflicts.

### DIFF
--- a/src/server/routes/auth.js
+++ b/src/server/routes/auth.js
@@ -16,8 +16,10 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+import _ from 'lodash';
 import express from 'express';
 import passport from 'passport';
+import request from 'superagent';
 import status from 'http-status';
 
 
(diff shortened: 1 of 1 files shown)


mwiencek:
Trying to test this locally and I'm running into the following issue:

```
(node:91464) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'user' of undefined
    at deleteUserAndGetResCode (/Users/michael/code/bookbrainz-site/src/server/routes/auth.js:102:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:91464) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
```

src/server/routes/auth.js
```diff
@@ -55,4 +57,68 @@ router.get('/logout', (req, res) => {
 	res.redirect(status.SEE_OTHER, '/');
 });
 
+async function deleteUserAndGetResCode(orm, metabrainzUserID, accessToken) {
+	if (!_.isString(accessToken)) {
+		return status.BAD_REQUEST;
+	}
+
+	const USER_INFO_URL = 'https://musicbrainz.org/oauth2/userinfo';
```
mwiencek: It'd be nice if the MusicBrainz server URL were configurable...I had to hand-edit this file and a couple in the `passport-musicbrainz-oauth2` module so I could test things against my local MB instance.

PR #202, opened 2018-08-21
Webpack

MonkeyDo:
### Problem
Ticket [BB-256](https://tickets.metabrainz.org/browse/BB-256): Use webpack for bundling

### Solution
Running the project in development (```npm run debug``` or ```npm run debug-watch-server```) now uses webpack to load, transpile and bundle the client code, with hot module replacement active.
That means changes to the javascript or less client files will automatically change the browser page without need for reloading the page, thus conserving the application state.

### Areas of Impact
Currently, this only impacts when running in development.
Production builds I have left as before until I figure out a few remaining kinks.


LordSputnik:
Merged in ade20c5dd03dc9915dcee79632679c5be0af159b

LordSputnik:
Rebased to add a prefix to one of the last 4 commit messages - so closing this manually.

PR #204, opened 2018-08-26
New relationship editor

LordSputnik:
### Problem
There is no way to edit relationships in the site.

### Solution
This PR consolidates the entity creation and editing code. It adds a client side relationship editor as part of the existing entity editor React app, and then adds functions to handle the new relationship data server-side, as part of the new entity creation and editing function. The relationship schema has also changed, so some updates are made to relationship rendering to deal with this.

### Areas of Impact
The client-side entity editor, the server side entity revision handling code, and relationship rendering code.

### DIFF
--- a/src/client/entity-editor/relationship-editor/relationship-editor.js
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.js
@@ -105,7 +105,7 @@ type RelationshipModalProps = {
 	initRelationship: ?_Relationship,
 	onCancel?: () => mixed,
 	onClose?: () => mixed,
-	onSave?: (_Relationship) => mixed
+	onAdd?: (_Relationship) => mixed
 };
 
 type RelationshipModalState = {
@@ -153,9 +153,9 @@ function getInitState(
 class RelationshipModal
 	extends React.Component<RelationshipModalProps, RelationshipModalState> {
 	static defaultProps = {
+		onAdd: null,
 		onCancel: null,
-		onClose: null,
-		onSave: null
+		onClose: null
 	};
 
 	constructor(props: RelationshipModalProps) {
@@ -179,11 +179,11 @@ class RelationshipModal
 		});
 	};
 
-	handleSave = () => {
-		const {onSave} = this.props;
-		if (onSave) {
+	handleAdd = () => {
+		const {onAdd} = this.props;
+		if (onAdd) {
 			if (this.state.relationship) {
-				onSave(this.state.relationship);
+				onAdd(this.state.relationship);
 			}
 		}
 	};
@@ -301,9 +301,9 @@ class RelationshipModal
 					<Button
 						bsStyle="success"
 						disabled={submitDisabled}
-						onClick={this.handleSave}
+						onClick={this.handleAdd}
 					>
-						Save
+						Add
 					</Button>
 				</Modal.Footer>
 			</Modal>
--- a/src/client/entity-editor/relationship-editor/relationship-editor.js
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.js
@@ -36,14 +36,6 @@ import _ from 'lodash';
  * action, which allows the rest of the app to
  */
 
-const ENTITY_TYPE_VALUES = [
-	{id: 'creator', label: 'Creator'},
-	{id: 'edition', label: 'Edition'},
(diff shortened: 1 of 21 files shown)


src/client/entity-editor/relationship-editor/relationship-editor.js
```diff
@@ -0,0 +1,339 @@
...
+} from './types';
+import EntitySearchField from '../common/entity-search-field';
+import React from 'react';
+import ReactSelect from 'react-select';
+import Relationship from './relationship';
+import _ from 'lodash';
+
+/*
+ * This is a simple React component. No Redux connection. Has an onSubmit
+ * action, which allows the rest of the app to
```
MonkeyDo: Theres a bit missing there :)

src/client/entity-editor/relationship-editor/relationship-editor.js
```diff
@@ -0,0 +1,339 @@
...
+						to a work as an author, or a work to another work
+						to show translation or derivation.
+					</p>
+					<hr/>
+					<Row>
+						<Col md={10} mdOffset={1}>
+							<form>
+								<Row>
+									<Col md={4}>
+										{this.renderEntityTypeSelect()}
```
MonkeyDo: I suggest reorganizing the editor modal in this way

1. Remove EntityTypeSelect
2. Amend EntitySelect so that a user can search for any entity, limited to those that the current entity can have a valid relationship with
  - For example, a Work doesn't seem to have any allowable relationship with a Publication, so Publications should be filtered out. I also imagine we don't want Areas appearing there either.
  - The search results can be arranged by entity type, under an entity type header in the select
3. Keep RelationshipSelect as is, showing the available relationships

src/client/entity-editor/relationship-editor/relationship-editor.js
```diff
@@ -0,0 +1,339 @@
...
+							/>
+						</Col>
+					</Row>
+					<Button bsStyle="danger" onClick={onCancel}>Cancel</Button>
+					<Button
+						bsStyle="success"
+						disabled={submitDisabled}
+						onClick={this.handleSave}
+					>
+						Save
```
MonkeyDo: A bit of a detail, but I'd suggest the button say "Add" or "Confirm" or something along those lines rather than "Save", as clicking it does not in fact save the relationship directly and you have to submit the changes.

src/client/entity-editor/relationship-editor/relationship-editor.js
```diff
@@ -0,0 +1,339 @@
...
+					value={this.state.entityType}
+					onChange={this.handleEntityTypeChange}
+				/>
+			</FormGroup>
+		);
+	}
+
+	renderEntitySelect() {
+		return (
+			<EntitySearchField
```
MonkeyDo: The EntitySearchField is limited to 10 results. Is that on purpose?

src/client/entity-editor/relationship-editor/relationship.js
```diff
@@ -0,0 +1,78 @@
...
+	);
+	const targetObject = getEntityObjectForDisplay(
+		reversed ? sourceEntity : targetEntity, link
+	);
+
+	const usedLinkPhrase = reversed ? reverseLinkPhrase : linkPhrase;
+
+	return (
+		<div>
+			<Entity {...sourceObject}/>
```
MonkeyDo: Maybe add the EntityType icon in front of each entity?
It would make sentences like "Harry Potter and the Philosopher’s Stone contains Harry Potter and the Philosopher’s Stone" a bit more understandable.

LordSputnik: In this particular bit of code, it already does do that - the `Entity` component renders the icon. I think the issue is with the entity display code (as used on the entity pages - see https://test.bookbrainz.org/creator/8f3d202f-fa37-4b71-9e81-652db0f8b83d)

src/client/entity-editor/relationship-editor/relationship-editor.js
LordSputnik: It's the default limit for Elasticsearch - see https://www.elastic.co/guide/en/elasticsearch/reference/5.5/search-request-from-size.html

We can make this bigger, but maybe better after this PR.

PR #205, opened 2018-09-03
Update vulnerable packages

Arthelon:
The new npm audit tool has revealed a number of vulnerabilities in the project's depdendencies. This PR aims to perform version updates on vulnerable dependencies while minimizing the changes to existing code and functionality.


LordSputnik:
I'll merge this, but FYI, `serve-static` is just the same thing as `express.static`. The idea of using `express-static-cache` was to keep the JS files in memory. I think the OS will do this automatically, based on what I've read. So I'll do a follow-up commit to just drop the explicit dependency and use a single express.static call.

Arthelon:
okay good to know. There are still some vulnerable dependencies that need to be updated manually. I might make another PR later on during the week.

PR #209, opened 2018-10-23
Add and cancel relationships with keys

PROTechThor:
Code to add relationship by pressing the Enter Key. Also close with Escape key.

### Problem
On creating or editing an entity in BookBrainz , if a user clicks "Add relationship" to open the relationship editor and fills in both fields, they should be able to press the enter key to add the relationship (equivalent to clicking the "Add" button) for convenience. For comparison, pressing the escape key closes the modal (equivalent to clicking the "Close" button).

### Solution
Pressing the Enter Key adds relationship. Pressing the escape key closes them modal. 

### Areas of Impact
Creating or editing an entity in BookBrainz. (https://bookbrainz.org/edition/create) The modal pulled up after clicking "Add Relationship" is the area impact.

### DIFF
--- a/src/client/entity-editor/relationship-editor/relationship-editor.js
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.js
@@ -206,17 +206,6 @@ class RelationshipModal
 			}
 		}
 	};
-	handleKeyPress(event) {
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Implemented in PR #214

PR #212, opened 2018-10-26
fix(IdentifierEditor): Remove blank identifiers

Ge0rg3:
### Problem

Ticket [BB-276](https://tickets.metabrainz.org/browse/BB-276):

> On creating or editing an entity, if I click "Add identifier" three times in the identifier editor popup, but only fill one, I still get "Edit 3 identifiers" on the link after closing the popup.
> It should realize there's only 1, and probably get rid of the empty identifier rows when I close the popup in the first place.

### Solution

Creates a new action ("REMOVE_EMPTY_IDENTIFIERS"), which removes all Identifiers without a value/type, and removes them when the Identifiers editor window is closed.

### Areas of Impact

Only [src/client/entity-editor/identifier-editor](https://github.com/Ge0rg3/bookbrainz-site/tree/master/src/client/entity-editor/identifier-editor), which relates to [this](http://46.101.26.141:9099/edition/create) area of Bookbrainz.

### DIFF
--- a/src/client/entity-editor/identifier-editor/reducer.js
+++ b/src/client/entity-editor/identifier-editor/reducer.js
@@ -18,7 +18,7 @@
 
 import {
 	ADD_IDENTIFIER_ROW, REMOVE_IDENTIFIER_ROW, UPDATE_IDENTIFIER_TYPE,
-	UPDATE_IDENTIFIER_VALUE
+	UPDATE_IDENTIFIER_VALUE, REMOVE_EMPTY_IDENTIFIERS
 } from './actions';
 import Immutable from 'immutable';
 
@@ -53,7 +53,10 @@ function reducer(
 			return state.setIn([payload.rowId, 'type'], payload.value);
 		case REMOVE_IDENTIFIER_ROW:
 			return state.delete(payload);
-
+		case REMOVE_EMPTY_IDENTIFIERS:
+			return state.filterNot(identifier => (
+				identifier.get('value') === "" || identifier.get('type') === null)
+			);
 		// no default
 	}
 	return state;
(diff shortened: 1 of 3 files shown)


Freso:
Just a small thing: You should preface the title with "BB-276" to our Jira automatically picks it up and links it from the ticket. :) I don't know enough about BB to comment on anything else.

Ge0rg3:
Okay, fixed 👌

src/client/entity-editor/identifier-editor/reducer.js
```diff
@@ -53,7 +53,10 @@ function reducer(
 			return state.setIn([payload.rowId, 'type'], payload.value);
 		case REMOVE_IDENTIFIER_ROW:
 			return state.delete(payload);
-
+		case REMOVE_EMPTY_IDENTIFIERS:
+			return state.filterNot(identifier =>
+				identifier.get('value') === '' || identifier.get('type') === null
```
MonkeyDo: Hi @Ge0rg3 !

I was able to test locally, and I see everything is working fine. The code is also clean and readable, everything in the right place. Well done !

I have one suggestion, after some consideration: removing identifiers only when they have both value **and** type empty.
The reason for that is that we currently have a validation system that will show users if they only entered one. For example, a user might paste in a value but forget to select a type (or think it was auto-detected). In that case, the validator will show an error, and the user can correct the identifier rather than have to start from scratch, wondering where the previous one went.

Whereas identifiers that are totally empty, we know we can get rid of as we're not destroying any information they could store.

I hope I'm making sense.
Thanks for sticking to it, I'm looking forward to merging the PR, it's a good improvement!

PR #214, opened 2018-10-30
BB-272 - feat(RelationshipEditor): Submit on enter

Ge0rg3:
### Problem

As desribed on [BB-272](https://tickets.metabrainz.org/browse/BB-272): 

> Once I've selected both an entity and a relationship type, letting me submit the form by just pressing Enter while on either of the fields would be a big time saver. After all, it's trivial to just edit / undo if I messed it up.

### Solution

This creates a handle for keypresses, that calls the handleAdd handle whenever the "enter" key is pressed.
**Change from ticket criteria:**
While the ticket says that the enter key should submit the data anywhere in the form, I've made it only submit when on the second input box. This is because when selecting a user in the first input box, the enter key can be used to make your selection, which would thus close the Editor before the user has a chance to enter anything in the second input box.  
This can be easily, as we would just have to move the onKeyUp tag to the start of the Modal dialogue.  
**Why use onKeyUp:**
While most online resources use onKeyPress, onKeyUp is the most effective here because since the user is already in an input box, the keypresses do not get passed to that event listener. A common React workaround to this is to use onInputKeyDown, but this makes the code far less readable imo.

### Areas of Impact

Only [src/client/entity-editor/relationship-editor/relationship-editor.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/identifier-editor/value-field.js) (the Relationship editor dialogue).

### DIFF
--- a/src/client/entity-editor/identifier-editor/value-field.js
+++ b/src/client/entity-editor/identifier-editor/value-field.js
@@ -45,7 +45,7 @@ function ValueField({
 	);
 
 	return (
-		<CustomInput label={label} type="text" {...rest}/>
+		<CustomInput autoFocus label={label} type="text" {...rest}/>
 	);
 }
 ValueField.displayName = 'ValueField';
--- a/src/client/entity-editor/identifier-editor/value-field.js
+++ b/src/client/entity-editor/identifier-editor/value-field.js
@@ -45,7 +45,7 @@ function ValueField({
 	);
 
 	return (
-		<CustomInput autoFocus label={label} type="text" {...rest}/>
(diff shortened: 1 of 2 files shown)


Ge0rg3:
Travis seems to be failing at one of the dependencies.

Ge0rg3:
@MonkeyDo I've made the changes you requested. The nested if statement is explained in the commit message (in case it seems like a strange choice).

MonkeyDo:
Great, thanks @Ge0rg3 !
I'll merge that now.

PR #215, opened 2018-10-31
Fix and improve the search page

MonkeyDo:
### Problem
Currently, the search page does not work properly. If you search from, say, the home page, you'll get redirected to the search page, with results showing. However, if you then try to use the input to search for something else, nothing happens and you can't type in the input.

We are also missing a way to search for specific entities, and extra information to make differentiating between similar search results a bit easier: at least disambiguation and aliases.

### Solution
This PR

1. fixes the search page bug (the `value` prop of the input field was set, and thus 'controlled'; setting `defaultValue`instead solves that problem)
2. introduces a dropdown to search by entity type
3. Adds disambiguation and aliases on the search results.

### Areas of Impact
Search pages and routes on the client and server, as well as `SearchField` and `SearchResults`components.

### DIFF
--- a/src/client/input.js
+++ b/src/client/input.js
@@ -74,9 +74,9 @@ export default class Input extends Component {
 		if (Array.isArray(buttons)){
 			return buttons.map((button, index) => this.renderButton(button, index));
 		}
-		return this.renderButton(buttons);
+		return this.renderButton(buttons, 0);
 	}
-	renderButton(button, index=0) {
+	renderButton(button, index:number) {
 		return button && <InputGroup.Button key={`btn${index}`}>{ button }</InputGroup.Button>;
 	}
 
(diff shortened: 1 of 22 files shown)


MonkeyDo:
Here's a preview of what the modified search page looks like:
screencapture-localhost-9099-search-2018-10-31-20_30_04

src/client/input.js
```diff
@@ -10,155 +10,161 @@ import ReactTooltip from 'react-tooltip';
...
+		return addon && <InputGroup.Addon>{ addon }</InputGroup.Addon>;
+	}
+
+	renderButtons(buttons) {
+		if (Array.isArray(buttons)){
+			return buttons.map((button, index) => this.renderButton(button, index));
+		}
+		return this.renderButton(buttons);
+	}
+	renderButton(button, index=0) {
```
LordSputnik: Why use a default value for index? That seems unnecessarily risky to me - could result in multiple elements having the same key.

MonkeyDo: I've removed the default index as you suggested, and added a (simple) pagination system, if you want to review that :)

PR #218, opened 2018-11-06
Display Entity Aliases

Ge0rg3:
### Problem

> Currently in BookBrainz, entity aliases do not appear on an entity display page. Take for example the display page for [Lemony Snicket](https://bookbrainz.org/creator/cb5a7089-b36c-463e-9380-331ce94c2873) and compare it to the [edit page](https://bookbrainz.org/creator/cb5a7089-b36c-463e-9380-331ce94c2873/edit) for the same entity, and click on "edit 1 alias". You will see that the entity has 1 alias (Daniel Handler), that is not shown on the display page but visible when editing.

*-from the GCI task page.*

### Solution
Adds a helper function which returns the aliases.
This is then printed under the title (as `h5` is a block), if there actually are aliases.  
#### Before:
image
#### After:
image

### Areas of Impact
The [title page](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/title.js), and the [entity helper](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/helpers/entity).

### DIFF
--- a/src/client/helpers/entity.js
+++ b/src/client/helpers/entity.js
@@ -168,7 +168,7 @@ export function getEntityDisambiguation(entity) {
 	return null;
 }
 
-export function getEntityAliases(entity) {
+export function getEntitySecondaryAliases(entity) {
 	const aliases = entity.aliasSet.aliases
 		.filter(item => item.id !== entity.defaultAlias.id)
 		.map(item => item.name)
--- a/src/client/helpers/entity.js
+++ b/src/client/helpers/entity.js
@@ -170,14 +170,14 @@ export function getEntityDisambiguation(entity) {
 
 export function getEntityAlias(entity) {
 	const aliases = entity.aliasSet.aliases
-									.map(item => item.name)
-									.filter(item => item !== entity.defaultAlias.name)
-									.join(', ');
+		.map(item => item.name)
+		.filter(item => item !== entity.defaultAlias.name)
+		.join(', ');
 	if (entity.aliasSet.aliases.length > 1) {
 		return <h5>{aliases}</h5>;
-	} else {
-		return null;
 	}
+	
+	return null;
 }
 
 export function getEntityUrl(entity) {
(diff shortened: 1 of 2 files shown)


src/client/helpers/entity.js
```diff
@@ -168,6 +168,18 @@ export function getEntityDisambiguation(entity) {
 	return null;
 }
 
+export function getEntityAlias(entity) {
```
MonkeyDo: I'm nitpicking a bit, but could you rename the function to `getEntityAliases` ?

src/client/helpers/entity.js
```diff
@@ -168,6 +168,18 @@ export function getEntityDisambiguation(entity) {
...
+	const aliases = entity.aliasSet.aliases
+		.map(item => item.name)
+		.filter(item => item !== entity.defaultAlias.name)
```
MonkeyDo: I think it would be more precise and safer to use the `defaultAliasId` property to filter out the default alias (or perhaps `defaultAlias.id`, I doubt it matters).
I'm thinking that you could theoretically have two different aliases with the exact same name, and that would be useful to know when you display the page (so that for example a user can remove the excess alias).

That also means that you'll need to call `.filter`before you call `.map`, since after the map you won't have access to the item.id anymore.

src/client/helpers/entity.js
```diff
@@ -168,6 +168,18 @@ export function getEntityDisambiguation(entity) {
...
+		.join(', ');
+	if (entity.aliasSet.aliases.length > 1) {
+		return <h5>{aliases}</h5>;
```
MonkeyDo: I tried it out on my computer, and I find the aliases a tad small. Could you try an h4?

src/client/helpers/entity.js
Ge0rg3: Good idea--it might also be worth removing dupicates in the AliasEditor at some point?

PR #220, opened 2018-11-09
General code cleanup (WIP)

LordSputnik:
_This PR is a work in progress, and intended for incremental review - it is not yet ready to be pulled into the main branch._

### Problem
A lot of the BookBrainz site uses unreliable, outdated and/or verbose code, including long promise chains, not enough abstraction, inadequate unit testing, confusingly named variables and functions, and missing type annotation and documentation. This is the first of several PRs, and will mainly focus on reducing verbosity and complexity of code.

### Solution
Several improvements are planned. The first is to eliminate unnecessary conditional code and variables, while maintaining or improving code clarity. The second will involve replacing promise chains with asynchronous functions, to make code more readable and make poor exception handling more obvious.

### Areas of Impact
No functional changes to any code. The aim is to make each commit a simple refactor for ease of review.

### DIFF
--- a/src/server/routes/entity/entity.js
+++ b/src/server/routes/entity/entity.js
@@ -61,9 +61,86 @@ const entityComponents = {
 	work: WorkPage
 };
 
+/**
+ * Creates a new database record whenever an editor visits an entity, to allow
+ * tracking of popular entities and handling of the page visit achievement.
+ *
+ * @param {Object} orm An initialized bookbrainz-data ORM
+ * @param {string} bbid The BBID of the entity that was visited
+ * @param {?number} editorID The ID of the editor, if logged in, otherwise null
+ *
+ * @returns {Promise<Object|boolean>} An object containing achievement data, or
+ *          false if an error occured
+ */
+function logEntityVisit(orm: Object, bbid: string, editorID: ?number) {
+	const {EditorEntityVisits} = orm;
+
+	if (!editorID) {
+		return Promise.resolve(false);
+	}
+
+	return new EditorEntityVisits({bbid, editorId: editorID})
+		.save(null, {method: 'insert'})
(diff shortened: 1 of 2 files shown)


src/server/routes/entity/entity.js
```diff
@@ -61,9 +61,86 @@ const entityComponents = {
...
+ * unlock IDs to be used in the unlock alert displayed to the editor.
+ *
+ * @param {Object} orm An initialized bookbrainz-data ORM
+ * @param {number[]} alertIDs A list of achievement unlock IDs to process
+ * @param {?number} editorID The ID of the editor, if logged in, otherwise null
+ *
+ * @returns {Promise<Object[]|boolean>} A list containing achievement
+ *          information, or false if an error occured
+ */
+function processAlertIDs(orm: Object, alertIDs: number[], editorID: ?number) {
```
MonkeyDo: Is there a reason why editorID is optional, considering the if block at lines 109-111 ?

LordSputnik: This is because the `displayEntity` function calls `processAlertIDs`, and there isn't always a logged-in editor when `displayEntity` is called. We could move the condition into `displayEntity` - wouldn't be difficult. Previously this check was on line 78.

MonkeyDo: Yes, that makes perfect sense.
I think I was just tired when I wrote this :)

LordSputnik:
I haven't abandoned this - things are just a little busy at the moment, but hopefully I should be able to finish before the end of the year. Perhaps I can limit the scope a little, and then we can merge earlier, and then I'll open another PR when I have something else to contribute.

MonkeyDo:
I'm in favor of limiting the scope and creating subsequent PRs, I think it'll help limit merge conflicts as much as possible.

PR #222, opened 2018-11-16
Search page improvements

MonkeyDo:
- Filtering out of null results on search pagination (BB-308)
- Call to action when no results found:
capture d ecran 2018-11-16 a 16 30 23


src/client/components/pages/search.js
```diff
@@ -83,7 +83,7 @@ class SearchPage extends React.Component {
 			.then((data) => {
 				this.setState(prevState => ({
 					nextEnabled: data.length >= prevState.size,
-					results: data
+					results: data.filter(Boolean)
```
leftmostcat: This reads a little confusing. I'd like it better if there were a function to explicitly check expected values rather than relying on what I find to be kind of unintuitive behaviors of the `Boolean` ctor in JS.

src/server/routes/search.js
```diff
@@ -51,7 +51,7 @@ router.get('/', (req, res, next) => {
 
 	search.searchByName(orm, query, collection, size || resultsPerPage, from)
 		.then((entities) => ({
-			initialResults: entities,
+			initialResults: entities.filter(Boolean),
```
leftmostcat: As above, I'd rather make this explicit.

src/client/components/pages/search.js
MonkeyDo: Fair enough, I'll get that replaced.

PR #226, opened 2018-11-26
Prevent Dropdown closing on click

TheFaR8:
https://tickets.metabrainz.org/browse/BB-269

### Problem

As shown on JIRA, the current add navbar is closing after clicking. 

### Solution

By this PR, it enables users to use "Ctrl + Click" in order to add some entities in one Add button

### Areas of Impact

Site interface

### DIFF
--- a/src/client/containers/layout.js
+++ b/src/client/containers/layout.js
@@ -47,6 +47,20 @@ class Layout extends React.Component {
 		this.renderNavHeader = this.renderNavHeader.bind(this);
 	}
 
+	handleDropdownToggle(newValue) {
+		if (this._forceOpen) {
+			this.setState({menuOpen: true});
+			this._forceOpen = false;
+		}
+		else {
+			this.setState({menuOpen: newValue});
+		}
+	}
+
+	handleDropdownClick() {
+		this._forceOpen = true;
+	}
+
 	renderNavHeader() {
 		const {homepage} = this.props;
 
@@ -100,26 +114,28 @@ class Layout extends React.Component {
 						<NavDropdown
 							eventKey={1}
 							id="create-dropdown"
+							open={this.state.menuOpen}
 							title={createDropdownTitle}
+							onToggle={this.handleDropdownToggle}
 						>
-							<MenuItem href="/work/create">
+							<MenuItem href="/work/create" onSelect={this.handleDropdownClick}>
 								{genEntityIconHTMLElement('Work')}
 								Work
 							</MenuItem>
-							<MenuItem href="/edition/create">
+							<MenuItem href="/edition/create" onSelect={this.handleDropdownClick}>
 								{genEntityIconHTMLElement('Edition')}
 								Edition
 							</MenuItem>
-							<MenuItem href="/publication/create">
+							<MenuItem href="/publication/create" onSelect={this.handleDropdownClick}>
 								{genEntityIconHTMLElement('Publication')}
 								Edition Group
 							</MenuItem>
 							<MenuItem divider/>
-							<MenuItem href="/creator/create">
+							<MenuItem href="/creator/create" onSelect={this.handleDropdownClick}>
 								{genEntityIconHTMLElement('Creator')}
 								Author
 							</MenuItem>
-							<MenuItem href="/publisher/create">
+							<MenuItem href="/publisher/create" onSelect={this.handleDropdownClick}>
 								{genEntityIconHTMLElement('Publisher')}
 								Publisher
 							</MenuItem>


src/client/containers/layout.js
```diff
@@ -47,6 +47,20 @@ class Layout extends React.Component {
...
+			this.setState({menuOpen: true});
+			this._forceOpen = false;
+		}
+		else {
+			this.setState({menuOpen: newValue});
+		}
+	}
+
+	handleDropdownClick() {
+		this._forceOpen = true;
```
MonkeyDo: When I run the code and test the ctrl+click, I get this error in the browser console:
`Uncaught TypeError: Cannot set property '_forceOpen' of undefined
    at handleDropdownClick`

This is an issue with React: the functions defined on the class do not have access to the `this` context for that class by default.
It's cumbersome, and you'll want to remember it to save yourself some looking around for the cause of an error, but here's the way to fix it: you need to bind the function to the class `this` context.
If you look in the class constructor, you'll notice we do that with other functions, in this format: `this.renderNavHeader = this.renderNavHeader.bind(this);`

src/client/containers/layout.js
```diff
@@ -100,26 +114,28 @@ class Layout extends React.Component {
 						<NavDropdown
 							eventKey={1}
 							id="create-dropdown"
+							open={this.state.menuOpen}
```
MonkeyDo: A good practice, when you add an item to the state, is to always give it an initial value.
That way, someone else reading your code will know what the initial state is and what to expect.

In this case, you'll want to add the initial value for `menuOpen` in the constructor, where `this.state` is defined

src/client/containers/layout.js
```diff
@@ -100,26 +114,28 @@ class Layout extends React.Component {
...
 							title={createDropdownTitle}
+							onToggle={this.handleDropdownToggle}
 						>
-							<MenuItem href="/work/create">
+							<MenuItem href="/work/create" onSelect={this.handleDropdownClick}>
```
MonkeyDo: You can bind the `onSelect` event callback to the parent `NavDropdown` item, exactly like you did for `onToggle`.
For the same result, it will be more efficient since you create only one event handler instead of five.

src/client/containers/layout.js
```diff
@@ -47,6 +47,20 @@ class Layout extends React.Component {
 		this.renderNavHeader = this.renderNavHeader.bind(this);
 	}
 
+	handleDropdownToggle(newValue) {
+		if (this._forceOpen) {
```
MonkeyDo: I would suggest renaming `forceOpen`to `keepMenuOpen` for clarity.


Is there a specific reason why you decided to use the component state for `menuOpen` but not for `forceOpen`?
I did try using the state, expecting that `handleDropdownToggle` would be called before `handleDropdownClick` setting the state had a chance to propagate, and sure enough that's what happened.
Was that a conscious decision you made for the same reason?

MonkeyDo:
One more thing !
Please review the [contribution guidelines](https://github.com/bookbrainz/bookbrainz-site/blob/master/CONTRIBUTING.md) for how to construct a proper commit title; there should be a prefix like `feat: ` or `feat(navbar): ` before your title.

TheFaR8:
Sure

src/client/containers/layout.js
TheFaR8: Done

src/client/containers/layout.js
TheFaR8: it was, yeah
but since you suggested this, I began to think that this is inconsistent, so I am changing it

src/client/containers/layout.js
```diff
@@ -42,23 +42,25 @@ const {Alert, MenuItem, Nav, Navbar, NavItem, NavDropdown} = bootstrap;
...
+		this.handleDropdownToggle = this.handleDropdownToggle.bind(this);
+		this.handleDropdownClick = this.handleDropdownClick.bind(this);
 	}
 
 	handleDropdownToggle(newValue) {
-		if (this._forceOpen) {
+		if (this.state.keepMenuOpen) {
 			this.setState({menuOpen: true});
-			this._forceOpen = false;
+			this.keepMenuOpen = false;
```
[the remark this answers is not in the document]
TheFaR8: Added 
(Previously, since I'm in rush to do something else, I forgot to change this and quickly shutted down the computer.. Sorry)

MonkeyDo:
Hi @TheFaR8 ,
You're doing great and we're almost there !

I'm sorry I didn't explain myself well in one of earlier my comments.
When I tried using the state for `keepMenuOpen`, it didn't work for me. I am seeing the same issue with the code changes you've pushed.
That being said, after some thought I have been able to make it work.

Now, for some details about the state issue:
If you set `keepMenuOpen` on the state in `handleDropdownClick`, which seems like a logical thing to do, you'll find that when `handleDropdownToggle` is called (almost instantly after), React didn't have the time yet to set `keepMenuOpen` correctly.
There is a way to ensure that the state has been written by passing a callback as the [second argument of the `setState` function](https://reactjs.org/docs/react-component.html#setstate).

Resolution:
The [dropdown's onSelect function](https://react-bootstrap.github.io/components/dropdowns/#menu-item-props) takes two arguments: `(eventKey, event)`.
In `handleDropdownClick` we need to capture the click event and stop it from continuing on to trigger `handleDropdownToggle` automatically.
You can do that by calling `event.stopPropagation();`
Then you can call `setState` as you did, and pass `this.handleDropdownToggle` as the callback function, ensuring that `keepMenuOpen` will be set correctly before calling `handleDropdownToggle`.


Let me know if this is clear enough for you or if you have any questions.
I'm happy we can take the time to make this nice and clean!

I have another comment I will add inline with the code :)

src/client/containers/layout.js
```diff
@@ -42,9 +42,25 @@ const {Alert, MenuItem, Nav, Navbar, NavItem, NavDropdown} = bootstrap;
...
 		this.renderNavContent = this.renderNavContent.bind(this);
 		this.renderNavHeader = this.renderNavHeader.bind(this);
+		this.handleDropdownToggle = this.handleDropdownToggle.bind(this);
+		this.handleDropdownClick = this.handleDropdownClick.bind(this);
+	}
+
+	handleDropdownToggle(newValue) {
+		if (this.state.keepMenuOpen) {
+			this.setState({menuOpen: true});
+			this.setState({keepMenuOpen: false});
```
MonkeyDo: `setState` takes an object with the variables you want to change. That object can have multiple keys, so you can set both variables in one go:
`this.setState({menuOpen: true, keepMenuOpen: false});`

TheFaR8:
Sorry, I don't actually get what you mean. 
I understand the addition of `event.stopPropagation()` here, but I don't actually catch (most of) things that appear afterwards. So, can you please make it simple? (My English is not very good)
Thanks

MonkeyDo:
Hi again,

Sorry if I'm going too fast or using complicated english !

It is the function `handleDropdownClick` itself that you want to modify to:

1. accept two arguments: `handleDropdownClick(eventKey, event) {…}`
2. call `event.stopPropagation()`
3 modify the call to `setState` to add a callback as second argument: `this.setState({keepMenuOpen: true}, this.handleDropdownToggle);`

TheFaR8:
Ahhh
okay
I got it now
Thanks

src/client/containers/layout.js
TheFaR8: Done

MonkeyDo:
That's working great now, and it is nice and clean !
Well done and thank you !

It wasn't intuitive, but I hope the event catching and setState callback mechanism makes sense to you now.

P.S: For what it's worth, I find your english quite good! I hadn't realised you are not a native speaker.

TheFaR8:
Thank you for the compliment (late response)
and by the way, should I add the license on the top of the code?

PR #230, opened 2019-01-15
fix(style): Resolve overlapping of container and footer class in mobi…

akhilesh26:
…le view.

### Problem

Footer and container class overlapped in mobile view. 
Shows in the screenshots.

### Solution

Changed view shows as:

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -54,7 +54,6 @@ The production configuration for webpack needs to be finalized before using this
 
 body {
 	margin-top: 64px;
-	margin-bottom: 64px;
 }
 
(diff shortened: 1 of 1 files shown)


MonkeyDo:
I just realised there is an issue with the footer in desktop size, where the bottom margin pushes the footer up:

capture d ecran 2019-01-16 a 13 44 21

After a quick test, I think you need to modify `.footer` css class to remove the `position` attribute (or set it to `relative` and reinstate `bottom: 0`.

Could you please have a look?

akhilesh26:
In my system, It is working right. But I will make changes as you suggest!

footer problem

PR #232, opened 2019-01-16
fix #BB-318 and make same as shown in search result

akhilesh26:
### Solution

Use brackets and small size for consistency as a search result.
bb318


MonkeyDo:
Thanks, that's great!

You can also close the ticket for BB-318 !

PR #233, opened 2019-01-16
docs: bookbrainz setup locally using docker toolbox

kartikpandey2:
### Problem
Setup bookbrainz locally using docker toolbox

### Solution
Added a note in README.md in "Running dependencies using Docker" section

### Areas of Impact
Nothing

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ If you wish, you can instead [install the database and search dependencies on yo
 and/or [run the NodeJS server locally](./NODEJS_SETUP.md) while using dockerized dependencies.
 Running the NodeJS server locally would be easier if you want to use a debugger, for instance.
 
-Note: If you are using docker-toolbox you need to replace [elasticsearch:9200](/config//config.json.example#L30) with Ip address of your docker-machine(192.168.99.100) in `config.json`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,6 +59,8 @@ If you wish, you can instead [install the database and search dependencies on yo
 and/or [run the NodeJS server locally](./NODEJS_SETUP.md) while using dockerized dependencies.
 Running the NodeJS server locally would be easier if you want to use a debugger, for instance.
 
+Note: If you are using docker-toolbox you need to replace [elasticsearch:9200](/config//config.json.example#L30) with Ip address of your docker-machine(192.168.99.100) in `config.json`.
```
MonkeyDo: As far as I understand, the IP address can be different from that default.
Maybe change the `(192.168.99.100)`to say `(default: 192.168.99.100)`, and add another line to explain how to find the ip address.
(If I'm not mistaken, you need to run `docker-machine ip default` on the host. Is that correct?)

There is also a double `/` in `/config//config.json.example#L30`, but the link still works.

kartikpandey2: Done

MonkeyDo:
Thanks @kartikpandey2 for sharing your knowledge with future users!

PR #241, opened 2019-01-22
feat(edition detail): add format and isbn in edition table to fix #BB-229

akhilesh26:
In the edition group page and publisher page a table of edition was showing.
But in this table only edition name and release date was showing which not sufficient to distinguish between editions.
to solve above problem, add format of edition and isbn with his link by this commit.

### Problem

Less information in edition to distinguish bw them.

### Solution

Add isbn and format of an edition in the table.
Now  publisher page:  
publisher
Publication page:  
publication

### Areas of Impact

**Client-side**  
`src/client/components/pages/entities/edition-table.js`
`src/client/helpers/entity.js`
**Server-side**  
`src/server/routes/entity/publication.js`
`src/server/routes/entity/publisher.js`

### DIFF
--- a/src/client/helpers/entity.js
+++ b/src/client/helpers/entity.js
@@ -214,19 +214,14 @@ export function getSortNameOfDefaultAlias(entity) {
 export function getISBNOfEdition(entity) {
 	if (entity.identifierSetId) {
 		const {identifiers} = entity.identifierSet;
-		let isbn = null;
-		isbn = identifiers.find(identifier => identifier.type.label === 'ISBN-13');
-		if (!isbn) {
-			isbn = identifiers.find(identifier => identifier.type.label === 'ISBN-10');
-		}
-		return isbn;
+		return identifiers.find(
+			identifier =>
+				identifier.type.label === 'ISBN-13' || identifier.type.label === 'ISBN-10'
+		);
 	}
 	return null;
 }
 
 export function getEditionFormat(entity) {
-	if (entity.formatId) {
-		return entity.editionFormat.label;
-	}
-	return '?';
+	return (entity.editionFormat && entity.editionFormat.label) || '?';
 }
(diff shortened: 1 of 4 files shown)


src/client/helpers/entity.js
```diff
@@ -210,3 +210,23 @@ export function genEntityIconHTMLElement(entityType, size = '', margin = true) {
 export function getSortNameOfDefaultAlias(entity) {
 	return entity.defaultAlias.sortName;
 }
+
+export function getISBNOfEdition(entity) {
+	if (entity.identifierSetId) {
+		const {identifiers} = entity.identifierSet;
+		let isbn = null;
+		isbn = identifiers.find(identifier => identifier.type.label === 'ISBN-13');
```
MonkeyDo: This can be refactored as below to avoid running .find twice:

`identifiers.find(identifier => identifier.type.label === 'ISBN-13' || identifier.type.label === 'ISBN-10' );`

src/client/helpers/entity.js
```diff
@@ -210,3 +210,23 @@ export function genEntityIconHTMLElement(entityType, size = '', margin = true) {
...
+		if (!isbn) {
+			isbn = identifiers.find(identifier => identifier.type.label === 'ISBN-10');
+		}
+		return isbn;
+	}
+	return null;
+}
+
+export function getEditionFormat(entity) {
+	if (entity.formatId) {
```
MonkeyDo: I would be more comfortable, for legibility, with something like that:
```
if (entity.editionFormat) {
[...]
```

Or even more succinct:
`return (entity.editionFormat && entity.editionFormat.label) || '?';`

src/client/helpers/entity.js
MonkeyDo: Acutally, I think you can `return` that `identifiers.find` directly.
It will return undefined if it doesn't find any, and considering how you're using it in edition-table.js that will work fine.

MonkeyDo:
That's another good addition, much better user experience.

MonkeyDo:
Thanks a lot! You can go ahead and close yet another ticket :)

MonkeyDo:
While testing this PR a non-threatening error surfaced. I commented on the wrong PR, sorry. Have a look here: https://github.com/bookbrainz/bookbrainz-site/pull/239#discussion_r251040245

PR #244, opened 2019-01-25
fix(propType warnings): remove isRequired and set default props to fix propType warnings.

akhilesh26:
### Problem

The warning occurs if propType marked as isRequire and value is null or undefined. 

### Solution

Remove isRequired from propTypes and set the default props.


MonkeyDo:
Great, thanks ! :)

PR #249, opened 2019-02-04
fix(editor): Fix BB-332: Run name deduplication search on component mount

MonkeyDo:
### Problem

When creating a new Edition from a Work, the nameSection is copied over from the Work.
The deduplication search mechanism is not run in this case, and does not enforce a disambiguation even though another edition of the same name exists.

### Solution

Turn NameSection functionnal component into class in order to use the `componentDidMount` lifecycle hook.
This is also a more elegant solution to the stopgap fix that was in place to fix an issue if a user types in the input before component is mounted (i.e. with slow loading JS)

### Areas of Impact
`src/client/entity-editor/name-section/name-section.js`

### DIFF
--- a/src/client/entity-editor/name-section/name-section.js
+++ b/src/client/entity-editor/name-section/name-section.js
@@ -73,160 +73,171 @@ import {getEntityDisambiguation} from '../../helpers/entity';
  *        the disambiguation is changed.
  * @returns {ReactElement} React element containing the rendered NameSection.
  */
-function NameSection({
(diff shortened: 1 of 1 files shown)


akhilesh26:
I am sorry @MonkeyDo for this issue. I fetched your PR to my local and its work fine for me. This is a good catch and very nicely fixed.

MonkeyDo:
It was no fault of your code !
Just a case I hadn't thought of in the ticket description.

Thanks for helping me test it!

PR #250, opened 2019-02-07
feat(workTable): Add work table on the edition page and edition table on the work page. #BB-330

akhilesh26:
.

1. Work table contains all the work contained by the edition.
2. Add edition table in the work page, the table contains all the editions who containnig that work.
3. Remove some unused lines of code of import section.

### DIFF
--- a/src/client/components/pages/entities/work-table.js
+++ b/src/client/components/pages/entities/work-table.js
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2019  Akhilesh Kumar (@akhilesh26)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import * as entityHelper from '../../../helpers/entity';
+
+import Icon from 'react-fontawesome';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+const {Button, Table} = bootstrap;
+
+const {getEntityDisambiguation} = entityHelper;
+
+function WorkTableRow({work}) {
+	const name = work.defaultAlias ? work.defaultAlias.name : '(unnamed)';
+	const disambiguation = getEntityDisambiguation(work);
+	const workType = work.workType ? work.workType.label : '?';
+	const workBBID = work.bbid;
+	return (
+		<tr>
+			<td>
+				<a href={`/Work/${workBBID}`}>{name}</a>
+				{disambiguation}
+			</td>
+			<td>{workType}</td>
+		</tr>
+	);
+}
+WorkTableRow.displayName = 'WorkTableRow';
+WorkTableRow.propTypes = {
+	work: PropTypes.object.isRequired
+};
+
+function WorkTable({entity}) {
+	return (
+		<div>
+			<h2>Works</h2>
+			<Table striped>
+				<thead>
+					<tr>
+						<th>Name</th>
+						<th>Type</th>
+					</tr>
+				</thead>
+				<tbody>
+					{
+						entity.works.map((work) => (
+							<WorkTableRow
+								key={work.bbid}
+								work={work}
+							/>
+						))
+					}
+				</tbody>
+			</Table>
+			<Button
+				bsStyle="success"
+				className="margin-top-d15"
+				href={`/work/create?${
+					entity.type.toLowerCase()}=${entity.bbid}`}
+			>
+				<Icon className="margin-right-0-5" name="plus"/>Add Work
+			</Button>
+		</div>
+	);
+}
+WorkTable.displayName = 'WorkTable';
+WorkTable.propTypes = {
+	entity: PropTypes.object.isRequired
+};
+
+export default WorkTable;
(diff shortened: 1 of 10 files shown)


akhilesh26:
Hi @MonkeyDo,
Please review this PR when you have time. If you will review this PR by today I will make the changes tomorrow morning. Otherwise I am on journey bw 8-11 Feb. But don't worrie, review this according to your time schedule.
Thanks :)

src/client/helpers/entity.js
```diff
@@ -225,3 +226,59 @@ export function getISBNOfEdition(entity) {
...
+// relationshipTypeId = 10 refers the relation (<Work> is contained by <Edition>)
+const relationshipTypeId = 10;
+
+/**
+ * Filter the relationship according to relationshipTypeId
+ *
+ * @param {object} entity - Entity with all relationships
+ * @returns {object} retrun the all relationshiops after removing the relatioship with relationshipTypeId = 10
+ */
+export function getFilteredRelationship(entity) {
```
MonkeyDo: I think it should be easy to make this function more reusable and pass a second `relationshipTypeId` argument instead of hardcoding it.

src/client/helpers/entity.js
```diff
@@ -225,3 +226,59 @@ export function getISBNOfEdition(entity) {
...
+	return entity.relationships.filter((relation) => relation.typeId !== relationshipTypeId);
```
MonkeyDo: I'd suggest a defensive approach and checking if entity.relationships is an array, returning and empty array if it is not the case.

src/client/helpers/entity.js
```diff
@@ -225,3 +226,59 @@ export function getISBNOfEdition(entity) {
...
+
+/**
+ * Get all works constained by entity according to relationshipTypeId
+ *
+ * @param {object} entity - Edition with all relationships
+ * @returns {object} Return all the works of Edition with relationshipTypeId = 10
+ */
+export function getWorksContainedByEdition(entity) {
+	let works = null;
+	if (entity.relationships) {
```
MonkeyDo: Similar to the above comment, I'd checking for `Array.isArray(entity.relationships)` here and in getEditionsContainsWork

src/client/helpers/entity.js
MonkeyDo: If you're interested, I think it would be quite neat to use an enum that can be used when calling this function, as well as other places that rely on a relationshipTypeId number.

It is out of the scope of this PR, though, and I'll create a separate ticket for it.
If you're interested in diving a bit more into flow (the type system BookBrainz uses), you can have a look at https://alligator.io/react/flow-enums/#creating-an-enum-from-an-existing-value

src/client/components/pages/entities/work-table.js
```diff
@@ -0,0 +1,89 @@
...
+WorkTableRow.displayName = 'WorkTableRow';
+WorkTableRow.propTypes = {
+	work: PropTypes.object.isRequired
+};
+
+function WorkTable({entity}) {
+	return (
+		<div>
+			<h2>Works</h2>
+			<Table striped>
```
MonkeyDo: If there are no works to be listed, I would replace the table with a more explicit call to action, like the one in the [search results|https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/search-results.js#L44], and perhaps a larger button

akhilesh26:
Thanks, @MonkeyDo! for review this PR and helping in some designing things for the work table and edition table. I created all of the changes. Any other changes also welcome!

src/client/helpers/entity.js
```diff
@@ -225,3 +226,61 @@ export function getISBNOfEdition(entity) {
...
+
+
+/**
+ * Get all works constained by entity according to relationshipTypeId
+ *
+ * @param {object} entity - Edition with all relationships
+ * @param {number} relationshipTypeId - filter the relationship according to this typeId
+ * @returns {object} Return all the works of Edition with relationshipTypeId = 10
+ */
+export function getWorksContainedByEdition(entity, relationshipTypeId) {
```
MonkeyDo: This function and the one below also need renaming, as the refactor has made them reusable for other entity types.

How about `getRelationshipTargetByTypeID` and `getRelationshipSourceByTypeID` ?
Please also review the documentation comments for both methods to reflect the changes.

To finish making it reusable, rename the variables you use in the methods (`works`/`editions` => `target`/`source`).
Instead or returning a full object with bbid and type, return only the filtered array or an empty array

That also means you will need to refactor WorksTable and EditionsTable to take new separate `works` and `editions` property respectively, instead or relying on `entity.works` and `entity.editions`.
(That will result in some silly verbose use for the `PublisherDisplyPage` for example, which will end up like this: `<EditionTable entity={entity} editions={entity.editions}/>`, but so be it, I find it more declarative.)

How does that sound to you?
A bit of a big refactor this late in the PR, but I hadn't grasped how all the pieces worked together.

MonkeyDo: I wanted to add that I'm not asking this for the sake of code purity: I have in mind other display pages where we will want to show certain relationships in their own section like we do here.
I think it makes a lot of sense to show a "main" type of relationship for an entity in a more prominent manner, and users will be expecting that.

akhilesh26:
@MonkeyDo thanks for the detailed review. I created the changes that you mention above. Sorry for all little mistakes. More comments also welcome for any improvements.
Thanks a lot!

src/server/helpers/utils.js
```diff
@@ -226,14 +226,15 @@ export function truncateTables(Bookshelf: Object, tables: Array<string>) {
...
  *
  * @param {string} modelType - type of the model or entity
- * @returns {array} array of addition relations
+ * @returns {array} array of additional relations
  */
 export function getAdditionalRelations(modelType) {
+	const additionalRelations = [];
 	if (modelType === 'Work') {
-		return ['disambiguation', 'workType'];
+		return additionalRelations.concat(['disambiguation', 'workType']);
```
MonkeyDo: Sorry, I wasn't very clear with this. 
I didn't mean to say to use concat here, but rather to always return an array, so that in another file I can call `myArray.concat(getAdditionalRelations(...))`.

So you can return array as you did before, but change the last line to be `return []`.

MonkeyDo:
Thanks for taking all the changes on board so easily, and so quick too.

akhilesh26:
@MonkeyDo,
I created change as you commented. Thanks!

MonkeyDo:
Great, thanks for this feature @akhilesh26, it's ready to merge and I'll deploy it to test.bookbrainz in the same breath. Eager to see it online !

PR #253, opened 2019-03-04
Fix button layout on mobile in the search page

aswinmprabhu:
### Problem
[BB-92](https://tickets.metabrainz.org/browse/BB-92?jql=project%20%3D%20BB%20AND%20resolution%20%3D%20Unresolved%20AND%20assignee%20in%20(EMPTY)%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC)

### Solution

Stack the buttons on the search page vertically when on mobile to prevent text from cutting off.

### Areas of Impact

Search page

### DIFF
--- a/src/client/components/pages/parts/search-results.js
+++ b/src/client/components/pages/parts/search-results.js
@@ -27,145 +27,143 @@ import {genEntityIconHTMLElement} from '../../../helpers/entity';
 
 const {Button, ButtonGroup, Col, Row, Table} = bootstrap;
 
-function SearchResults(props) {
-	const noResults = !props.results || props.results.length === 0;
-	if (noResults) {
+class SearchResults extends React.Component {
+	constructor() {
+		super();
+		this.state = {
+			isSmallScreen: true
+		};
+	}
+
+	componentDidMount() {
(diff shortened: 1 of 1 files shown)


aswinmprabhu:
My bad. I did not think of flexbox because I was stuck on the vertical prop of ButtonGroup.
> I can help in setting that up if you wish!

Yes. I use VSCode. I used prettier (it should auto detect the .editorconfig). Even after disabling prettier, the problem persists.

MonkeyDo:
@aswinmprabhu I use VSCode too, that simplifies things :)

Try changing these settings for the workspace, see if that solves it:
```
"editor.formatOnPaste": false,
"editor.formatOnSave": false,
"editor.formatOnType": false
```


I would also recommend installing the ESLint extension, which will help you conform to the style rules for the project by showing errors/warnings on the file and in the 'Problems' tab (menu View > Problems):
https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

That way, you don't have to run ESLint manually (ESLint errors will make the continuous integration tests fail), which can still be done by running the command `npm run lint`

PR #255, opened 2019-03-06
Fix button layout and input margin on search page

aswinmprabhu:
### Problem

[BB-92](https://tickets.metabrainz.org/browse/BB-92?jql=project%20%3D%20BB%20AND%20resolution%20%3D%20Unresolved%20AND%20assignee%20in%20(EMPTY)%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC)

### Solution

Stack the buttons on the search page vertically when on mobile to prevent text from cutting off.  
Add some margin to the search input on mobile.

### Areas of Impact

Search page

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -35,7 +35,7 @@ The production configuration for webpack needs to be finalized before using this
 @import "~react-virtualized/styles.css";
 @import "~react-virtualized-select/styles.css";
  */
- 
+
 /* Modifications to Lobes */
 /* ====================== */
 @brand-secondary: @bookbrainz;
@@ -246,7 +246,7 @@ html {
 	border-radius: @input-border-radius; // Note: This has no effect on <select>s in some browsers, due to the limited stylability of <select>s in CSS.
 	.transition(~"border-color ease-in-out .15s, box-shadow ease-in-out .15s");
 	.form-control-focus();
-	
+
 }
 
 .Select-placeholder {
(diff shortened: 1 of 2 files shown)


src/client/stylesheets/style.less
```diff
@@ -325,11 +325,26 @@ hr.wide {
...
\ No newline at end of file
+}
+
+@media only screen and (max-width: 600px) {
+	#searchpage-button-group {
+		display: flex;
+		flex-wrap: wrap;
+	}
+
+	#searchpage-button-group .btn {
```
MonkeyDo: I think you could generalise it to `#searchpage-button-group > *`, any direct child of the button-group.
That way we don't rely on the class name.

src/client/stylesheets/style.less
```diff
@@ -325,11 +325,26 @@ hr.wide {
...
+		width: 90%;
```
MonkeyDo: This introduces a margin on the right.
100% looks fine to me; what was the reason for 90%?

src/client/stylesheets/style.less
```diff
@@ -325,11 +325,26 @@ hr.wide {
...
 		background-color:@dropdown-link-hover-bg;
 	  }
 	}
-}
\ No newline at end of file
+}
+
+@media only screen and (max-width: 600px) {
+	#searchpage-button-group {
+		display: flex;
```
MonkeyDo: It turns out that, considering the media query, you don't even need flex for what you are doing here.

You can remove this css block if you also remove, in the js file, the `justified` property on the `ButtonGroup` as well as `lg={6} lgOffset={3}` on its parent, and finally find and remove `style={{lineHeight: '1em', whiteSpace: 'normal'}` for the Edition Group button.

That way, I think we achieve the cleanest fix.

MonkeyDo:
That's working very nicely now, and it's even a cleaner situation than before.

Nicely done !


P.S: Please have a look at the [contribution guide](https://github.com/bookbrainz/bookbrainz-site/blob/master/CONTRIBUTING.md) for how to format commit messages. `Make requested changes` won't be very clear once I merge it to the rest of the repository history :p

PR #260, opened 2019-03-08
doc(revision): update to the configuration section of README.md page

rihbyne:
### Problem
This section was missing a detail on callback URL required to request authorization code when register bookbrainz as developer application with MB.

### Solution
When registering the developer applicaition, the callback URL needs to match exactly the one specified in config/config.json

### Areas of Impact
README.md


rihbyne:
@MonkeyDo thank you for the kind words! :)

PR #261, opened 2019-03-08
fix(editor): Added link and entity type in the search entity while ad…

pv-912:
Solved issue [BB-274](https://tickets.metabrainz.org/browse/BB-274) as well as [BB-275](https://tickets.metabrainz.org/browse/BB-275)

Link the URL to the respective searched entity and also added to show the type of the entity when adding the relationship.


src/client/entity-editor/common/entity.js
```diff
@@ -19,49 +19,52 @@
...
 	const contents = (
 		<span>
 			{
 				type && genEntityIconHTMLElement(type)
 			}
 			{' '}
-			{nameComponent}
-			{' '}
+			<a href={url} rel="noopener noreferrer" target="_blank">
+				{text} &nbsp; ( {type} )
```
MonkeyDo: You'll still want to use the `nameComponent` instead of `text`, as it defaults to `unnamedText` if `text` is empty.

MonkeyDo: I don't think adding the entity type in full text adds anything, considering the name is already preceded by the entity icon.
I don't think this solves BB-274, as you're not adding any other information that could help identify two entities of same type with the same name (see use-case in the ticket description). 
I think it's a bigger job, and you'll need to show different info depending on the type of entity
(see [this ticket comment](https://tickets.metabrainz.org/browse/BB-274?focusedCommentId=48269&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-48269)).

src/client/entity-editor/common/entity.js
```diff
@@ -19,49 +19,52 @@
...
 
-import Icon from 'react-fontawesome';
+// import Icon from 'react-fontawesome';
 import React from 'react';
 import {genEntityIconHTMLElement} from '../../helpers/entity';
 
 
 type EntityProps = {
 	disambiguation?: ?string,
-	link?: string | false,
```
MonkeyDo: Missing the new `id` property

src/client/entity-editor/common/entity.js
```diff
@@ -19,49 +19,52 @@
...
 	unnamedText?: string
 };
 
 function Entity(
-	{disambiguation, link, text, type, unnamedText}: EntityProps
+	{disambiguation, id, text, type, unnamedText}: EntityProps
 ) {
-	const nameComponent = text || <i>{unnamedText}</i>;
+	const url = type === 'Area' ?
+		`//musicbrainz.org/area/${id}` :
```
MonkeyDo: Are you sure this link will work? I can't seem to find a way to test it.

src/client/entity-editor/common/entity.js
```diff
@@ -19,49 +19,52 @@
...
+	// const nameComponent = text || <i>{unnamedText}</i>;
 	const contents = (
 		<span>
 			{
 				type && genEntityIconHTMLElement(type)
 			}
 			{' '}
-			{nameComponent}
-			{' '}
+			<a href={url} rel="noopener noreferrer" target="_blank">
```
MonkeyDo: For me, clicking on one of these links only selects the item and the dropdown closes, instead of opening the link.
I think you'll need to bind an onclick handler to your <a> tag and use `event.preventDefault()` inside the handler for it to work.

I should note that one the dropdown is closed, I can click the link successfully.

src/client/entity-editor/common/entity.js
```diff
@@ -19,49 +19,52 @@
...
 				disambiguation &&
 				<span className="disambig"><small>({disambiguation})</small></span>
 			}
 		</span>
 	);
 
-	if (link) {
-		return <a href={link}>{contents}</a>;
-	}
+	// if (link) {
```
MonkeyDo: You forgot to clean up after yourself :p

src/client/entity-editor/common/entity.js
pv-912: Okay, I will try to add more info according to the entity.

src/client/entity-editor/common/entity.js
pv-912: Sure, I will add it.

src/client/entity-editor/common/entity.js
pv-912: I have taken this from here https://github.com/bookbrainz/bookbrainz-site/blob/4498bddf016efb0b96fbd4f56a9e4af9c7a56f43/src/client/components/pages/parts/search-results.js#L109

src/client/entity-editor/common/entity.js
pv-912: one the dropdown is closed, I can click the link successfully.... yes that is what I have tried to do. And okay I will try to bind the onclick handler.

src/client/entity-editor/common/entity.js
pv-912: Actually, I thought may it have some use in future that y I don't delete it. I will clean it in next commit.

src/client/entity-editor/common/entity.js
```diff
@@ -34,31 +35,30 @@ type EntityProps = {
...
 		<span>
 			{
 				type && genEntityIconHTMLElement(type)
 			}
 			{' '}
-			<a href={url} rel="noopener noreferrer" target="_blank">
-				{text} &nbsp; ( {type} )
-			</a>
-
+			<section onClick={eventHandlerk(event, url)}>
```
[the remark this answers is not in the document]
MonkeyDo: Here, in your onClick property, you actually call the function eventHandlerk instead of passing the function itself (`onClick={function(arguments)}` instead of `onClick={eventHandlerk}`).

That's why a new tab was opening: the function was called everytime the component was rendered!
That's an important distinction to keep in mind. I know I made that mistake before!

You don't need to pass arguments. The onClick function will call your callback with the first argument `event`, and the `url` is defined in the render function, no need to pass it to `eventHandlerk`.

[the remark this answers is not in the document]
MonkeyDo: A few other comments:

- You previously had an `<a>` tag instead of the `<section>`. It was better with an `<a>`: specific styling is applied that lets the user know it's a clickable link.
- What does the `k` stand for at the end of `eventHandlerk`?
- I touched on the subject before, but you'll want to do all these modifications on a new component (I suggest the name `SearchResultEntity`) because this one is used in other places and we don't want these modifications everywhere.

src/client/entity-editor/common/linked-entity.js
```diff
@@ -20,50 +20,63 @@
...
+		event.preventDefault();
 		if (url) {
 			window.open(url, '_blank');
 		}
 	}
 
 	const nameComponent = text || <i>{unnamedText}</i>;
 	const contents = (
-		<span>
+		<div className="react-select__value-container" style={{cursor: 'pointer', fontSize: '15px', padding: '8px'}} onClick={parentEventHandler}>
```
MonkeyDo: The className made me cringe. That's not a right way to go about it.
Thankfully, if you look at LinkedEntityProps, classname is passed down, so you can add it to the arguments along with `{onSelect, option}` line 41,  and then pass it in this div. Don't forget to remove the `style` as well.

src/client/entity-editor/common/entity-linked-search-field.js
```diff
@@ -101,6 +93,30 @@ function EntityLinkedSearchField(
...
+	 * @returns {Object} the formatted data
+	 * @param {Array<LanguageOption>} languageOptions - The list of possible languages for an
+	 * entity.
+	 */
+
+	function entityToOption(entity) {
+		const id = isArea(entity) ? entity.id : entity.bbid;
+		const languageId = _.get(entity, ['defaultAlias', 'languageId']);
+		const languageIndex = languageOptions.find(index => index.value === languageId);
+		const language = languageOptions[languageIndex.value].label;
```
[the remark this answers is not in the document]
MonkeyDo: And since that returns the object, you'll want to make sure you pass the label rather than the whole object (you did that correctly there, just a reminder to do so again)

src/client/entity-editor/common/entity.js
```diff
@@ -47,6 +48,7 @@ function Entity(
 				disambiguation &&
 				<span className="disambig"><small>({disambiguation})</small></span>
 			}
+			<span style={{float: 'right'}}>{language}</span>
```
MonkeyDo: Maybe worth adding a `className="text-muted small"` there, just to make it a bit more distinct.

src/client/entity-editor/common/entity-linked-search-field.js
```diff
@@ -0,0 +1,154 @@
...
+ * @param {boolean} props.empty - Passed to the ValidationLabel within the
+ *        component to indicate that the field is empty.
+ * @param {Array<LanguageOption>} props.languageOptions - The list of possible languages for an
+ *        entity.
+ * @param {string | Array<string>} props.type - Determines the types of entity
+ *        to search for.
+ * @param {string} props.label - The text to be used for the input label.
+ * @returns {Object} A React component containing the rendered input.
+ */
+function EntityLinkedSearchField(
```
MonkeyDo: I would suggest naming this component to reflect its more general purpose, maybe something like EntitySearchFieldOption ?

src/client/entity-editor/common/entity-linked-search-field.js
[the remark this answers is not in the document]
pv-912: Yes, I consoled the output but didn't think to get the value directly form here. Sure, I will correct it.

src/client/entity-editor/common/entity.js
pv-912: Okay.

src/client/entity-editor/common/linked-entity.js
pv-912: Okay

pv-912:
@MonkeyDo I have made the requested changes as told but getting an error that the options are automatically disabled when searching. Can u please look into this.

src/client/entity-editor/common/entity-search-field-option.js
```diff
@@ -0,0 +1,128 @@
...
+	 *
+	 * @param {Object} entity the entity to convert
+	 * @returns {Object} the formatted data
+	 * @param {Array<LanguageOption>} languageOptions - The list of possible languages for an
+	 * entity.
+	 */
+	entityToOption(entity) {
+		const id = this.isArea(entity) ? entity.id : entity.bbid;
+		const languageId = _.get(entity, ['defaultAlias', 'languageId']);
+		const language = this.props.languageOptions.find(index => index.value === languageId).label;
```
MonkeyDo: This will throw an error if `.find` does not return anything.
It's unlikely, but it's worth making sure it doesn't happen. For example, below in the return value you can return `language: language && language.label`.

src/client/entity-editor/common/entity.js
```diff
@@ -48,6 +48,7 @@ function Entity(
 				disambiguation &&
 				<span className="disambig"><small>({disambiguation})</small></span>
 			}
+			<span className="text-muted small">{language}</span>
```
MonkeyDo: I am not convinced you should add this here. I understand the reason, and it's a good reason, but the Entity component is not used only in this select but also in other components.

I think for now seeing the extra details only in the select options is sufficient.

src/client/entity-editor/common/linked-entity.js
```diff
@@ -0,0 +1,91 @@
...
+		onSelect(option, event);
+	}
+	function childEventHandler(event) {
+		event.stopPropagation();
+		event.preventDefault();
+		if (url) {
+			window.open(url, '_blank');
+		}
+	}
+	// className="react-select__value-container" style={{cursor: 'pointer', fontSize: '15px', padding: '8px'}}
```
MonkeyDo: Don't forget to clean up after yourself please

src/client/entity-editor/common/linked-entity.js
```diff
@@ -0,0 +1,91 @@
...
+			&nbsp;&nbsp;
+			{
+				disambiguation &&
+				<span className="disambig"><small>({disambiguation})</small></span>
+			}
+			{' '}
+			<a onClick={childEventHandler}>
+				<FontAwesome name="external-link-alt"/>
+			</a>
+			<span className="text-muted small">{language}</span>
```
MonkeyDo: the `float:right` style was good here, can you please add it back?

src/client/entity-editor/common/linked-entity.js
```diff
@@ -0,0 +1,91 @@
...
+		</div>
+	);
+
+	return contents;
+}
+
+// LinkedEntity.displayName = 'LinkedEntity';
```
MonkeyDo: Why is this commented out?

src/client/entity-editor/common/entity-search-field-option.js
pv-912: okay.

pv-912:
> I still have the error below in the browser console. Did you try running the code?
> 
> ```
> Warning: Stateless function components cannot be given refs. Attempts to access this ref will fail.
> Check the render method of `Select`.
>     in LinkedEntity (created by Select)
> ```
> That is most likely the reason why selecting an option is not working.
> I imagine that the `LinkedEntity` component also needs to be turned into a class (from reading that error message), but I haven't tried it

src/client/entity-editor/common/linked-entity.js
pv-912: Sure.

pv-912:
@MonkeyDo I am not getting any error in console.

src/client/entity-editor/common/entity.js
MonkeyDo: What I mean is the language should not be added in this component.

pv-912:
@MonkeyDo  I have done the requested changes except [this](https://github.com/bookbrainz/bookbrainz-site/pull/261#discussion_r266384496) as it was not clear to me. Also, I am still getting the same error that option is not getting selected on clicking on it. 
Are u still getting that error in your in the console?

src/client/entity-editor/common/entity.js
MonkeyDo: No, I would just not add the language.

MonkeyDo:
> @MonkeyDo I have done the requested changes except [this](https://github.com/bookbrainz/bookbrainz-site/pull/261#discussion_r266384496) as it was not clear to me. Also, I am still getting the same error that option is not getting selected on clicking on it.
> Are u still getting that error in your in the console?

The error is gone after the last commit, that's good.

After some debugging, I also found out why selecting doesn't seem to work: there is no `onChange` function passed to the select component. Such a function is passed as a prop to `EntitySearchFieldOption`, so what you need to do is add `{...this.props}` to `ImmutableAsyncSelect`. That way you also pass along the properties passed by the parent, including `onChange`.

It is also annoying that when you click on the link icon to open in a new tab, the select is closed and the input emptied.
Try adding `onBlurResetsInput={false}`. You might need to play with other properties as well, you can find them all here: https://github.com/JedWatson/react-select/tree/v1.x#select-props

src/client/entity-editor/common/entity-search-field-option.js
```diff
@@ -0,0 +1,128 @@
...
+				options: response.body.map(this.entityToOption)
+			}));
+	}
+
+	render() {
+		const labelElement = <ValidationLabel empty={this.props.empty} error={this.props.error}>{this.props.label}</ValidationLabel>;
+		return (
+			<CustomInput label={labelElement} tooltipText={this.props.tooltipText}>
+				<ImmutableAsyncSelect
+					className="react-select__option"
```
MonkeyDo: Please remove this line. It's hacky and not needed.

pv-912: Sure.

pv-912:
@MonkeyDo  done the requested changes.

pv-912:
@MonkeyDo  I have replaced the `EntitySearchField`  with `EntitySearchFieldOption` in other components also. Also removed the unused file containing the entity-search-field component.

src/client/entity-editor/common/entity-search-field-option.js
```diff
@@ -0,0 +1,129 @@
...
+}
+
+EntitySearchFieldOption.displayName = 'EntitySearchFieldOption';
+EntitySearchFieldOption.propTypes = {
+	empty: PropTypes.bool,
+	error: PropTypes.bool,
+	label: PropTypes.string.isRequired,
+	languageOptions: PropTypes.array.isRequired,
+	tooltipText: PropTypes.string,
+	type: PropTypes.array.isRequired
```
MonkeyDo: The `type` prop can either be a string or an array (for example in `client/entity-editor/relationship-editor/relationship-editor.js`), so the correct definition would be
```
type: PropTypes.oneOfType([
		PropTypes.string,
		PropTypes.arrayOf(PropTypes.string)
]).isRequired
```

src/client/entity-editor/common/entity-search-field-option.js
```diff
@@ -0,0 +1,129 @@
...
+		);
+	}
+}
+
+EntitySearchFieldOption.displayName = 'EntitySearchFieldOption';
+EntitySearchFieldOption.propTypes = {
+	empty: PropTypes.bool,
+	error: PropTypes.bool,
+	label: PropTypes.string.isRequired,
+	languageOptions: PropTypes.array.isRequired,
```
MonkeyDo: For an Area search dropdown (for example in `client/entity-editor/creator-section/creator-section.js`) the `languageOptions` property is not needed. You will need to change to `languageOptions: PropTypes.array` and add a default empty array in `EntitySearchFieldOption .defaultProps` underneath.

pv-912:
@MonkeyDo done the requested changes. Have a look.

PR #267, opened 2019-03-18
Fix #BB-302: Replace date field and date validation mechanism

akhilesh26:


src/client/entity-editor/validators/base.js
```diff
@@ -99,6 +99,9 @@ export function validateBoolean(
 export function validateDate(
 	value: mixed, required: boolean = false
 ): boolean {
+	if (!value) {
```
MonkeyDo: Looking at the code, maybe this block should be after the call to `absentAndRequired`?

MonkeyDo:
Also note that currently the tests are failing: https://travis-ci.org/bookbrainz/bookbrainz-site/builds/508330194?utm_source=github_status&utm_medium=notification

akhilesh26:
@MonkeyDo Please check once again. Please focus on validation, error and warn messages and UI.
Please suggest more improvements for a better experience. Once this will be finalized and your rename PR will merge, I will create PR by replacing this date field component everywhere.

src/client/entity-editor/validators/creator.js
```diff
@@ -49,8 +50,16 @@ export function validateCreatorSectionEndArea(value: any): boolean {
...
-): boolean {
-	return validateDate(endValue) && dateIsBefore(beginValue, endValue);
+): object {
+	const {isValid, errorMessage} = validateDate(endValue);
+
+	if (isValid) {
+		if (dateIsBefore(beginValue, endValue)) {
+			return {errorMessage: '', isValid: true};
+		}
+		return {errorMessage: 'Date of death must be greated than Date of birth', isValid: false};
```
MonkeyDo: It would be best here to use the `labelsForCreator` helper function, which means you'll have to pass a new `isGroup` argument to `validateCreatorSectionEndDate`.
See how it is used here:
https://github.com/bookbrainz/bookbrainz-site/blob/9e843b78a86e93a27363ff6db1125e29a69fb348/src/client/components/pages/entities/creator.js#L45-L51

akhilesh26:
@MonkeyDo, I modified some date validation test in the last commit, please have a look, so that I can modify all tests accordingly.
Thanks!

MonkeyDo:
@akhilesh26
So far what you have changed in the tests looks good.
There are still a few failing tests for Authors. I would suggest having a look at fixing those that use the `validateDate` base validator as they apply for all entities.
I think `validateDate` breaks for null or undefined values, and should be made more robust.


Once all the tests pass for Authors we can have another round of review.

Here's my output for the failing Author tests:
```
1) validateAuthorSection* functions
     validateAuthorSectionBeginDate
       should pass a null value:
   TypeError: Cannot destructure property `year` of 'undefined' or 'null'.
    at validateDate (src/client/entity-editor/validators/base.js:98:29)
    at validateAuthorSectionBeginDate (src/client/entity-editor/validators/author.js:43:34)
    at Context.it (test/src/client/entity-editor/validators/helpers.js:130:18)

2) validateAuthorSection* functions
     validateAuthorSectionEndDate
       should reject if the begin date occurs after the end one:
   AssertionError: expected { errorMessage: '', isValid: true } to be false
    at Context.it (test/src/client/entity-editor/validators/helpers.js:153:4)

3) validateAuthorSection* functions
     validateAuthorSectionEndDate
       should pass if the begin date is empty/undefined/invalid:
   TypeError: Cannot read property 'day' of null
    at isNullValue (src/client/entity-editor/validators/base.js:104:15)
    at dateIsBefore (src/client/entity-editor/validators/base.js:108:6)
    at validateAuthorSectionEndDate (src/client/entity-editor/validators/author.js:61:7)
    at _data.INVALID_BEGIN_DATE_PAIR.reduce (test/src/client/entity-editor/validators/helpers.js:160:12)
    at Array.reduce (<anonymous>)
    at Context.it (test/src/client/entity-editor/validators/helpers.js:159:43)

4) validateAuthorSection* functions
     validateAuthorSectionEndDate
       should reject if the end date is invalid:
   AssertionError: expected { errorMessage: '', isValid: true } to be false
    at Context.it (test/src/client/entity-editor/validators/helpers.js:171:4)

5) validateAuthorSection* functions
     validateAuthorSection
       should reject an Object with an invalid begin date:
   AssertionError: expected undefined to be false
    at Context.it (test/src/client/entity-editor/validators/test-author.js:123:3)

6) validateAuthorSection* functions
     validateAuthorSection
       should reject an Object with an invalid end date:
   AssertionError: expected undefined to be false
    at Context.it (test/src/client/entity-editor/validators/test-author.js:139:3)

7) validateAuthorSection* functions
     validateAuthorSection
       should pass any other non-null data type:
   TypeError: Cannot destructure property `year` of 'undefined' or 'null'.
    at validateDate (src/client/entity-editor/validators/base.js:98:29)
    at validateAuthorSectionBeginDate (src/client/entity-editor/validators/author.js:43:34)
    at validateAuthorSection (src/client/entity-editor/validators/author.js:84:3)
    at Context.it (test/src/client/entity-editor/validators/test-author.js:174:18)

8) validateAuthorSection* functions
     validateAuthorSection
       should pass a null value:
   TypeError: Cannot destructure property `year` of 'undefined' or 'null'.
    at validateDate (src/client/entity-editor/validators/base.js:98:29)
    at validateAuthorSectionBeginDate (src/client/entity-editor/validators/author.js:43:34)
    at validateAuthorSection (src/client/entity-editor/validators/author.js:84:3)
    at Context.it (test/src/client/entity-editor/validators/test-author.js:179:18)

9) validateAuthorSection* functions
     validateForm
       should reject any other non-null data type:
   TypeError: Cannot destructure property `year` of 'undefined' or 'null'.
    at validateDate (src/client/entity-editor/validators/base.js:98:29)
    at validateAuthorSectionBeginDate (src/client/entity-editor/validators/author.js:43:34)
    at validateAuthorSection (src/client/entity-editor/validators/author.js:84:3)
    at validateForm (src/client/entity-editor/validators/author.js:104:3)
    at Context.it (test/src/client/entity-editor/validators/test-author.js:276:18)

10) validateAuthorSection* functions
     validateForm
       should reject a null value:
   TypeError: Cannot destructure property `year` of 'undefined' or 'null'.
    at validateDate (src/client/entity-editor/validators/base.js:98:29)
    at validateAuthorSectionBeginDate (src/client/entity-editor/validators/author.js:43:34)
    at validateAuthorSection (src/client/entity-editor/validators/author.js:84:3)
    at validateForm (src/client/entity-editor/validators/author.js:104:3)
    at Context.it (test/src/client/entity-editor/validators/test-author.js:281:18)
`
``

**Comment by MonkeyDo** [comment]
I just spotted something: if you edit an author that already has a date entered, if you click on the date picker the default date when it open is today's date.
It would be great if it opened on the existing date if there's one.

**Comment by akhilesh26** [comment]
> I just spotted something: if you edit an author that already has a date entered, if you click on the date picker the default date when it open is today's date.
> It would be great if it opened on the existing date if there's one.

The date picker will not show partial date correctly in the calendar.

**Comment by MonkeyDo** [comment]
> The date picker will not show partial date correctly in the calendar.


This is for a full date though:
Capture d’écran 2019-05-17 à 17 32 58

**Comment by akhilesh26** [comment]
> > The date picker will not show partial date correctly in the calendar.
> 
> This is for a full date though:
> Capture d’écran 2019-05-17 à 17 32 58

I will do that, but It will work correctly for the full dates. Would It be ok?

**Comment by akhilesh26** [comment]
@MonkeyDo Please check all functionality and test cases once again. 
To open existing date in date picker, it requires to pass the Unix timestamp of existing date. I am trying to convert the existing date to a unix timestamp.

File: test/src/client/entity-editor/validators/helpers.js [hunk 3]
```diff
@@ -150,7 +150,7 @@ export function testValidateEndDateFunc(
 			const result = INVALID_DATE_PAIR.reduce((res, datePair) =>
 				res || endDateValidationfunc(datePair.first, datePair.second),
 			false);
-			expect(result).to.be.false;
+			expect(result.isValid).to.be.false;
```
MonkeyDo: Here and everywhere you use `.reduce` on a valid/invalid date array, shouldn't `.isValid` be checked in the reduce callback rather than on the result?
```
const result = INVALID_DATE_PAIR.reduce((res, datePair) =>
	res || endDateValidationfunc(datePair.first, datePair.second).isValid,
false);
expect(result).to.be.false;
```

src/client/entity-editor/validators/base.js
```diff
@@ -96,41 +94,47 @@ export function validateBoolean(
...
-	value: mixed, required: boolean = false
-): boolean {
-	if (absentAndRequired(value, required)) {
-		return false;
-	}
-
-	if (!nilOrString(value)) {
-		return false;
-	}
+export function validateDate(value: mixed) {
```
MonkeyDo: Here is a way to get the validateDate and isNullValue  to accept all kinds of values, including nulls and empty objects:

```
export function validateDate(value: mixed) {
	const year = _.get(value, 'year');
	const month = _.get(value, 'month');
	const day = _.get(value, 'day');
	const {isValid, errorMessage} = dateValidator(day, month, year);
	return {errorMessage, isValid};
}

export function isNullValue(date) {
	return !_.get(date, 'day') && !_.get(date, 'month') && !_.get(date, 'year');
}
```

MonkeyDo:
Great, I think that's all the feedback I had for the tests, well done!

The last piece of the puzzle is opening the date picker to the correct date, if there is one.
Regarding partial dates (missing the day or day and month), I think you should still try to pass it to the Date constructor:
`new Date("1993-09")` is valid and returns the date `1993-08-31T22:00:00.000Z`, and `new Date("1993")` = `1993-01-01T00:00:00.000Z`.
I think this is an acceptable behavior for the datepicker.

MonkeyDo:
Great work @akhilesh26!
The date fields looks good, the date picker is working well with it, the date validation has been improved, and we can now enter partial dates again.
Thanks for the good work !
Ready to merge.

akhilesh26:
Thanks, @MonkeyDo, to review the PR. But don't merge it now. I will push one or two commits for cleaning the code.

MonkeyDo:
After some more testing, I noticed an error when using incomplete dates:
Capture d’écran 2019-05-27 à 18 54 51

This should be a correct date pair, shouldn't it?

The dissolved date becomes valid if I add a day to either date field.

src/client/entity-editor/common/new-date-field.js
```diff
@@ -0,0 +1,154 @@
...
+							type="text"
+							value={this.state.day}
+							onChange={this.handleDayChange}
+						/>
+					<InputGroup.Button style={{ fontSize: "inherit"}}>
+							<DatePicker
+								peekNextMonth
+								showMonthDropdown
+								showYearDropdown
+								selected={momentDate.isValid() ? momentDate : null}
```
MonkeyDo: You can replace moment entirely here and use a small utility function to detect if a date is valid:
```
function isValidDate(d) {
  return d instanceof Date && !isNaN(d);
}
```

(Not mine, found it on [StackOverflow](https://stackoverflow.com/a/1353711))

akhilesh26:
> After some more testing, I noticed an error when using incomplete dates:
> Capture d’écran 2019-05-27 à 18 54 51
> 
> This should be a correct date pair, shouldn't it?
> 
> The dissolved date becomes valid if I add a day to either date field.

Fixed: 
Screenshot

src/client/entity-editor/validators/creator.js
MonkeyDo: This is still to do, should be a quick fix.
Capture d’écran 2019-05-27 à 19 51 07

akhilesh26: I tried to fix this, it is not easy because of tests reusability, Please suggest generic error msg which will work for both type of author person and group.

MonkeyDo: I took the liberty of pushing the code changes for this. I didn't run into any issue with the tests; what issue did you encounter?

akhilesh26: `validateCreatorSectionEndDate`  function is also called from [testValidateEndDateFunc](https://github.com/bookbrainz/bookbrainz-site/blob/2f081a0de9128a3cb3d601c88a70a8ab69f655bf/test/src/client/entity-editor/validators/helpers.js#L145) as `endDateValidationfunc`. `testValidateEndDateFunc`  function is common for author and publisher also. If intruduce a new argumenet `isGroup` in  `validateCreatorSectionEndDate` then how will we call `endDateValidationfunc` for both author and publisher?

MonkeyDo: The code I pushed  does not require `authorType` to be defined in [validateAuthorSectionEndDate](https://github.com/bookbrainz/bookbrainz-site/pull/267/commits/8f7aed3a8cbf47bf9c030305dc1523d443ad28bf#diff-ad68a18cccd05bafe91d57126c7f7ce3R59), and falls back graciously, so there's no need to pass it in testValidateEndDateFunc.
We don't test for the exact message, so it does not make any difference for those tests.

akhilesh26: Thanks for the help.

MonkeyDo:
@akhilesh26 I pushed a couple of modifications/cleanup.
If you don't see anything else to clean up or change, I think this should be ready to merge !

akhilesh26:
Thanks a lot  @MonkeyDo, 
I would like to test overall work once, then I will let you know to merge.

PR #268, opened 2019-03-21
fix(server): fix issue #BB-336 by checking publicationBbid is available or not

akhilesh26:
### Problem

The edition was added when Publication was not compulsory in Edition form. 

### Solution

Check that publicationBbid is available or not.

### Areas of Impact

`src/server/routes/entity/edition.js`

### DIFF
--- a/src/server/routes/entity/edition.js
+++ b/src/server/routes/entity/edition.js
@@ -256,7 +256,7 @@ function editionToFormState(edition) {
 			null : entityToOption(edition.publisherSet.publishers[0])
 	);
 
-	const publication = entityToOption(edition.publication);
+	const publication = !edition.publicationBbid ? null : entityToOption(edition.publication);
 
 	const editionSection = {
 		depth: edition.depth,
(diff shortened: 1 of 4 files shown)


src/server/routes/entity/edition.js
```diff
@@ -256,7 +256,7 @@ function editionToFormState(edition) {
 			null : entityToOption(edition.publisherSet.publishers[0])
 	);
 
-	const publication = entityToOption(edition.publication);
+	const publication = !edition.publicationBbid ? null : entityToOption(edition.publication);
```
MonkeyDo: I'd rather see a fix to `return null` if there is no entity passed to `entityToOption`, that way we protect against other misuses.

There are other `entityToOption` functions in other components where the same fix would be useful, if you search for them.

MonkeyDo: > that way we protect against other misuses

What I mean by that is that if `entityToOption` is called with a null or undefined or empty object somewhere else in the file, now or in future code, we won't run into the same problem of remembering to have to check if it exists.

Doing the same in other components that have `entityToOption` is the same idea: defend against non-existing entity rather than risk throwing an error.

akhilesh26:
@MonkeyDo thanks for giving such detailed review.
I created change according to your suggestion.

akhilesh26:
@MonkeyDo Changes did!

PR #271, opened 2019-04-01
Fix BB-323: Rename Creator and Publication entities

MonkeyDo:
### Problem
We are renaming 
- Creator -> Author
  - Creator Credit -> Author Credit
- Publication -> Edition Group

I previously superficially renamed the entites (on the UI for the most part) but in some instances the entity names are derived from the ORM models, or refer to certain sql tables.

### Solution
This PR, along with now merged [PR 152 on bookbrainz-data-js](https://github.com/bookbrainz/bookbrainz-data-js/pull/152), renames the entities everywhere:
- SQL schema
  - SQL migration script
- routes
- file names
- object properties
- tests
- etc.

That should be all of it for renaming.

Previous routes (`/creator` and `/publication`) have been preserved to ensure retro-compatibility. For example MusicBrainz can point to `/creator`s.

### Areas of Impact
Pretty much all of it.

### Setup
You will need to run `npm install` to make sure you have the new version of `bookbrainz-data-js`. 
Then you'll need to run sql migration scripts to update the database and schema: 
In a terminal window, from the root repository
1. psql -U postgres -h localhost -d bookbrainz -f sql/migrations/import/up.sql
2. psql -U postgres -h localhost -d bookbrainz -f sql/migrations/2019-04/up.sql
3. psql -U postgres -h localhost -d bookbrainz -f sql/scripts/create_triggers.sql

To revert the database and schema changes, run:
1. psql -U postgres -h localhost -d bookbrainz -f sql/migrations/2019-04/down.sql
2. psql -U postgres -h localhost -d bookbrainz -f sql/migrations/import/down.sql
3. Then check out the branch you want to be on and run
4. psql -U postgres -h localhost -d bookbrainz -f sql/scripts/create_triggers.sql


src/server/helpers/entityRouteUtils.js
```diff
@@ -196,7 +196,7 @@ export function makeEntityCreateOrEditHandler(
  * when one entity created from other.
  * @param {object} props - props related to new entity
  * @param {number} relationshipTypeId - relationshipId number for initaial relationship
- * @param {object} targetEntity - details about target entitiy like publication, publisher and creator
+ * @param {object} targetEntity - details about target entitiy like editionGroup, publisher and creator
```
akhilesh26: Replace creator -> author.

src/server/routes/entity/edition.js
```diff
@@ -122,14 +122,14 @@ router.get(
...
-		const {Publication, Publisher, Work} = req.app.locals.orm;
+		const {EditionGroup, Publisher, Work} = req.app.locals.orm;
 		const propsPromise = generateEntityProps(
 			'edition', req, res, {}
 		);
 
-		if (req.query.publication) {
-			propsPromise.publication =
-				Publication.forge({bbid: req.query.publication})
+		if (req.query['edition-group']) {
```
akhilesh26: Is any specific reason to use square Bracket instead of the dot notation to retrieve `edition-group` from `req.query`? I think It is creating some inconsistency between dot notation and square bracket. If it is necessary, please mention the reason in the inline comment at that place.

MonkeyDo: `req.query.edition-group` is not valid JS because of the dash between edition and group.
I'll add a comment.

PR #272, opened 2019-04-04
Use metabrainz node image which has runit and consul

paramsingh:
### Problem

The docker image didn't have consul and other stuff such as runit needed to run this on hetzner.

### Solution

Create a new docker image for node (https://github.com/metabrainz/docker-node/blob/master/10/Dockerfile) and use it instead.

### Areas of Impact

Ideally, no behaviour should change, I haven't tested it very thoroughly though.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,10 @@
-FROM node:10
+FROM metabrainz/node:10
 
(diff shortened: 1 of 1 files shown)


paramsingh:
Included in #273

PR #273, opened 2019-04-04
Production and development Docker setup for Hetzner

paramsingh:


package.json
```diff
@@ -14,6 +14,7 @@
     "WIP-build-client-webpack": "cross-env NODE_ENV=production SSR=true webpack --progress --config webpack.client.js",
     "prestart": "npm run build-less & npm run build-client-js & npm run build-server-js",
     "start": "cross-env SSR=true node ./lib/server/app.js",
+    "start-prod": "npm run build-client-js & npm run build-server-js & cross-env SSR=true node ./lib/server/app.js",
```
MonkeyDo: Why this instead of `start`?
Keep in mind that `prestart` just above will be run first any time you call `start`.

I see you're running less at the end of the Dockerfile. What's the reason for doing it there instead of in npm scripts (build-less, run by default when running `npm start`) ?

Dockerfile
```diff
@@ -1,7 +1,12 @@
...
+ARG DEPLOY_ENV
 
 ARG BUILD_DEPS=" \
-    build-essential"
+    build-essential \
+    git \
+    python-dev \
+    libpq-dev"
 
 ARG RUN_DEPS=" \
```
MonkeyDo: This can be removed.
In the `RUN` command below, replace `curl -sL https://deb.nodesource.com/setup_10.x | bash - && \` with `apt-get update && \`  and remove the reference to `$RUN_DEPS`

At the moment, I think it may be reinstalling node :p

PR #274, opened 2019-04-05
Setup cron and regular data dumps for hetzner

paramsingh:
### Problem

The cron and data dumps setup hadn't been migrated.

### Solution

* Add scripts to create dumps and copy to MusicBrainz FTP (ftp://ftp.musicbrainz.org/pub/musicbrainz/bookbrainz)
* Setup dockerfile with changes for the new cron service

### Areas of Impact

Not much really, but the bookbrainz.org/dumps/latest.bz2 link will need to redirect to ftp now.

### DIFF
--- a/docker/prod/rc.local
+++ b/docker/prod/rc.local
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+# BookBrainz rc.local
+# Enables a specific runit service depending on what the
+# CONTAINER_NAME variable is set to
+
+
+if [ "${CONTAINER_NAME}" = "bookbrainz-web-prod" ]
+then
+    rm -f /etc/service/webserver/down
+    exit 0
+fi
+
+if [ "${CONTAINER_NAME}" = "bookbrainz-cron-prod" ]
+then
+    rm -f /etc/service/cron/down
+    exit 0
+fi
+
+echo "init script has no clue which service to start. Set env var CONTAINER_NAME"
+exit 1
--- a/scripts/create-dumps.sh
+++ b/scripts/create-dumps.sh
@@ -48,3 +48,5 @@ chown bookbrainz:bookbrainz ./*
 chmod 644 ./*
 
 popd
+
+exit 0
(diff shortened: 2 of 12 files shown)


paramsingh:
I have this running on prince rn, so a dump should appear by tomorrow.

scripts/create-dumps.sh
```diff
@@ -1,41 +1,46 @@
...
 # Switch directory
-pushd /home/bookbrainz/bookbrainz.org/bookbrainz-docker-prod/dumps
+pushd /home/bookbrainz/data/dumps
+
 
 DUMP_FILE=bookbrainz-dump-`date -I`.sql
 
-export PGPASSWORD="postgres"
+echo "Creating data dump..."
+export PGPASSWORD="bookbrainz"
```
MonkeyDo: I was hoping we'd get rid of a hardcoded password entirely…
Should this be set as an env variable instead?

docker/prod/rc.local
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
mayhem: Isn't a beta version of this file missing?

docker/prod/rc.local
```diff
@@ -0,0 +1,21 @@
...
+
+# BookBrainz rc.local
+# Enables a specific runit service depending on what the
+# CONTAINER_NAME variable is set to
+
+
+if [ "${CONTAINER_NAME}" = "bookbrainz-web-prod" ]
+then
+    rm -f /etc/service/webserver/down
```
mayhem: This still looks wrong to me at first glance, but is really quite cool.

scripts/create-dumps.sh
mayhem: Our other databases do not have passwords -- we should probably work to get rid of this one too.

scripts/create-dumps.sh
```diff
@@ -1,41 +1,46 @@
 #!/usr/bin/env bash
```
mayhem: Let's start following https://github.com/metabrainz/metabrainz-howto-guides/blob/master/writing_effective_data_processing_scripts.md and have our data scripts return proper exit codes.

paramsingh:
Ready for review again.

scripts/create-dumps.sh
MonkeyDo: @paramsingh 
Rob says: 
> a last line of `exit 0` should suffice.

paramsingh: Done!

PR #275, opened 2019-04-05
test.bookbrainz.org on hetzner

paramsingh:
### Problem

Need to migrate test.BB as well.

### Solution

Not much, make the database name a configurable value.

### Open Questions

Right now I've made new.BB and test-new.BB both connect to the same elasticsearch and redis instance. Is this gonna be a problem, or will the indexes/rediskeys have different names / be independent?

See this in action at https://test-new.bookbrainz.org, I think we can make test.bookbrainz.org point to gateways now and shut down test.BB in the VM.

### DIFF
--- a/config/config.json.ctmpl
+++ b/config/config.json.ctmpl
@@ -14,20 +14,19 @@
 		"callbackURL": "{{template "KEY" "musicbrainz/callback_url"}}"
 	},
 	"session": {
-		"maxAge": 2592000000,
-		"secret": "{{template "KEY" "secret_key"}}",
-		"secure": false,
         {{if service "bookbrainz-redis"}}
         {{with index (service "bookbrainz-redis") 0}}
 		"redis": {
 			"host": "{{.Address}}",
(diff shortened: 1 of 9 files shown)


MonkeyDo:
Changes I've committed:

- Gracefully shut down application (close connections, relinquish bound port [more advanced cleanup in the future?])
- Force-kill application after configurable amount of time
- Shorten the amount of time before consul-template force-kills the application
- Build the less and js files in the Dockerfile so as to ship a fully built image

config/config.json.ctmpl
```diff
@@ -5,7 +5,8 @@
 {
 	"site": {
 		"proxyTrust": true,
-		"log": "error"
+		"log": "error",
+		"forceExitAfterMs": 5000
```
mayhem: A ha!

MonkeyDo: Two prong approach: 1. ask nicely to terminate connections and relinquish port and 2. THE HAMMER

PR #278, opened 2019-04-25
fix(Dockerfile): Add missing bzcat

shivam-tripathi:
Populating database via `database-init-docker.sh` fails because a requirement bzcat is not installed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,7 @@ ENV PG_MAJOR 9.5
 RUN echo 'deb http://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' $PG_MAJOR > /etc/apt/sources.list.d/pgdg.list
 RUN apt-get update \
     && apt-get install -y --no-install-recommends postgresql-client-$PG_MAJOR \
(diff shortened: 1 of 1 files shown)


shivam-tripathi:
Also, running `bzcat $DUMP_FILE | psql -h postgres -U postgres -d bookbrainz` doesn't create some relations and views causing local docker run of the site to fail. 
Some of the missing pieces are: `bookbrainz.author`, `bookbrainz.author_header` etc.

MonkeyDo:
> Also, running `bzcat $DUMP_FILE | psql -h postgres -U postgres -d bookbrainz` doesn't create some relations and views causing local docker run of the site to fail.
> Some of the missing pieces are: `bookbrainz.author`, `bookbrainz.author_header` etc.

Right. I hadn't thought about that. After merging PR #271 the schema is different, and the online dumps are using the old schema. I don't have a good solution for this, short of waiting until the prod DB is migrated to the new schema.
You can try checking out the `stable` branch that does not have the PR merged

shivam-tripathi:
Would be good to temporarily update the README file

PR #280, opened 2019-05-02
Improve displaying of deleted entities

MonkeyDo:
### Problem
Deleted entities appear as '(unnamed)' both on the home page and on the entity display page.
The entity display page does not show clearly that an entity has been deleted, and allows to edit and delete a deleted entity, which does not make sense.

### Solution
Implemented a getEntityParentAlias function in the ORM and used that in the display page.
Modified the index page recent revisions query to include parent revision and its default alias.

Clearly show on an entity's display page that it has been deleted.


MonkeyDo:
screencapture-localhost-9099-author-627cd764-81ed-4bb1-8876-d99fcd480b54-2019-05-06-18_55_38
Capture d’écran 2019-05-06 à 18 56 46

PR #282, opened 2019-05-07
Update dockerfile

MonkeyDo:
### Problem
When I build the image for test environment (`push.sh test test`) the build fails with this error:
`COPY failed: stat /var/lib/docker/tmp/docker-builder424901826/docker/test/cron.service: no such file or directory``
Indeed, there's no `cron.service` file in `docker/test`.

### Solution
I'm adding a `test` build target (without the cron-related setup) and modifying  push.sh to use that.

Also some small cleanup/modifications in the dockerfile


MonkeyDo:
Currently, building the test image works, but the service in the container fails to start with error message:
`runsv webserver: fatal: unable to start ./run: file does not exist`

MonkeyDo:
All right, I fixed the issue I was having, and cleaned up after myself.
Ready for review.

@paramsingh I do have one thing I'd like to check with you:
I've added the webserver config file (`config/config.json`) in `.dockerignore` to make sure I don't copy over my local config and leak tokens.
That means I have to mount the config file in docker-compose. It's working fine. 
The question is: is there a better way you can think of to achieve the same result (copy local config in local dev, don't copy over config in prod builds)?

docker-compose.yml
```diff
@@ -16,6 +16,8 @@ services:
       - elasticsearch
       - redis
       - postgres
+    volumes:
+      - "./config/config.json:/home/bookbrainz/bookbrainz-site/config/config.json:ro"
```
paramsingh: I see this going away when the entire directory is mounted with automatic reloading of sites. So looks good to me for now.

docker/push.sh
```diff
@@ -17,12 +17,18 @@ git describe --tags --dirty --always > .git-version
...
 docker build -t metabrainz/bookbrainz:$TAG \
         --target bookbrainz-prod \
         --build-arg GIT_COMMIT_SHA=$(git rev-parse HEAD) \
         --build-arg DEPLOY_ENV=$ENV .
-echo "Done!"
+RESULT=$?
+if [ $RESULT -eq 0 ]; then
+  echo "Done!"
+else
+  echo "Docker build command failed with error code $RESULT, exiting."
```
paramsingh: nice!

PR #283, opened 2019-05-09
Hide deprecated relationships in relationship editor

MonkeyDo:
Like it says on the tin !

A cleanup SQL script is needed to convert those Edition-Edition Group relationships to the Edition Group property of an Edition


MonkeyDo:
Hi @LordSputnik !
I'd love your opinion and a second pair of eyes on the sql script.
The idea is to replace all the deprecated Edition <-> Edition Group relationships with edition.edition_group, and delete the deprecated relationships from the entities' current relationship sets.

What do you think about this approach? Is that what you had in mind?

PR #288, opened 2019-05-28
feat: add initial api lookup endpoint for entities with test

akhilesh26:
### BB API 

Run this initial setup for BB API.  
Install dependencies using `npm install`  
To check the endpoint, run the project without docker with `npm run start-api`.
Endpoints implemented:  
`http://localhost:9098/work/<bbid>`  
`http://localhost:9098/work/<bbid>/aliases`  
`http://localhost:9098/work/<bbid>/identifiers`  

To run the test for this endpoint use `npm test ./test/src/api/routes/test-work.js`


akhilesh26:
The test will not pass because data is not pushed into bookbeainz_test. To pass the test, replace 
 database `bookbrainz` in place of  `bookbraiz_test`  in `test.json` file.

test/src/api/endpoints/test-work.js
```diff
@@ -0,0 +1,129 @@
+/* eslint-disable */
+
+process.env.NODE_ENV = 'test';
```
MonkeyDo: I don't think you should need this. I don't think it's a good idea to leave it there either.
It should be set when using `npm test`

test/src/api/endpoints/test-work.js
```diff
@@ -0,0 +1,129 @@
...
+	id: 1,
+	name: 'work name',
+	sortName: 'Work sort name'
+};
+
+const revisionAttribs = {
+	authorId: 1,
+	id: 1
+};
+const aBBID = '13ab71b4-908f-4c72-88a9-4b94a56c7d3e';
```
MonkeyDo: You should make this a random uuid. I can see a couple of potential issues otherwise that could be avoided.
I suggest following what is done in bookbrainz-data-js and installing the 'faker' package, then using that for bbids, names, etc.
https://github.com/bookbrainz/bookbrainz-data-js/search?q=faker&unscoped_q=faker

akhilesh26:
@MonkeyDo, Please check it once again, mainly setup related things. I will try to complete the first-week work in this weekend.

akhilesh26:
@MonkeyDo Insertion of data in the test database is creating some problem on running the test, please check it.

src/api/app.js
```diff
@@ -0,0 +1,186 @@
...
+// Set up constants that will remain valid for the life of the app
+let siteRevision = 'unknown';
+git.short((revision) => {
+	siteRevision = revision;
+});
+
+const repositoryUrl = 'https://github.com/bookbrainz/bookbrainz-site/';
+
+app.use((req, res, next) => {
+	// Set up globally-used properties
```
MonkeyDo: Lines 118 to 130 can be removed

src/api/app.js
```diff
@@ -0,0 +1,186 @@
...
+	saveUninitialized: false,
+	secret: config.session.secret,
+	store: new RedisStore({
+		host: _get(config, 'session.redis.host', 'localhost'),
+		port: _get(config, 'session.redis.port', 6379)
+	})
+}));
+
+
+// Set up constants that will remain valid for the life of the app
```
MonkeyDo: Lines 109 to 113 can be removed

src/api/app.js
```diff
@@ -0,0 +1,186 @@
...
+	app.use(logger('dev'));
+}
+
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({
+	extended: false
+}));
+app.use(compression());
+
+// Set up serving of static assets
```
MonkeyDo: Lines 69 to 91 included can be removed

src/api/app.js
```diff
@@ -0,0 +1,186 @@
...
+
+// Initialize application
+const app = express();
+app.locals.orm = BookBrainzData(config.database);
+
+const rootDir = path.join(__dirname, '../../');
+
+app.set('trust proxy', config.site.proxyTrust);
+
+app.use(favicon(path.join(rootDir, 'static/images/icons/favicon.ico')));
```
MonkeyDo: This line can be removed

src/api/helpers/formatWorkData.js
```diff
@@ -1,47 +1,66 @@
...
+				sortName: work.defaultAlias.sortName,
+				aliasLanguage: work.defaultAlias.language.name
+			},
 			disambiguation: work.disambiguation ?
 				work.disambiguation.comment : null,
 			languages: work.languageSet ?
-                work.languageSet.languages.map((language) => language.name) : null,
-            workType: work.typeId,
+				work.languageSet.languages.map((language) => language.name) : null,
+			workType: work.workType.label,
```
akhilesh26: Please update test-helper with inserting the workType, the test is failing at `work.workType.label`.

MonkeyDo:
I took the liberty to refactor some of the helpers to common code, as it's an annoying job and I want you to be able to focus on the API and tests.
This was to make sure that when we build the API we don't depend on any client or server code, to separate the projects as much as possible (using the common code folder)

src/api/helpers/formatWorkData.js
```diff
@@ -0,0 +1,66 @@
...
+
+/* eslint-disable */
+import _ from 'lodash';
+
+export function getWorkBasicInfo(work: object) {
+	return _.isNil(work) ? null :
+		{
+			bbid: work.bbid,
+			defaultAlias: {
+				name: work.defaultAlias.name,
```
MonkeyDo: Here and everywhere else where accessing a nested property (that can throw an uncaught error like  "cannot find property 'name' of undefined"), please use lodash's _.get method which does so safely (without throwing an error).

You should also use the same method to always return a default value ([`_.get(object, "path", default)`](https://lodash.com/docs/4.17.11#get)) that the client can expect (for example `_.get(work, 'disambiguation.comment', null)`)
For example, for languages, `languages: _.get(work, 'languageSet.languages', []).map((language) => language.name)` will ensure that you can use .map , and that you always return an array (empty array in this case) so that the consumer knows it is safe to use array methods when using results from the api.

src/api/routes/work.js
```diff
@@ -0,0 +1,57 @@
...
+	const relations = ['aliasSet.aliases.language'];
+	const workWithAliases = await getWorkFromDB(req, relations);
+	if (_.isNil(workWithAliases.aliasSet)) {
+		return res.status(404).send({message: 'This Work is not founded'});
+	}
+	const workAliasesList = getWorkAliases(workWithAliases.toJSON());
+	return res.status(200).send(workAliasesList);
+});
+
+router.get('/:bbid/identifiers', async (req, res, next) => {
```
MonkeyDo: Trying to get a non existant bbid throws an uncaught error, and the request doesn't finish:

`(node:9285) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'identifierSet' of null
    at router.get (/Users/Lenore/Work/Brainz/BookBrainz/bookbrainz-site-docker/lib/api/routes/work.js:62:47)
(node:9285) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:9285) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.`

akhilesh26:
Hello @MonkeyDo,
Please review the changes and newly pushed code when you have time. Some tests are failing because of the area addition in create-entity, please update the create-entity test helper by adding `beginArea` and `endArea`  in createAuhtor function so that the author will be created successfully in tests.

src/api/app.js
```diff
@@ -0,0 +1,125 @@
...
+	})
+}));
+
+
+// Set up routes
+routes(app);
+
+// Catch 404 and forward to error handler
+app.use((req, res, next) => {
+	next(new error.NotFoundError(null, req));
```
MonkeyDo: This returns a simple web page with 404 status and the error as text.
This was appropriate for the web server, but that's not what we want for the API.

Instead, we'll want something simpler like:
```
app.use((req, res, next) => {
	res.status(404).send(`Incorrect endpoint ${req.baseUrl}`);
}
```

MonkeyDo: You'll also want to add tests where you call an inexistant endpoints (for example GET `/not-an-endpoint`), as well as for each existing endpoint an example of unsupported path (for example `/work/5992275-901b-415f-9e93-b3c0ca518455/not-valid`)
The tests should expect 404 responses.

PR #289, opened 2019-06-05
fix: fix BB-341 disable login and register links when redis is down

code-master5:
Adds support for disabling the login and register links (one in the top menu bar and one in the middle of the homepage) when a user can't login (Redis session missing).

### DIFF
--- a/src/server/routes/index.js
+++ b/src/server/routes/index.js
@@ -52,15 +52,25 @@ router.get('/', async (req, res, next) => {
 			requireJS: Boolean(res.locals.user)
 		});
 
+		const disableSignUp = props.alerts.some(
+			alert => alert.message.includes(
+				'signing in and signing up are disabled'
+			)
+		);
+
 		/*
 		 * Renders react components server side and injects markup into target
 		 * file object spread injects the app.locals variables into React as
 		 * props
 		 */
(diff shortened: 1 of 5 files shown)


MonkeyDo:
Hi @code-master5 !
Thanks for the PR, that was a nice and easy fix.

I took the liberty of modifying the CSS, because the style for disabled navbar buttons used the same colour for the text and background, making it unreadable. I'll open a PR on the theme itself (https://github.com/metabrainz/lobes) for that same fix.

src/server/routes/index.js
```diff
@@ -52,15 +52,25 @@ router.get('/', async (req, res, next) => {
 			requireJS: Boolean(res.locals.user)
 		});
 
+		const disableSignUp = props.alerts.some(
```
MonkeyDo: A safer way to do this is set `req.signUpDisabled` (renamed for clarity) in `app.js` in this block: 
https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/app.js#L132-L138

Then instead of doing `props.alerts.some(…`, you can check `req.signUpDisabled`

code-master5:
Hi @MonkeyDo ! 

I have made some changes according to your suggestions. Please review when you've time.

PR #297, opened 2019-08-04
feat(browseRequests): add browse endpoints with test and documentation in progress

akhilesh26:
### Problem

Implementation of Browse endpoints.

### Solution

Start writing tests for those endpoints


akhilesh26:
Hi @MonkeyDo,
Here the initial phase of tests for browse endpoints. My idea is to return the list of entity bbid instead of entity data. First, User will browse the entity bbid related to another entity, Then fetch the detail of an entity with BBID by using lookup requests if required. What is your plan to browse endpoints? 
Suggest more endpoints for browse requests with the requirement explanation and check these test written in this PR. Thanks!

akhilesh26:
Hi @MonkeyDo
I refactored test cases according to option-4, Please look it once for more improvements, When you have time. Thanks!

test/src/api/helpers.js
```diff
@@ -0,0 +1,138 @@
...
+	expect(res.body.editions[0].entity.relationshipTypeID).to.be.a('interger');
+}
+
+export async function testEditionGroupBrowseRequest(url) {
+	const res = await chai.request(app).get(url);
+	expect(res.status).to.equal(200);
+	expect(res.body).to.be.an('object');
+	expect(res.body).to.have.all.keys(
+		'bbid',
+		'edtitionGroups'
```
MonkeyDo: editionGroups, not edtitionGroups

MonkeyDo: Same applies to lines 100 -> 110

src/api/helpers/middleware.js
```diff
@@ -1,18 +1,58 @@
...
+
 
 import * as commonUtils from '../../common/helpers/utils';
 import _ from 'lodash';
+import log from 'log';
 
 
 /* eslint-disable*/
 
+export function validateAuthorBrowseRequest (req, res, next) {
```
akhilesh26: Please help to validate browse request here and This function return URL not valid if the request has invalid query params.

src/api/helpers/utils.js
```diff
@@ -40,3 +43,62 @@ export function allowOnlyGetMethod(req, res, next) {
...
+	const filteredRelationships = await relationships
+		.map(relationship => {
+			// Current entity is the source of the relationship
+			if (entity.bbid === relationship.sourceBbid) {				
+				// The other entity is of the browsed type we are looking for
+				// We need a good way to compare entity type strings here and same thing below
+				// Allow for capitalization mistakes? (.toLowercase() on both?)
+				if (relationship.targetEntityType === browsedEntityType) {
+					return {
+						entity: relationship.target,
```
akhilesh26: How to load entity basic detail here? relationship.target and relationship.source have only bbid and entity type.

akhilesh26:
Hi @MonkeyDo,
Browse request is working for `/author?work=<bbid>` and `/author?edition=<bbid>`. Please help to get the required format response and validate the URL for the author's browser request as commented above. I will implement all browse endpoint on the basis of this.

src/api/routes/edition.js
```diff
@@ -305,8 +305,16 @@ router.get('/',
 	loadEntityRelationshipsForBrowse(),
 	async (req, res, next) => {
 		function relationshipsFilterMethod(relatedEntity) {
-			if (req.query.type) {
-				return _.toLower(_.get(relatedEntity, 'editionType.label')) === req.query.type;
+			const edtionFormatMatched = _.toLower(relatedEntity.editionFormat) === req.query.format;
+			const editionLanguageMatched = relatedEntity.includes(_.upperFirst(req.query.language));
```
akhilesh26: Here is used the format filter, instead of status. the format is more meaningful than status. Edition formats are paperback, hardcover, ebook extra.

akhilesh26:
@MonkeyDo Hi,
I updated filter validation according to the proposal, I think, all functionalities working now for browse request except edition  & edition-group case. Please go through once and let me know if anything more on functionality. Then I will update documentation and tests also. Thanks!

src/api/routes/author.js
```diff
@@ -209,10 +238,64 @@ router.get('/:bbid/identifiers',
...
+			}
+			return true;
+		}
+		const authorRelationshipList = await utils.getBrowsedRelationships(
+			req.app.locals.orm, res.locals, 'Author',
+			getAuthorBasicInfo, authorBasicRelations, relationshipsFilterMethod
+		);
+		return res.status(200).send({
+			bbid: req.query.bbid,
+			relatedAuthors: authorRelationshipList
```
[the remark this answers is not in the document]
akhilesh26: it gives the clarity on response.

akhilesh26:
@MonkeyDo Hi!
Thanks for the detailed review, I will try to resolve all these ASAP. The feeling was not well in the because of the headache in last days, so not able to complete these, but I will try to complete all ASAP. Thanks for such support.
I added all most all the tests for work browse request, I request you to look at those and if they are implemented correctly then I will extend for all entities.

src/api/routes/author.js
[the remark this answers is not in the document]
MonkeyDo: I see your point.
I would then suggest rolling back to what you originally had in the tests and proposal, and removing the `related` part to have the response structure be `{bbid:…, authors:[…]}`.

endurance21:
@MonkeyDo
I see this work, unfinished 
From where should I get started  on this?..

MonkeyDo:
Hi @endurance21!
It would be quite useful if you wanted to pick this API project back up! I believe most of the coding part for this PR is completed, and only the tests require completion.

The API has lookup endpoints and browse endpoints.
The lookup endpoints you can test live (with its live documentation along with it) at https://api.test.bookbrainz.org/1/api-docs/
You can find a description of the entire project here (a GSOC project from last year): https://community.metabrainz.org/t/gsoc-2019-json-web-api-for-bookbrainz/421988

I think the best way to get started on it would be to do a PR review yourself, meaning reading through this PR code and pointing out anything that looks incorrect or could be improved upon
.
Then the next obvious step would be to fix the requested changes from my last review.

After that, we can take it from there and I will do another full review with a view to getting this PR merged :)

Don't hesitate if you have questions!

endurance21:
@MonkeyDo
Okay.

MonkeyDo:
I forgot some useful details you can find here: https://github.com/bookbrainz/bookbrainz-site#running-the-api

prabalsingh24:
Screenshot_2020-04-24 Screenshot(1)


This is the response for `http://localhost:9098/1/work?author=4e8f8704-7c34-4a00-8ce5-f02bf26ca7df`

Do you think this should be having alias of the author?

MonkeyDo:
> This is the response for `http://localhost:9098/1/work?author=4e8f8704-7c34-4a00-8ce5-f02bf26ca7df`
> 
> Do you think this should be having alias of the author?

No, we want people to fetch that —possibly beforehand— with `/author/4e8f8704-7c34-4a00-8ce5-f02bf26ca7df`

test/test-helpers/create-entities.js
```diff
@@ -263,20 +263,32 @@ export async function createEdition(optionalBBID) {
...
+		};
+	}
+	const workType = await new WorkType({id: workAttribs.typeId, label: `Work Type ${workAttribs.typeId}`})
+		.fetch({
+			require: false
+		});
+	if (!workType) {
+		await new WorkType({id: workAttribs.typeId, label: `Work Type ${workAttribs.typeId}`})
+			.save(null, {method: 'insert'});
+	}
```
prabalsingh24, line 291: What do you think of this @MonkeyDo ? Bookshelf don't really have an upsert option. Do you know any better way to get this done?

prabalsingh24:
@MonkeyDo Have a look at this when you're free :)

test/test-helpers/create-entities.js
MonkeyDo: Unfortunately not :/

src/api/routes/edition.js
```diff
@@ -223,10 +250,82 @@ router.get('/:bbid/identifiers',
...
+ *             $ref: '#/definitions/BrowsedEditions'
+ *       404:
+ *         description: Related entity not found
+ *       406:
+ *         description: Invalid BBID paased in query params
+ */
+
+router.get('/',
+	validateBrowseRequestQueryParameters(['author', 'edition', 'edition-group', 'work', 'publisher']),
+	makeEntityLoader(null, utils.relationshipsRelations, 'Entity not found', true),
```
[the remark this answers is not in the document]
MonkeyDo: @prabalsingh24 The comment above is about loading related entities that do not appear in relationships.

There are two cases where this happens, both for Edition entities: Edition Group and Publisher.
Both can be defined as relationships (Edition has Edition Group… and Edition was published by…), but I am fixing these and replacing them respectively with the editionGroupBbid and publisherSetId columns on the edition_data table.
Don't ask me why they were modified like that, it was before my time and I never got a good answer for it.

So for Edition<->Edition Group link the comment above was implemented (search for `if (req.query.modelType === 'EditionGroup') …` and `if (req.query.modelType === 'Edition') …`).

Now the same should be implemented for Edition<->Publisher links. An Edition entity can have a publisherSet with an array of Publisher entities in it (currently the UI only allows adding one. go figure…).
The Edition ORM model has a `publisherSet()` method and the Publisher ORM model has an `editions(fetchOptions)` method to simplify all this.

To illustrate the issue, here's an example of a publisher that appears in two Editions' publisherSet: https://bookbrainz.org/publisher/9dc3a2f2-3d7d-4dda-b25b-b6de0de898fd
Fetching  http://localhost:9098/1/edition?publisher=9dc3a2f2-3d7d-4dda-b25b-b6de0de898fd returns no results, since they are not defined as relationships.

src/api/routes/publisher.js
```diff
@@ -203,10 +236,75 @@ router.get('/:bbid/identifiers',
...
+ *             $ref: '#/definitions/BrowsedPublishers'
+ *       404:
+ *         description: Work not found or Edition not found
+ *       406:
+ *         description: Invalid BBID paased in query params
+ */
+
+router.get('/',
+	formatQueryParameters(),
+	validateBrowseRequestQueryParameters(['author', 'edition', 'work', 'publisher']),
```
MonkeyDo, line 280: As commented above, there is are extra special cases for links between Publisher and Edition that are not defined by relationships.
http://localhost:9098/1/publisher?edition=978c6f53-61e4-48cd-a75c-f220374215b3 returns an empty array where I would expect this entity to be returned: https://bookbrainz.org/publisher/9dc3a2f2-3d7d-4dda-b25b-b6de0de898fd

test/src/api/routes/test-edition-group.js
```diff
@@ -151,3 +161,83 @@ describe('GET /EditionGroup', () => {
...
+	});
+
+	it('should return list of EditionGroups associated with the edition (with Type Filter)', async () => {
+		const res = await chai.request(app).get(`/edition-group?edition=${edition.get('bbid')}&type=Edition+Group+Type+1`);
+		await browseEditionGroupBasicTests(res);
+		expect(res.body.editionGroups.length).to.equal(1);
+		expect(_.toLower(res.body.editionGroups[0].entity.editionGroupType)).to.equal('edition group type 1');
+	});
+
+	it('should return list of EditionGroups associated with the edition (with Type Filter)', async () => {
```
MonkeyDo, line 201: Same test description as the test above.

MonkeyDo: I also didn't see any test like that for the other entity types browse endpoints with non-existing types

test/src/api/routes/test-publisher.js
```diff
@@ -153,3 +163,139 @@ describe('GET /Publisher', () => {
...
 });
 
+/* eslint-disable no-await-in-loop */
+describe('Browse Publishers', () => {
+	let work;
+	before(async () => {
+		await truncateEntities();
+		// create 10 publishers 5 each of type 1 and type 2
+		const publisherBBIDs = [];
+		for (let areaId = 1; areaId <= 5; areaId++) {
```
MonkeyDo, line 173: Why 5 of each where 2 of each would do?

test/src/api/routes/test-work.js
```diff
@@ -84,72 +93,264 @@ describe('GET /work', () => {
...
+			null,
+			null,
+			[{id: frenchId}]
+		);
+
+		// create 10 works. 5 of French Language and 5 of English Language
+		// with 2 works of each workType ( one of french and one english)
+		const workBBIDs = [];
+		const workAttrib = {};
+		for (let workTypeId = 1; workTypeId <= 5; workTypeId++) {
```
MonkeyDo, line 219: Why 5 of each where 2 would do?

src/api/routes/publisher.js
prabalsingh24: Oh okayy. I didn't know this. i thought it was only for edition-editionGroup :sweat_smile:

test/src/api/routes/test-publisher.js
prabalsingh24: Umm I don't know why I did this. I'll change it to 2. :D

prabalsingh24:
> I think instead a 400 should be returned. Along with the existing error message ('BBID is not valid uuid') I think that's clear enough.


Aaah, This makes sense now. I was wondering why we did 400 error for this [invalid-bbid-pr](https://github.com/bookbrainz/bookbrainz-site/pull/402) and 406 here.

src/api/routes/publisher.js
MonkeyDo: I too forgot about it !

src/api/routes/edition-group.js
```diff
@@ -228,10 +254,16 @@ router.get('/',
...
+			const {bbid} = req.query;
 			const relationships = editionGroupBasicRelations.map(rel => `editionGroup.${rel}`);
-			await makeEntityLoader(null, relationships, 'Entity not found', true, false)(req, res, next);
-			const {editionGroup} = res.locals.entity;
-			// edition will belong to only one edition-group
+			const edition = await new Edition({bbid}).fetch({
+				require: false,
+				withRelated: relationships
+			});
+			const editionJSON = edition ? edition.toJSON() : {};
```
MonkeyDo, line 264: Here it looks like you're loading the Edition twice: once in the `makeEntityLoader` middleware definde on the route and once manually here.
This `await new Edition({bbid})` could be remove, as res.locals.entity already has the entity. What it doesn't have is the related EditionGroup (the extra editionGroup.XXX relationships defined a line above), which could be added to the `makeEntityLoader` call.
In this case for EditionGroups that can only have relationships to Editions its easy enough:
```makeEntityLoader(null, utils.relationshipsRelations.concat(editionGroupBasicRelations.map(rel => `editionGroup.${rel}`)), 'Entity not found', true),```

However to make it work for other entities, you could add a new flag to `makeEntityLoader` which, when set to true, would also load related `relations.map(rel => ${_.camelCase(modelType)}.${rel})`.
That might work.

If you don't want to add a flag, you could also check if `req.baseUrl`, `isBrowse`  and `req.query.modelType` and make a decision based on that:
For example, if `isBrowse` is true and baseUrl ends with `/edition-group` and `modelType` is Edition, then also load those extra relationships `relations.map(rel => ${_.camelCase(modelType)}.${rel})`. Seems more convoluted to me and moves some logic in makeEntityLoader instead of a simple flag on the browse route, but it's doable if a new flag is an issue.

File: src/api/routes/author.js [hunk 13]
```diff
@@ -204,15 +233,66 @@ router.get('/:bbid/identifiers',
...
+ *     operationId: getRelatedAuthorByBbid
+ *     produces:
+ *       - application/json
+ *     parameters:
+ *       - name: author/edition/edition-group/publisher
+ *         in: query
+ *         description: BBID of the corresponding entity
+ *         required: true
+ *         type: bbid
+ *     responses:
```
MonkeyDo, line 264: Another parameter is missing here:
```
name: type
in: query
description: filter by author type
type: string
enum: [person, group]
```

src/api/routes/edition-group.js
```diff
@@ -183,15 +194,87 @@ router.get('/:bbid/identifiers',
...
+ *     operationId: getRelatedPublisherByBbid
+ *     produces:
+ *       - application/json
+ *     parameters:
+ *       - name: edition
+ *         in: query
+ *         description: BBID of the edition
+ *         required: true
+ *         type: bbid
+ *     responses:
```
MonkeyDo, line 225: Another parameter is missing here:
```
name: type
in: query
description: filter by Edition Group type
type: string
enum: [book, leaflet, newspaper, magazine, journal]
```

I'm not sure how smart it is to have an enum when new elements can be added in the DB, but It might still be nicer to have it in the docs than leave it to the users to find out which types exist

src/api/routes/edition.js
```diff
@@ -218,15 +254,109 @@ router.get('/:bbid/identifiers',
...
+ *     operationId: getRelatedEditionByBbid
+ *     produces:
+ *       - application/json
+ *     parameters:
+ *       - name: author/edition/edition-group/publisher/work
+ *         in: query
+ *         description: BBID of the corresponding entity
+ *         required: true
+ *         type: bbid
+ *     responses:
```
MonkeyDo, line 285: Two parameters missing here:
```
name: format
in: query
description: filter by Edition format
type: string
enum: [Paperback, Hardcover, eBook, Library Binding, Audiobook]
```
and
```
name: language
in: query
description: filter by Edition language (ISO 639 three-letter code)
type: string
```

src/api/routes/edition.js
```diff
@@ -218,15 +254,109 @@ router.get('/:bbid/identifiers',
...
+	formatQueryParameters(),
+	validateBrowseRequestQueryParameters(['author', 'edition', 'edition-group', 'work', 'publisher']),
+	makeEntityLoader(null, utils.relationshipsRelations, 'Entity not found', true),
+	loadEntityRelationshipsForBrowse(),
+	async (req, res, next) => {
+		function relationshipsFilterMethod(relatedEntity) {
+			if (req.query.format) {
+				const editionFormatMatched = _.toLower(relatedEntity.editionFormat) === _.toLower(req.query.format);
+				if (req.query.language) {
+					const editionLanguageMatched = relatedEntity.languages.includes(_.upperFirst(_.toLower(req.query.language)));
```
MonkeyDo, line 306: This should filter by ISO 639 two- or three-letter code rather than by name. (I think)
Same below.

src/api/routes/publisher.js
```diff
@@ -198,15 +231,93 @@ router.get('/:bbid/identifiers',
...
+ * '/publisher':
+ *   get:
+ *     tags:
+ *       - Browse Requests
+ *     summary: Gets a list of Publishers which are related to the other entity
+ *     description: BBID of an Author or an Edition or an EditionGroup or a Publisher is passed as a query parameter, and it's related Publishers are fetched
+ *     operationId: getRelatedPublisherByBbid
+ *     produces:
+ *       - application/json
+ *     parameters:
```
MonkeyDo, line 256:
```
name: type
in: query
description: filter by Publisher type
type: string
enum: [publisher distributor]
```
and

```
name: language
in: query
description: filter Publisher by area 
type: string
```

test/src/api/routes/test-author.js
```diff
@@ -156,3 +165,117 @@ describe('GET /Author', () => {
...
+	it('should throw 404 error for incorrect bbid', (done) => {
+		chai.request(app)
+			.get(`/author?work=${aBBID}`)
+			.end(function (err, res) {
+				if (err) { return done(err); }
+				expect(res).to.have.status(404);
+				return done();
+			});
+	});
+});
```
MonkeyDo, line 281: You could also add a test to check the a 400 error is returned if  passing more than one type parameter, for all entity types expect Edition Group

src/api/routes/edition-group.js
MonkeyDo: I ended up solving this with a custom middleware in the browse routes definitions that calls makeEntityLoader with the extra related elements to avoid fetching again in the route's main body.

There is one that cannot be optimised currently: `/edition?publisher=…`, since Publisher().editions() !== Publisher.related('editions') (https://github.com/bookbrainz/bookbrainz-data-js/blob/master/src/models/data/publisherData.js#L36).
This could possibly be refactored to use [hasMany().through()](https://bookshelfjs.org/api.html#Model-instance-through) to allow fetching with the Publisher.fetch({withRelated: 'editions.xyz'}) format.

MonkeyDo:
:tada:

src/api/routes/author.js
```diff
@@ -250,31 +250,26 @@ router.get('/:bbid/relationships',
...
- *         description: BBID of the Work
- *         required: false
- *         type: bbid
- *       - name: edition
- *         in: query
- *         description: BBID of the Edition
- *         required: false
+ *         description: BBID of the corresponding entity
+ *         required: true
  *         type: bbid
```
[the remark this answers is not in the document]
MonkeyDo: Will be addressed in PR #461

PR #298, opened 2019-08-08
Deployment Docker setup for test API

MonkeyDo:
Setting up the API Docker image build target, and all the files needed to deploy webservice-test

Also set up docker-compose file for running the api locally with Docker.
The `elasticsearch` section of the docker-compose file is currently commented out, as we don't use the search at this point in the API development. It will need to be commented out once the search endpoints have been implemented.


MonkeyDo:
@akhilesh26 I'm making quite a few changes with this PR, so let me know if you need any help to rebase #297

PR #299, opened 2019-08-16
feat(search): add search endpoint for API

akhilesh26:

### DIFF
--- a/src/api/routes/search.js
+++ b/src/api/routes/search.js
@@ -16,21 +16,21 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import * as search from '../../common/helpers/search';
-import {keys as _keys, snakeCase as _snakeCase, isNil} from 'lodash';
+import 	* as search from '../../common/helpers/search';
+import {snakeCase as _snakeCase} from 'lodash';
 import express from 'express';
-import {formatSearchRespose} from '../helpers/formatEntityData';
+import {formatSearchResponse} from '../helpers/formatEntityData';
 
 
 const router = express.Router();
 
 /**
  *	@swagger
- * '/search?q=Harry&colloction=work&limit=5&offset=0':
+ * '/search?q=Harry&type=work&size=5&from=0':
  *   get:
  *     tags:
  *       - Search Requests
- *     summary: Get the list of entity accourding to query and query params
+ *     summary: Get a list of entities according to search query and params
  *     description: query parameter is used to match in the default alias and other parameters are optional
  *     operationId: getSearchResult
  *     produces:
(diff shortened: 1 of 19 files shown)


akhilesh26:
Hi @MonkeyDo,
When the search request goes from the test-search.js, the result does not come from elasticsearch properly that's why tests are failing. When the same request goes from Postman, the proper response is coming. I tried to find the issue but unable to reach. Please help to resolve this.
my test.json is [here](https://gist.github.com/akhilesh26/af331716418225dfd498d031be12279f)

akhilesh26:
Hi @MonkeyDo,
Please review this also, when you have time. Let me know if any changes will be required.

MonkeyDo:
I've done a bit of reading, and come to the conclusion the easiest way to set up the tests is to mimic ElasticSearch without running any real instance.
You can use the `nock` package to intercept calls to ElasticSearch (using the values defined in the test.json config file as it is done [here](https://github.com/bookbrainz/bookbrainz-site/blob/da362b4612283ffefd917eee229e4d26b54bd19d/test/bookbrainz-data.js#L20) ) and return a fake response that mimics ElasticSearch.

Here is what a usual ES response looks like:

MonkeyDo:
After some tinkering, I was able to set Travis to run an instance of elasticsearch when running the tests, so we don't need to try to mock it. 
My attemps at mocking ES were pretty unsuccessful, and I think this is better in any case.

I also modified the tests to create more entities, and ensure only the correct subset is returned (with the right name, and with the collection filtering)

src/api/routes/search.js
```diff
@@ -0,0 +1,80 @@
...
+ *       200:
+ *         description: List of relationships with BBID of the Publisher entity
+ *         schema:
+ *             $ref: '#/definitions/SearchResultModel'
+ */
+
+router.get('/', async (req, res) => {
+	const {orm} = req.app.locals;
+	const query = req.query.q;
+	const collection = req.query.collection || null;
```
MonkeyDo: After some thought, instead of `collection`, I think it would be clearer to use `type` to represent the entity type, as we use the term 'entity type' in the api docs in various places.

It would also be closer to the MusicBrainz implementation which is always a plus: https://musicbrainz.org/doc/Development/XML_Web_Service/Version_2/Search

And finally, if we end up having user collections in the future and decide to call them 'collection', it will prevent confusion.

akhilesh26: I did `entityType`, it is more clear than `type` because the type is used to under entity.  Means type of work, type of publisher extra.

prabalsingh24:
@MonkeyDo 
```https://test.bookbrainz.org/search?q=harry&collection=author```
I think we should change `collection` -> `type` in the API and the main website as well.
We're displaying them as `All Types` in the search bar, I think `type` makes more sense. ( It will get confusing with user-collections too :P)

**Comment by prabalsingh24** [comment]
also I am changing `offset` --> `from` and `limit` --> `size` in the api to make it same as the website

**Comment by MonkeyDo** [comment]
> @MonkeyDo
> `https://test.bookbrainz.org/search?q=harry&collection=author`
> I think we should change `collection` -> `type` in the API and the main website as well.
> We're displaying them as `All Types` in the search bar, I think `type` makes more sense. ( It will get confusing with user-collections too :P)

I agree entirely!
I had pretty much the same arguments in a previous review comment :p
https://github.com/bookbrainz/bookbrainz-site/pull/299#discussion_r318149989

**Comment by prabalsingh24** [comment]
> I had pretty much the same arguments in a previous review comment :p

Oops, I didn't see that :P

File: src/api/routes/search.js [hunk 2]
```diff
@@ -0,0 +1,89 @@
...
+ *         in: query
+ *         description: Search string for entity search
+ *         required: true
+ *         type: string
+ *       - name: type
+ *         in: query
+ *         description: Entity type for search
+ *         required: false
+ *         type: string
+ *         default: null
```
MonkeyDo, line 49: you can also add `enum: [author, edition, edition-group, publisher, work]`

test/src/api/routes/test-search.js
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright (C) 2019  Akhilesh Kumar
+ *
```
MonkeyDo, line 3: Wanna add your name?

src/common/helpers/utils.js
```diff
@@ -19,3 +19,24 @@ const _bbidRegex =
...
+
+export function getEntityModelByType(orm: Object, type: string): Object {
+	const entityModels = getEntityModels(orm);
+
+	if (!entityModels[type]) {
+		throw new Error(`Unrecognized entity type: '${type}'`);
+	}
+
+	return entityModels[type];
+}
```
MonkeyDo, line 42: There are now two versions of these utility functions
The ones in  /src/server/helpers/utils.js have JSDoc comments.

It would be good to copy over the JSDoc in src/common/helpers/utils.js, and use only those everywhere in src/server (that's why they're in /common, so they can be used by API and server)

prabalsingh24:
We will have to modify this, if/when this [423](https://github.com/bookbrainz/bookbrainz-site/pull/423) gets merged

PR #300, opened 2019-08-19
feat(API): add basic rate limiter

akhilesh26:
Add rate limiting to the API routes.
Slows down the response after X requests in a specifid time window.

Initial settings are: 
- browse and lookup endpoints: 100 requests /  5 minutes, then += 500ms* requests after that
- search endpoint: 100 requests /  5 minutes, then += 500ms* requests after that

Request # 101 is delayed by 500ms, # 102 is delayed by 1000ms, etc.

This will prevent over server from DOS attacks and control the request traffic

Slow Down configurations are only, for example, We can update those after a discussion on IRC.

### Problem

Sever may slow or down on high request traffic.

### Solution

Implemented request slow down with configuration.  To implement this [express-slow-down](https://www.npmjs.com/package/express-slow-down) is used.

### Areas of Impact

Server-side

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,44 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@apidevtools/json-schema-ref-parser": {
+      "version": "8.0.0",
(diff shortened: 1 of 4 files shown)


MonkeyDo:
This look sane and easy to configure, good work!
I tested locally and it's working great.

I need to discuss with other team members what the limits should be, and also do some test as to how to deploy it before I merge this.

PR #305, opened 2019-10-23
BB-353 fix: Editing some author entities return 500 error.

panwarabhishek345:
### Problem
BB-353 fix: Editing some author entities return 500 error

### Solution

### Areas of Impact


panwarabhishek345:
@MonkeyDo This new PR is fine?

PR #306, opened 2019-10-23
BB-353 fix: Editing some author entities return 500 error. 

panwarabhishek345:
### Problem
#305 BB-353 fix: Editing some author entities return 500 error

### Solution

### Areas of Impact

### DIFF
--- a/src/client/containers/layout.js
+++ b/src/client/containers/layout.js
@@ -25,24 +25,24 @@
 /* eslint import/no-commonjs: "warn" */
 /* eslint global-require: "warn" */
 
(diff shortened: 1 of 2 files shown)


MonkeyDo:
@panwarabhishek345 I would like to see a clean PR with only the changes related to the 500 error.
These new ones still have the three commits, only one being related to the 500 error.
Considering it's a small change on a single file, the easiest would be creating an entirely new branch based on master, re-creating your changes, committing and opening a new PR from there.

panwarabhishek345:
Alright I will do it

PR #307, opened 2019-10-23
add passing warnMessage from new-date-field

sbvkrishna:
### Problem
**Validation text for name (defaultAlias) shows the one for dates**
Issue [BB-354](https://tickets.metabrainz.org/browse/BB-354) 

### Solution
From **new-date-field**, passed `warnMessage` as a prop to **validation-label** 

### Areas of Impact
Changes in **new-date-field** and **validation-label**

### DIFF
--- a/src/client/entity-editor/common/validation-label.js
+++ b/src/client/entity-editor/common/validation-label.js
@@ -85,10 +85,11 @@ function ValidationLabel({
 	empty,
 	error,
 	errorMessage,
+	warnMessage,
 	warn
 }: Props) {
-	const warnElement = (warn && !empty && !error) &&
-		<span className={contextualColor(empty, error, warn)}> Are you sure? You entered a date in the future!</span>;
+	const warnElement = (warn && warnMessage && !empty && !error) &&
+		<span className={contextualColor(empty, error, warn)}> {warnMessage} </span>;
 	const errorElement = errorMessage &&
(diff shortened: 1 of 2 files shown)


src/client/entity-editor/common/validation-label.js
```diff
@@ -85,10 +85,11 @@ function ValidationLabel({
 	empty,
 	error,
 	errorMessage,
+	warnMessage,
 	warn
 }: Props) {
-	const warnElement = (warn && !empty && !error) &&
-		<span className={contextualColor(empty, error, warn)}> Are you sure? You entered a date in the future!</span>;
+	const warnElement = (warn && warnMessage && !empty && !error) &&
```
MonkeyDo: There is no need to check for the presence of `warnMessage` here, for two reasons:
If no `warnMessage` prop is passed to the component, it will default to an empty string (see my other comment).
Also, the warning is used in most cases without a warningMessage, so we want to make it works without it. (now I am wondering if that should be the same for `errorMessage &&` on the line below. I will investigate and possibly open another issue.)

sbvkrishna:
I have resolved the linting issues. Will make a commit now

MonkeyDo:
@sbvkrishna I'll let you close the issue yourself, it's satisfying to got to close an issue after you worked on it :)
 https://tickets.metabrainz.org/browse/BB-354

PR #308, opened 2019-10-23
fix(profile):(Updated)Fixed the 500 error when Editing some author entities.

panwarabhishek345:
### Problem
#305 BB-353 fix: Editing some author entities return 500 error

### Solution

### Areas of Impact

### DIFF
--- a/src/server/routes/entity/author.js
+++ b/src/server/routes/entity/author.js
@@ -31,7 +31,7 @@ import {
 } from '../../helpers/entityRouteUtils';
 
 import _ from 'lodash';
-import {escapeProps} from '../../helpers/props';
+import { escapeProps } from '../../helpers/props';
 import express from 'express';
 import target from '../../templates/target';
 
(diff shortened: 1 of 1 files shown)


MonkeyDo:
Thanks for reopening that !
The commit history looks clean !

Glad this was an easy fix. This bug was bugging me ! :)
I'll wait for the continuous integration tests to pass and I'll be able to merge.

panwarabhishek345:
Yeah, Hopefully the tests will pass.
Thanks & regards
Abhishek Panwar

MonkeyDo:
@panwarabhishek345 Unfortunately, there are some linting errors that make the tests fail.

I'm not sure what code editor you are using, but it is deciding to make some indentation changes to the file on its own.
I'm not sure what the way is to deactivate that option, and will let you look into that.

Here are the linting issues that prevent the tests from passing; nothing important, just cosmetic: https://travis-ci.org/bookbrainz/bookbrainz-site/jobs/601880523#L593


In the meantime, you can try running the command `./node_modules/.bin/eslint ./src/server/routes/entity/author.js --fix` which will automatically fix the linting issues for this file, and you can commit the result.

panwarabhishek345:
Below the Pull request it says the tests are still pending. Anyway I am
creating a new PR with linting issues resolved.

panwarabhishek345:
I guess there is no need of creating a new PR. It should work if I push my
changes to the 500-error-fix branch. Right?

panwarabhishek345:
Please check the status.

PR #310, opened 2019-10-28
BB-339 fix(display): Converted 500 error to 404 error with a meaningful message, when a revision is not found.

panwarabhishek345:
### Problem

The site displays 500 error when visiting a non-existing revision page

### Solution
Simple condition to check if revision exists or not. If it doesn't exist, 404 error is thrown.

### Areas of Impact
src/server/routes/revision.js


panwarabhishek345:
Thanks, trying to explore the project slowly through these PRs.

PR #311, opened 2019-10-30
fix(display): add 'Revision History not found' message

sbvkrishna:
### Problem

To show meaningful message when User Revision history is empty.

### Solution

Add 'Revision History not found' message if revision history is empty.
Also add "call to action".
### Areas of Impact

- src/client/components/pages/parts/call-to-action.js
- src/client/components/pages/parts/editor-revisions.js
- src/client/components/pages/parts/search-results.js

### DIFF
--- a/src/client/components/pages/parts/editor-revisions.js
+++ b/src/client/components/pages/parts/editor-revisions.js
@@ -30,10 +30,24 @@ import React from 'react';
 const {ListGroup, ListGroupItem} = bootstrap;
 const {formatDate, isWithinDayFromNow} = utilsHelper;
 
+/**
+ * Renders the document and displays the 'EditorRevisionsTab' component.
+ * @returns {ReactElement} a HTML document which displays the 'EditorRevisionTab'.
+ * @param {object} props - Properties passed to the component.
+ */
 function EditorRevisionsTab(props) {
 	const {editor} = props;
 	const {revisions} = editor;
 
+	/**
+	 * Renders the data related to Revision such as 'author' and 'date'.
+	 * It also displays the revison note which is a summary of the changes
+	 * made in the revision.
+	 * @param {object} revision - The revision to be represented by the
+	 * rendered component.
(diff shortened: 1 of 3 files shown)


src/client/components/pages/entity-revisions.js
```diff
@@ -118,10 +118,15 @@ class EntityRevisions extends React.Component {
...
-					{revisions.map(this.renderRevision)}
-				</ListGroup>
+				{revisions.length > 0 ?
+					<div>
+						<h2>Revision History</h2>
+						<ListGroup>
+							{revisions.map(this.renderRevision)}
+						</ListGroup>
+					</div> :
+					<h2>Revision History not found.</h2>
```
MonkeyDo: I think this would be better phrased as "No revision to show", meaning we did find what you are looking for (compared for example to typing a random number in the address bar), but there is nothing to show.

sbvkrishna:
True. Actually I was thinking of [Editor Revision History](https://bookbrainz.org/editor/1499/revisions) having no history and after seeing both entity-revision & editor-revision are used to render revisions, I taught of adding 'Revisions not found' to both of them. 
So only [Editor Revision History](https://bookbrainz.org/editor/1499/revisions) needs it.

src/client/components/pages/parts/editor-revisions.js
```diff
@@ -32,34 +32,40 @@ function EditorRevisionsTab(props) {
...
-							{revision.note}
-						</ListGroupItem>
-					);
-				})}
-			</ListGroup>
+			{revisions.length > 0 ?
+				<ListGroup>
+					{revisions.map(renderRevision)}
+				</ListGroup> :
+				<h4> No revision to show</h4>
```
MonkeyDo: A small change: I think `No revisions to show`, whit the plural, sounds better.

sbvkrishna:
Sounds great. I can JSDoc it too while I'm at it!

PR #312, opened 2019-11-02
fix(core): Fixed the IndexOutOfBoundsException of SchemaSpy.sh

panwarabhishek345:
### Problem
#BB-359 The SchemaSpy.sh script doesn't work as expected. If an empty string is passed as one of the arguments from the terminal, the script fails with IndexOutOfBoundsException.

### Solution
All the args variables are surrounded by double quotes to accept empty string args properly

### Areas of Impact
sql/script/SchemaSpy.sh

### DIFF
--- a/sql/scripts/SchemaSpy.sh
+++ b/sql/scripts/SchemaSpy.sh
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
-if [ $# -ne 4 ]; then
-  echo 'Usage: SchemaSpy.sh [username] [password] [database] [host]';
-fi
+throw_error () {
+  echo "Error, Please provide ${1}" >&2
+  exit 1
+}
 
 if [ ! -e "schemaSpy.jar" ]; then
   wget http://sourceforge.net/projects/schemaspy/files/latest/download?source=files -O schemaSpy.jar;
@@ -12,4 +13,42 @@ if [ ! -e "postgresql-9.4-1202.jdbc41.jar" ]; then
   wget https://jdbc.postgresql.org/download/postgresql-9.4-1202.jdbc41.jar;
 fi
 
-java -jar schemaSpy.jar -t pgsql -host $4 -s bookbrainz -db $3 -u $1 -p $2 -o _build -dp postgresql-9.4-1202.jdbc41.jar;
+while getopts "d:h:p:u:" opt; do
+  case $opt in
+    d)  DATABASE=$OPTARG;;
+    h)  HOST=$OPTARG;;
+    p)  PASSWORD=$OPTARG;;
+    u)  _USERNAME=$OPTARG;;
+    *) echo "Error Occurred, Usage: SchemaSpy.sh -u [username] -p [password] -d [database] -h [host]" >&2
+       exit 1
+  esac
+done
+
+CMD="java -jar schemaSpy.jar -t pgsql -s bookbrainz "
+
+if [ ! -z "$HOST" ]
+then
+  CMD+="-host ${HOST} "
+else
+  throw_error "host: -h [host]"
+fi
+
+if [ ! -z "$DATABASE" ]
+then
+  CMD+="-db ${DATABASE} "
+else
+  throw_error "database: -d [database]"
+fi
+
+if [ ! -z "$_USERNAME" ]
+then
+  CMD+="-u ${_USERNAME} "
+else
+  throw_error "username: -u [username]"
+fi
+
+CMD+="-p '${PASSWORD}' "
+CMD+="-o _build -dp postgresql-9.4-1202.jdbc41.jar;"
+
+echo $CMD
+eval $CMD


panwarabhishek345:
Hello @MonkeyDo! Yes, the other fields shouldn't ever be empty as per me too.

Regarding the fact that we could conditionally handle the password, it is very subjective. I prefer keeping the solution as generalised as possible with lesser complexities. But someone might prefer not passing a password at all if it is empty.

So if you like the conditional solution, I can try that :)

MonkeyDo:
@panwarabhishek345 Ah, I think I wasn't clear, sorry!

I think passing an empty password is a valid case, however passing empty strings to the other flags isn't.
So the password should stay as you did, but for the other flags (-h for host, -d for database and -u for user) you should check if they are empty and not pass the flag at all if they are, as described in that stackoverflow question.

panwarabhishek345:
@MonkeyDo, So you are saying, if any of the three flags are empty then not run the script itself? ( Might as well throw an error)

MonkeyDo:
Ah, no, I mean if any of the other three strings are empty, then don't pass that specific flag.
If [host] is empty, don't pass the -h… flag, for example.

panwarabhishek345:
@MonkeyDo, alright got it. I'll do it. I saw the stack overflow solution properly now.

MonkeyDo:
By the way @panwarabhishek345 , I don't know if you've seen in the [IRC channel logs](https://chatlogs.metabrainz.org/brainzbot/metabrainz/2019-10-29/?msg=4476465&page=1) (you were not online at the time), after you asked me the other day I created this work in progress document explaining the database schema and revision system: https://docs.google.com/document/d/1rEQj7c3jfW_Wvrukr5M57Oy22LePR2NbdRSquPQYsf8

I'd be more than happy to answer any question and help move the document forward. Feel free to add comments on the document itself for any question or clarification

panwarabhishek345:
The doc looks great and will be very helpful for someone to get started.
The SchemaSpy scripts gives only the info about the schema but it doesn't
have the schema description which your docs have. Would read it and if have
doubts, will let you know.

And yes, I am new to open source and IRC, didn't know there is an option to
see the chat logs. But now I know :)

sql/scripts/SchemaSpy.sh
```diff
@@ -12,4 +8,35 @@ if [ ! -e "postgresql-9.4-1202.jdbc41.jar" ]; then
...
+then
+  CMD+="-db ${DATABASE} "
+fi
+
+if [ ! -z "$USERNAME" ]
+then
+  CMD+="-u ${USERNAME} "
+fi
+
+CMD+="-p" "${PASSWORD}"
```
MonkeyDo: Whenever I try to add a `-p mypassword` flag, I get an error `line 39: mypassword: command not found`.

That's because as it is, this line add "-p" to CMD and then _runs_ `${PASSWORD}` instead of adding it  to CMD as well.
You can turns this into `CMD+="-p '${PASSWORD}' "` to add the the flag and the quoted password—to avoid the IndexOutOfBoundsException— to CMD.

panwarabhishek345:
Hi! @MonkeyDo, Just made the appropriate changes.

MonkeyDo:
@panwarabhishek345 I see the password flag line has been change, that looks great now. 

I don't see any change regarding the other flags. Did you see [this comment](https://github.com/bookbrainz/bookbrainz-site/pull/312#pullrequestreview-315465104)? Or am I missing something?

panwarabhishek345:
Oh, my bad. You are referring to throwing of errors when other params are empty right? I didn't see that. Will make the changes soon.

panwarabhishek345:
@MonkeyDo Made the remaining changes.

sql/scripts/SchemaSpy.sh
```diff
@@ -12,4 +13,41 @@ if [ ! -e "postgresql-9.4-1202.jdbc41.jar" ]; then
...
+then
+  CMD+="-u ${_USERNAME} "
+else
+  throw_error "username: -u [username]"
+fi
+
+CMD+="-p '${PASSWORD}' "
+CMD+="-o _build -dp postgresql-9.4-1202.jdbc41.jar;"
+
+eval $CMD
```
MonkeyDo: You could also `echo $CMD` before the eval, to add some readability to the process.
We don't want to overdo it with logging, but I think that one line is a good compromise.

panwarabhishek345:
Oh that was a silly mistake, may be exams are taking a toll on me. Lol! Will make the changes.

panwarabhishek345:
@MonkeyDo Made the appropriate changes.

MonkeyDo:
Good to merge ! 🎉 

Great improvement on this script, and as I mentioned, a good template for future scripts.

panwarabhishek345:
Thank you!

MonkeyDo:
> Thank you!

On the contrary, thank **_you_** !

PR #313, opened 2019-11-02
feat(search): The aliases are now searchable.

panwarabhishek345:
### Problem
#BB-310 Search for entity by alias added and was not possible earlier.

### Solution
Added multi_match query and indexed entity aliases in elastic search.

### Areas of Impact
src/server/helpers/search.js

### DIFF
--- a/src/server/helpers/search.js
+++ b/src/server/helpers/search.js
@@ -24,6 +24,7 @@ import Promise from 'bluebird';
 import _ from 'lodash';
 import httpStatus from 'http-status';
 
+
 const _index = 'bookbrainz';
 const _bulkIndexSize = 128;
 
@@ -160,7 +161,7 @@ export function autocomplete(orm, query, collection) {
 	else {
 		queryBody = {
 			match: {
-				'defaultAlias.name.autocomplete': {
+				'aliasSet.aliases.name.autocomplete': {
 					minimum_should_match: '80%',
 					query
 				}
@@ -215,7 +216,7 @@ export async function generateIndex(orm) {
 		mappings: {
 			_default_: {
 				properties: {
-					defaultAlias: {
+					'aliasSet.aliases': {
 						properties: {
 							name: {
 								fields: {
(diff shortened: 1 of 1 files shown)


MonkeyDo:
:tada:
Good choice of issue, that's a great improvement for an easy solution !

I'm not going to merge this right now, even though the search page part works perfectly, because there are other areas that use the search, and I need to determine if including aliases is what we want everywhere.

For example, when creating a new entity, we do a search based on the name to see if the new entity would be a possible duplicate. I am not 100% sure yet that searching for aliases here would be okay, and it might depend on the entity type.

Capture d’écran 2019-11-05 à 13 17 27

panwarabhishek345:
That's totally fine!

MonkeyDo:
I thought about this a bit more, and did some testing.

First of all, my concerns over what other components it could affect were unfounded, so that's good !

Second, I think you can actually revert to a single `match` query where you used a multi_match, but change the indexing so that `'aliasSet.aliases'` is indexed instead of `defaultAlias`.
The reason for that is that the defaultAlias is already on of the aliases contained in aliasSet.aliases.

panwarabhishek345:
@MonkeyDo That's great. If it is then yes, we can use single match query. And what about the "creating a new entity" page?

MonkeyDo:
That uses another mechanism to do the search on the database, and is unaffected by the changes in this PR.


I did find an important bug however, when testing the search component in the relationship editor.
Since you're working on search functionality, I figured you might be interested in having a go at it: https://tickets.metabrainz.org/browse/BB-361
It's another component entirely, but uses the results of the work you did in this PR

panwarabhishek345:
@MonkeyDo  Made the appropriate changes to the commit. Now only aliasSet.aliases is indexed and queried.

MonkeyDo:
OK, I've done more testing, and I am satisfied this PR works nicely with no undesirable side-effects !

Thank you !

panwarabhishek345:
Perfect!

PR #315, opened 2019-11-06
fix(profile): Fix AchievementsTab Badges on editor page [BB-360]

sbvkrishna:
### Problem

**Issue [BB-360](https://tickets.metabrainz.org/browse/BB-360)**: AchievementsTab Badges on Editor page doesn't work

### Solution

Add **isOwner** boolean to check the owner.

### Areas of Impact

- src/client/components/pages/parts/editor-achievements.js
- src/client/controllers/editor/achievement.js
- src/server/routes/editor.js

### DIFF
--- a/src/client/components/pages/parts/editor-achievements.js
+++ b/src/client/components/pages/parts/editor-achievements.js
@@ -83,7 +83,7 @@ class EditorAchievementTab extends React.Component {
 		const locked = this.renderAchievements(false);
 
 		let rankUpdate;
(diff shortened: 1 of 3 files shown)


sbvkrishna:
@MonkeyDo Actually it gives an error-500 if we try to visit **Editor Achievements** (example) when user is not logged in, because req.user is [undefined](https://github.com/bookbrainz/bookbrainz-site/pull/315/files#diff-35d9e14e00b021896f2e3afddf7018d5R302)

PR #316, opened 2019-11-06
fix(profile,achievement): fix accessing req.user.id when user not logged-in

sbvkrishna:
### Problem

Visiting Editor Achievements page gives 500 error when user is not logged-in (commented [here](https://github.com/bookbrainz/bookbrainz-site/pull/315#issuecomment-550389833))
### Solution

Check if req.user is undefined, then access req.user.id
**Verified it works now.**
### Areas of Impact

- src/server/routes/editor.js


MonkeyDo:
Good catch !

sbvkrishna:
Thanks for merging. Actually, it was my bad. Will try to avoid this next time.

PR #322, opened 2019-11-18
fix: Reassign alias language id safely

MonkeyDo:
Fix issue with language property being reassigned by object destructuring (introduced in PR #308 ).
Also refactor other entities to use the same protection (use languageID instead of language.id in case language is undefined)

### DIFF
--- a/src/server/routes/entity/author.js
+++ b/src/server/routes/entity/author.js
@@ -105,10 +105,11 @@ router.get(
 );
 
 function authorToFormState(author) {
+	/** The front-end expects a language id rather than the language object. */
 	const aliases = author.aliasSet ?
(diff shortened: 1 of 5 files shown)


MonkeyDo:
@panwarabhishek345 For your information, I found a bug pr #308 that I hadn't seen coming !
Basically, with `{language: languageID, …rest}`, we assign language, but if language is present in `rest`, it gets reassigned by `…rest`. 
So instead, we should destructure rest first, and only after that, reassign language.

panwarabhishek345:
True, I should have taken care of that.

MonkeyDo:
I didn't see it at all, neither during review nor testing, which means I didn't test well enough on my side !

It's a sneaky one, hard to see when using object destructuring !

PR #324, opened 2019-11-19
fix(editor): Client side filtering of autocomplete suggestions disabled in Async.

panwarabhishek345:
### Problem
#BB-361 Entity search in relationship editor only shows 2 items

### Solution
Added filterOptions={false} prop to the ImmutableAsyncSelect to disable client side filtering of autosuggested options.

### Areas of Impact
src/client/entity-editor/common/entity-search-field-option.js


MonkeyDo:
Great job finding this , and thanks for the PR !

panwarabhishek345:
@MonkeyDo Yeah, it was a tough find. But the result is worth the efforts :)

MonkeyDo:
Agreed! That's an elegant fix.
Combined with your previous PR for search by aliases, it's a really important improvement!

panwarabhishek345:
Thanks✌️

PR #326, opened 2019-11-25
feat(configuration): allow overwriting config file path | BB-366

sbvkrishna:
### Problem

[BB-366](https://tickets.metabrainz.org/browse/BB-366) Allow overwriting config file path when running local server 

### Solution

Added functionality to pass config file path with `--config` flag.
`npm start -- --config ./config/config.local.json`

### Areas of Impact

src/common/helpers/config.js

### DIFF
--- a/src/common/helpers/config.js
+++ b/src/common/helpers/config.js
@@ -16,19 +16,54 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import {readFileSync} from 'fs';
+import {existsSync, readFileSync} from 'fs';
+import {normalize} from 'path';
 
 
-/* Pull in environment-specific configuration. */
-const env = process.env.NODE_ENV || 'development'; // eslint-disable-line no-process-env,max-len
+let configContents;
 
-let configFileBasename = 'config';
-if (env === 'test') {
-	configFileBasename = 'test';
+// Use Default configuration (for Docker)
+function useDefaultConfig() {
+	/* Pull in environment-specific configuration. */
+	const env = process.env.NODE_ENV || 'development'; // eslint-disable-line no-process-env,max-len
+
+	let configFileBasename = 'config';
+	if (env === 'test') {
+		configFileBasename = 'test';
+	}
+	configContents =
+			readFileSync(`config/${configFileBasename}.json`);
 }
 
-const configContents =
-	readFileSync(`config/${configFileBasename}.json`);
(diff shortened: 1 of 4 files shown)


src/common/helpers/config.js
```diff
@@ -16,19 +16,47 @@
...
+function checkConfigOverwrite() {
+	const args = process.argv;
+	const configIndex = args.indexOf('--config');
+
+	// Check for '--config' followed by 'configPathFile'
+	if (configIndex !== -1 && configIndex < args.length - 1) {
+		let configFilePath = args[configIndex + 1];
+		configFilePath = normalize(configFilePath);
+		if (existsSync(configFilePath)) {
+			configContents = readFileSync(configFilePath);
```
MonkeyDo: The `readFileSync` method can throw an error.
Therefore it should be in a try…catch block, either:
- just for this line
- for the whole `checkConfigOverwrite` method or
- line 58 where you call the method.

src/common/helpers/config.js
```diff
@@ -16,19 +16,47 @@
...
 }
 
-const configContents =
-	readFileSync(`config/${configFileBasename}.json`);
+function checkConfigOverwrite() {
+	const args = process.argv;
+	const configIndex = args.indexOf('--config');
+
+	// Check for '--config' followed by 'configPathFile'
+	if (configIndex !== -1 && configIndex < args.length - 1) {
```
MonkeyDo: I'm not sure I understand the reason for that  `configIndex < args.length - 1` check.
Could you explain that to me?

MonkeyDo: Ah, I think I get it now. Is it to catch an empty `--config` flag, meaning `--config` _not_ followed by a path?

sbvkrishna: yes

src/common/helpers/config.js
sbvkrishna: Yeah true! As I'm checking if the path exists using `existsSync(configFilePath)`, I thought it wasn't needed, but it's actually required since `existsSync` gives true for folders too, so `readFileSync` fails.

MonkeyDo: Correct!
There could also be an unexpected error while reading the file (due to an operating system error or some other surprise for example), and we don't want to possibly prevent the application from launching.

sbvkrishna:
Thank you for looking into this :) `try catch`ing the entire block would be good imo.
Will update the PR soon!

PR #328, opened 2019-12-02
DatePicker fix

ruinername:
### Problem

DatePicker disabled by default when date-input-field is empty

### Solution

isCommonEraDate value checking fix, so that it evaluates to 'true' for both '1' & '0'

### Areas of Impact

Frontend: DatePickers


sbvkrishna:
Related to [ticket BB-389](https://tickets.metabrainz.org/browse/BB-389)

PR #331, opened 2019-12-05
fix : removing "undefined" text at the bottom of error pages

mutsunokamidog:
### Problem

The word "undefined" appears on error pages.

### Solution

Checks variables to make sure that "undefined" does not appear on error pages.

### DIFF
--- a/src/server/templates/target.js
+++ b/src/server/templates/target.js
@@ -48,19 +48,19 @@ export default ({
 }) => {
 	const pageScript =
 		`<script id='page' type='application/json'>${page}</script>`;
-	var scriptTags = null;
-	if (props !== null && typeof props !== "undefined" && script !== null && typeof script !== "undefined"){
+	let scriptTags = null;
+	if (props !== null && typeof props !== 'undefined' && script !== null && typeof script !== 'undefined') {
 		scriptTags = `<script id='props' type='application/json'> ${props}</script>
-		<script src='${script}'></script>`
+		<script src='${script}'></script>`;
 	}
-	return (`
+	return `
 	<!DOCTYPE html>
 	<html>
 		<head>
 			<title>${title ? `${title} – BookBrainz` :
-		'BookBrainz – The Open Book Database'}</title>
-		${!dev ?
-		`<link rel='stylesheet' href='/stylesheets/style.css' />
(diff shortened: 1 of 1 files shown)


src/server/templates/target.js
```diff
@@ -69,7 +69,7 @@ export default ({
 			<div id='target'>${markup}</div>
 			<script src='/js/bundle.js'></script>
 			${page ? pageScript : ''}
-			${props && script &&
+			${props !== null && script !== null ? /** checking if the variables are undefined or null */
```
MonkeyDo: Good initiative to document the code !
I think our automated tools will complain and require the comment line to be on a separate line above the code.

Also, your comment says you're checking for  "undefined or null", but the code only checks for null, and not undefined, as they are different in javascript.
You'll need to check for both null and undefined: `props !== null && typeof props !== "undefined" && …`

This code is starting to be a bit verbose at this point, so I would move the logic before the `return` statement on line 51.
Something like (in pseudo-code):
```
let scriptTags = null;
if(props !== null && typeof props !== "undefined" && MY OTHER CONDITIONS){
  scriptTags = `<script .......`
}
return (
    .....
    {scriptTags}
  </body>
  .....
```
We instantiate `scriptTags` as `null`, so if the logic in the if block is not executed, it will stay as `null`.
React will interpret that correctly and simply not display anything !

mutsunokamidog: Hi @MonkeyDo , 

Thank you so much for your help!! :DD

src/server/templates/target.js
```diff
@@ -69,9 +74,7 @@ export default ({
 			<div id='target'>${markup}</div>
 			<script src='/js/bundle.js'></script>
 			${page ? pageScript : ''}
-			${props && script &&
-				`<script id='props' type='application/json'> ${props}</script>
-				<script src='${script}'></script>`}
+			${scriptTags}
```
sbvkrishna: string `The value is ${null}` is actually interpolated as string "The value is null"
So, we need to add a check condition to like done with `${page ? pageScript : ''}` in the previous line.
But I think there's a better solution as which I noted above :)

mutsunokamidog: Ohh ok I see now, thanks!

PR #333, opened 2019-12-12
feat: Entity merging tool

MonkeyDo:
### Problem
Plenty of duplicates in the database, and no way to merge them…

### Solution
This implements merging routes to allow users to add entities to a merge queue.
merge-4

Once submitted, the user is redirected to a merging page with dropdowns to select which options to keep.
If the value is the same for all selected entities, it shows that single option (dropdown disabled).
If corrections need to be made, the user needs to modify the entity after merging.
merge-5

Merge revisions are show in a special way, and have a merge icon in front to denote them:

merge-6

merge-7

### Areas of Impact
- new /merge/… routes
- database schema changes (for revisions)
- small modifications of the existing entity create/edit process to account for merges.
- new merge page and merge field elements
- refactoring of some common utilities, trying to reuse as much of the existing entity save mechanism as possible
- uses new version of bookbrainz-data


src/client/components/pages/entities/links.js
```diff
@@ -42,8 +42,8 @@ function EntityLinks({entity, identifierTypes, urlPrefix}) {
 			</Col>
 			<Col md={4}>
 				<EntityIdentifiers
-					identifierSet={entity.identifierSet}
 					identifierTypes={identifierTypes}
+					identifiers={entity.identifierSet && entity.identifierSet.identifiers || []}
```
sbvkrishna: Empty array `[]` we are passing here actually evaluates to `true` (Reference MDN [Truthy values](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), [Falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)) insteading of the expected `false`, failing to render Entities _with no identifiers_.
https://github.com/bookbrainz/bookbrainz-site/blob/82781824c8b5b3abde7c93e156b5f63ec3cf190d/src/client/components/pages/entities/identifiers.js#L29-L30

endurance21:
@MonkeyDo
Is there something I can work upon, here?.

MonkeyDo:
> @MonkeyDo
> Is there something I can work upon, here?.

This one is a pretty complex PR to jump into like that, but there is something you can do to help: testing ! :D
I published this PR on test.bookbrainz.org (the database is separate from the main website, just for testing).
If you could try out the merging tool that would be very helpful; you can either find duplicates by searching randomly (there's currently no tool or page to show duplicates), or create new fake entities to merge.
In particular, any scenario that involves relationships and links to other entities is preferable (I tested simple merges quite extensively already, but who knows what possibility I haven't thought of…)
Anything that looks out of place, broken, complicated to understand or could be improved, please create a ticket for :)

endurance21:
@MonkeyDo
Okay. I will report as soon as I wrap up my work!.

PR #339, opened 2020-01-10
Feature-revisionsPage

prabalsingh24:
### Problem

https://tickets.metabrainz.org/browse/BB-388
Adding a new revisionsPage

### Solution

I've created a new route '/revisions' which displays all the recent activities.

Screenshot from 2020-01-13 19-09-23

updated revision page

Screenshot from 2020-01-17 18-42-59

### DIFF
--- a/scripts/build-client-js.sh
+++ b/scripts/build-client-js.sh
@@ -10,7 +10,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		deletion.js \
 		index.js \
 		registrationDetails.js \
-		revision.js \
+		revisionsListPage.js \
 		search.js \
 		statistics.js \
 	-p [ factor-bundle \
@@ -22,11 +22,11 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		-o ../../../static/js/deletion.js \
 		-o ../../../static/js/index.js \
 		-o ../../../static/js/registrationDetails.js \
-		-o ../../../static/js/revision.js \
+		-o ../../../static/js/revisionsListPage.js \
 		-o ../../../static/js/search.js \
 		-o ../../../static/js/statistics.js \
 	] > ../../../static/js/bundle.js
-	
+
 	uglifyjs -cm -- ../../../static/js/entity-editor.js  	  | gzip --best > ../../../static/js/entity-editor.js.gz
 	uglifyjs -cm -- ../../../static/js/editor/edit.js 	  | gzip --best > ../../../static/js/editor/edit.js.gz
 	uglifyjs -cm -- ../../../static/js/editor/achievement.js  | gzip --best > ../../../static/js/editor/achievement.js.gz
@@ -35,7 +35,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 	uglifyjs -cm -- ../../../static/js/deletion.js 		  | gzip --best > ../../../static/js/deletion.js.gz
 	uglifyjs -cm -- ../../../static/js/index.js 		  | gzip --best > ../../../static/js/index.js.gz
 	uglifyjs -cm -- ../../../static/js/registrationDetails.js | gzip --best > ../../../static/js/registrationDetails.js.gz
-	uglifyjs -cm -- ../../../static/js/revision.js 		  | gzip --best > ../../../static/js/revision.js.gz
+	uglifyjs -cm -- ../../../static/js/revisionsListPage.js 		  | gzip --best > ../../../static/js/revisionsListPage.js.gz
 	uglifyjs -cm -- ../../../static/js/search.js 		  | gzip --best > ../../../static/js/search.js.gz
 	uglifyjs -cm -- ../../../static/js/bundle.js     | gzip --best > ../../../static/js/bundle.js.gz
 popd
--- a/scripts/build-client-js.sh
+++ b/scripts/build-client-js.sh
@@ -11,7 +11,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		index.js \
 		registrationDetails.js \
 		revision.js \
-		revisionsListPage.js \
+		revisions.js \
 		search.js \
 		statistics.js \
 	-p [ factor-bundle \
@@ -24,7 +24,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		-o ../../../static/js/index.js \
 		-o ../../../static/js/registrationDetails.js \
 		-o ../../../static/js/revision.js \
-		-o ../../../static/js/revisionsListPage.js \
+		-o ../../../static/js/revisions.js \
 		-o ../../../static/js/search.js \
 		-o ../../../static/js/statistics.js \
 	] > ../../../static/js/bundle.js
@@ -38,7 +38,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 	uglifyjs -cm -- ../../../static/js/index.js 		  | gzip --best > ../../../static/js/index.js.gz
 	uglifyjs -cm -- ../../../static/js/registrationDetails.js | gzip --best > ../../../static/js/registrationDetails.js.gz
 	uglifyjs -cm -- ../../../static/js/revision.js 		  | gzip --best > ../../../static/js/revision.js.gz
-	uglifyjs -cm -- ../../../static/js/revisionsListPage.js 		  | gzip --best > ../../../static/js/revisionsListPage.js.gz
+	uglifyjs -cm -- ../../../static/js/revisions.js 		  | gzip --best > ../../../static/js/revisions.js.gz
 	uglifyjs -cm -- ../../../static/js/search.js 		  | gzip --best > ../../../static/js/search.js.gz
 	uglifyjs -cm -- ../../../static/js/bundle.js     | gzip --best > ../../../static/js/bundle.js.gz
 popd
--- a/scripts/build-client-js.sh
+++ b/scripts/build-client-js.sh
@@ -10,6 +10,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		deletion.js \
 		index.js \
 		registrationDetails.js \
+		revision.js \
 		revisionsListPage.js \
 		search.js \
 		statistics.js \
@@ -22,6 +23,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		-o ../../../static/js/deletion.js \
 		-o ../../../static/js/index.js \
 		-o ../../../static/js/registrationDetails.js \
+		-o ../../../static/js/revision.js \
 		-o ../../../static/js/revisionsListPage.js \
 		-o ../../../static/js/search.js \
 		-o ../../../static/js/statistics.js \
@@ -35,6 +37,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 	uglifyjs -cm -- ../../../static/js/deletion.js 		  | gzip --best > ../../../static/js/deletion.js.gz
 	uglifyjs -cm -- ../../../static/js/index.js 		  | gzip --best > ../../../static/js/index.js.gz
 	uglifyjs -cm -- ../../../static/js/registrationDetails.js | gzip --best > ../../../static/js/registrationDetails.js.gz
+	uglifyjs -cm -- ../../../static/js/revision.js 		  | gzip --best > ../../../static/js/revision.js.gz
 	uglifyjs -cm -- ../../../static/js/revisionsListPage.js 		  | gzip --best > ../../../static/js/revisionsListPage.js.gz
 	uglifyjs -cm -- ../../../static/js/search.js 		  | gzip --best > ../../../static/js/search.js.gz
 	uglifyjs -cm -- ../../../static/js/bundle.js     | gzip --best > ../../../static/js/bundle.js.gz
--- a/scripts/build-client-js.sh
+++ b/scripts/build-client-js.sh
@@ -7,6 +7,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		editor/achievement.js \
 		editor/editor.js \
 		entity/entity.js \
+		error.js \
 		deletion.js \
 		index.js \
 		registrationDetails.js \
@@ -20,6 +21,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		-o ../../../static/js/editor/achievement.js \
 		-o ../../../static/js/editor/editor.js \
 		-o ../../../static/js/entity/entity.js \
+		-o ../../../static/js/error.js \
 		-o ../../../static/js/deletion.js \
 		-o ../../../static/js/index.js \
 		-o ../../../static/js/registrationDetails.js \
@@ -34,6 +36,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 	uglifyjs -cm -- ../../../static/js/editor/achievement.js  | gzip --best > ../../../static/js/editor/achievement.js.gz
 	uglifyjs -cm -- ../../../static/js/editor/editor.js 	  | gzip --best > ../../../static/js/editor/editor.js.gz
 	uglifyjs -cm -- ../../../static/js/entity/entity.js	  | gzip --best > ../../../static/js/entity/entity.gz
+	uglifyjs -cm -- ../../../static/js/error.js	  | gzip --best > ../../../static/js/error.js.gz
 	uglifyjs -cm -- ../../../static/js/deletion.js 		  | gzip --best > ../../../static/js/deletion.js.gz
 	uglifyjs -cm -- ../../../static/js/index.js 		  | gzip --best > ../../../static/js/index.js.gz
 	uglifyjs -cm -- ../../../static/js/registrationDetails.js | gzip --best > ../../../static/js/registrationDetails.js.gz
--- a/src/client/components/pages/index.js
+++ b/src/client/components/pages/index.js
@@ -27,7 +27,6 @@ import {genEntityIconHTMLElement, getEntityLabel} from '../../helpers/entity';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
 import React from 'react';
-import {isNil as _isNil} from 'lodash';
 
 
 const {Alert, Button, Col, Grid, ListGroup, ListGroupItem, Row} = bootstrap;
@@ -230,11 +229,14 @@ class IndexPage extends React.Component {
 										>
 											<Row>
 												<Col md={2}>{`r${entity.revisionId}`}</Col>
-												<Col md={6}>
+												<Col md={5} >
 													{genEntityIconHTMLElement(entity.type)}
 													{getEntityLabel(entity)}
 												</Col>
-												<Col md={4}>
+												<Col md={3}>
+													{entity.editor.name}
+												</Col>
+												<Col md={2}>
 													{formatDate(new Date(entity.createdAt))}
 												</Col>
 											</Row>
--- a/src/client/components/pages/index.js
+++ b/src/client/components/pages/index.js
@@ -166,7 +166,7 @@ class IndexPage extends React.Component {
 	}
 
 	renderContent() {
-		const {recent} = this.props;
+		const {results} = this.props;
 		const disableSignUp = this.props.disableSignUp ? {disabled: true} : {};
 
 		return (
@@ -216,14 +216,14 @@ class IndexPage extends React.Component {
 						Register!
 					</Button>
 				</div>
-				{recent &&
+				{results &&
 					<div>
 						<hr/>
 						<Row>
 							<Col md={12}>
 								<h2 className="text-center">Recent Activity</h2>
 								<ListGroup>
-									{recent.map((entity) => (
+									{results.map((entity) => (
 										<ListGroupItem
 											href={`/revision/${entity.revisionId}`}
 											key={entity.revisionId}
@@ -262,7 +262,7 @@ class IndexPage extends React.Component {
 IndexPage.displayName = 'IndexPage';
 IndexPage.propTypes = {
 	disableSignUp: PropTypes.bool,
-	recent: PropTypes.array.isRequired
+	results: PropTypes.array.isRequired
 };
 IndexPage.defaultProps = {
 	disableSignUp: false
--- a/src/client/components/pages/index.js
+++ b/src/client/components/pages/index.js
@@ -213,10 +213,19 @@ class IndexPage extends React.Component {
 					</Button>
 				</div>
 				<RevisionsTable
-					results={this.props.results}
+					results={this.props.recent}
 				/>
 				<div>
-					<a className="text-right col-xs-12" href="/revisions">See all revisions</a>
+					<Row>
+						<Col mdOffset={10}>
+							<Button
+								bsStyle="primary"
+								href="/revisions"
+							>
+								See all revisions
+							</Button>
+						</Col>
+					</Row>
 				</div>
 			</Grid>
 		);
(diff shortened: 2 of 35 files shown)


prabalsingh24:
I haven't add a link -see all revisions - in index.js yet. You'll have to go to /revisions to see the new page. 
Also I think it shouldn't be named /revisions as there is already a route /revision present and these two names are very similar.

src/client/components/pages/revisionsListPage.js
```diff
@@ -0,0 +1,134 @@
...
+import * as bootstrap from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import React from 'react';
+import RevisionsTable from './parts/revisions-diplayTable';
+import request from 'superagent-bluebird-promise';
+
+
+const {Pager, Button, ButtonGroup, DropdownButton, MenuItem} = bootstrap;
+
+class RevisionsListPage extends React.Component {
```
MonkeyDo: Let's replace wherever you have 'RevisionsList' everywhere in the various files with 'Revisions'  (including 'RevisionsListPage' -> 'RevisionsPage')

scripts/build-client-js.sh
```diff
@@ -11,6 +11,7 @@ cross-env BABEL_ENV="browser" browserify -t [babelify] \
 		index.js \
 		registrationDetails.js \
 		revision.js \
+		revisionsListPage.js \
```
MonkeyDo: This file needs to be renamed: revisionsListPage.js -> revisions.js, and so needs changing in a few places in this file

src/client/components/pages/parts/revisions-diplayTable.js
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (C) 2015  Ohm Patel
+ *               2016  Sean Burke
```
MonkeyDo: This should have your name instead :)

src/client/components/pages/parts/revisions-diplayTable.js
```diff
@@ -0,0 +1,82 @@
...
+
+function RevisionsTable(props) {
+	const {results} = props;
+	const tableCssClasses = 'table table-striped';
+
+	return (
+
+		<div>
+			<div>
+				<h1>Recent Activities</h1>
```
MonkeyDo: Activities -> Activity

src/client/controllers/revisionsListPage.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+const propsTarget = document.getElementById('props');
+const props = propsTarget ? JSON.parse(propsTarget.innerHTML) : {};
+const markup = (
+	<AppContainer>
+		<Layout {...extractLayoutProps(props)}>
+			<RevisionsListPage
+				from={props.from}
+				results={props.recent}
```
MonkeyDo: Rename 'recent' to 'results'

src/server/routes/revisionsList.js
```diff
@@ -0,0 +1,145 @@
...
+/* GET home page. */
+router.get('/', async (req, res, next) => {
+	const {orm} = req.app.locals;
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+
+	function render(entities) {
+		const props = generateProps(req, res, {
+			from,
+			recent: _.take(entities, size),
```
MonkeyDo: Rename 'recent' to 'results'

src/server/routes/revisionsList.js
```diff
@@ -0,0 +1,145 @@
...
+		 * Renders react components server side and injects markup into target
+		 * file object spread injects the app.locals variables into React as
+		 * props
+		 */
+		const markup = ReactDOMServer.renderToString(
+			<Layout {...propHelpers.extractLayoutProps(props)}>
+				<RevisionsListPage
+					// disableSignUp={req.signUpDisabled}
+					from={props.from}
+					results={props.recent}
```
MonkeyDo: Rename 'recent' to 'results'

src/server/routes/revisionsList.js
```diff
@@ -0,0 +1,145 @@
...
+
+router.get('/revisions', async (req, res, next) => {
+	const {orm} = req.app.locals;
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+
+	function render(entities) {
+		const props = generateProps(req, res, {
+			from,
+			recent: _.take(entities, size),
```
MonkeyDo: Rename 'recent' to 'results'

src/server/routes/revisionsList.js
```diff
@@ -0,0 +1,145 @@
...
+			size
+		});
+		res.send(props.recent);
+	}
+
+	const entityModels = utils.getEntityModels(orm);
+
+	try {
+		const orderedEntities = await utils.getOrderedEntities(from, size, entityModels, orm);
+		return render(orderedEntities);
```
MonkeyDo: Here you can do `res.send(_.take(orderedEntities,size);` instead (no need for `return`in this case), and remove the `render` function

src/client/components/pages/revisionsListPage.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import React from 'react';
+import RevisionsTable from './parts/revisions-diplayTable';
```
MonkeyDo: Lets keep the naming consistent and rename this file to revisions-table.js

src/server/helpers/utils.js
```diff
@@ -50,10 +50,68 @@ export function getEntityModels(orm: Object): Object {
...
+						revision.created_at AS created_at
+					FROM bookbrainz.${SQLViewName} AS entity
+					JOIN bookbrainz.revision ON revision.id = entity.revision_id
+					LEFT JOIN bookbrainz.alias ON alias.id = entity.default_alias_id
+					LEFT JOIN bookbrainz.revision_parent ON revision_parent.child_id = entity.revision_id AND entity.default_alias_id IS NULL
+					LEFT JOIN bookbrainz.${SQLViewName} AS parent ON parent.revision_id = revision_parent.parent_id AND entity.default_alias_id IS NULL
+					LEFT JOIN bookbrainz.alias as parent_alias ON parent_alias.id = parent.default_alias_id AND entity.default_alias_id IS NULL
+					WHERE entity.master = true
+					ORDER BY revision.created_at DESC
+					LIMIT ${size + from};`)
```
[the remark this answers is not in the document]
prabalsingh24: We discussed this in the irc but later I realised that limit x offset y won't work.
We're getting revisions from 5 different entityModels- (work,edition,etc), and then we're sorting them.
Let's say we want to get revisions from 20 to 40. We're not sure if revisions that are supposed to be in this range will lie in 20 to 40 range of each entityModels. They can be anywhere between 1 to 40

sbvkrishna:
Also, in case you forgot adding styles to `RevisionsPage`, similar to [`SearchPage styles`](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/stylesheets/style.less#L337-L371) :)

prabalsingh24:
> Also, in case you forgot adding styles to `RevisionsPage`, similar to [`SearchPage styles`](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/stylesheets/style.less#L337-L371) :)

prabalsingh24:
> Also, in case you forgot adding styles to `RevisionsPage`, similar to [`SearchPage styles`](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/stylesheets/style.less#L337-L371) :)

Thanks

MonkeyDo:
@prabalsingh24 : No need to close the PR :) You can just reopen the PR and keep pushing commits to it.

prabalsingh24:
Things left to do.
1.  Add a link - show all revisions - in the index page
2. What do you think of slice method while calculating orderedRevisions? Check my comment above. 
3. Also I think we can have those two function - getOrdredRevisons and addEditorToOrderedRevisions in separate file - /helper/revisions. Right now they are in /helper/utilis

src/server/helpers/utils.js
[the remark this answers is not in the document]
MonkeyDo: The problem is that it won't scale well like that.

What I propose is something like this: (for the revisions page, leaving the index as it was before)

1. Get the latest revisions using limit and offset and transform it into an object. Let's call this the revisions object
2. map that to get an array of revision ids
3. Get all author_revisions whose id is in the array of revision ids (with related entity)
4. For each author_revision id, add the related entity to an 'entities' array in the revisions object by corresponding id
5. Repeat steps 3 and 4 for each entity type

You'll now have and object with revisions with an entities array with the related entity.
You can display each revision in a table row, with the list of related entities.

What do you think?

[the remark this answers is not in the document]
prabalsingh24: schema

This is what you mean right? First we get revisions , then  store corresponding editor_name,editor_id, then find the corresponding entity, (eg. publisher in the picture)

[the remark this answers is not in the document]
prabalsingh24: This call can be used https://bookshelfjs.org/api.html#Model-instance-fetchPage to fetch revisions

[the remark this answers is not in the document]
prabalsingh24: I couldn't use fetchPage. I had to import pagination plugin in the bookbrainz-data in node_modules. 
I created PR in bookbrainz-data for it.

prabalsingh24:
https://github.com/bookbrainz/bookbrainz-site/pull/339#discussion_r368187668

src/server/helpers/utils.js
```diff
@@ -49,75 +49,75 @@ export function getEntityModels(orm: Object): Object {
...
+			withRelated: [
+				'author'
+			]
+		});
+	const revisions = await revisionPromise;
+	const orderedRevisions = [];
+	for (let i = 0; i < revisions.toJSON().length; i++) {
 		// eslint-disable-next-line no-await-in-loop
-		await addEditorToOrderedRevisions(orderedRevisions[i], orm.Revision);
+		const temp = await getCompleteRevision(revisions.toJSON()[i], orm);
```
prabalsingh24: I forgot to change the temp variable name :P

prabalsingh24:
ERROR TypeError: (intermediate value).orderBy(...).fetchPage is not a function
The test failed because of this. 
https://github.com/bookbrainz/bookbrainz-data-js/pull/213
pagination plugin is need to use fetchPage

prabalsingh24:
Also currently, the deleted revision's name isn't showing properly, I have to add parentAlias for that. I'll do that. 
Meanwhile what do you think of utlis.getOrderedRevisions function

src/server/helpers/utils.js
```diff
@@ -49,11 +49,119 @@ export function getEntityModels(orm: Object): Object {
...
+		EditionGroupRevision,
+		EditionRevision,
+		PublisherRevision,
+		WorkRevision
+	];
+}
+async function getCompleteRevision(revision, orm) {
+	const {Entity, AliasSet} = orm;
+	const RevisionModels = getRevisionModels(orm);
+	for (let i = 0; i < RevisionModels.length; i++) {
```
MonkeyDo: OK, I tried a few things to end up with a pretty efficient way to fetch everything we need.

At the moment, you'll get a returned object that doesn't have the correct structure. I'll let you restructure that accordingly, I have to run away from my screen :)

```
for (let i = 0; i < RevisionModels.length; i++) {
	const Model = RevisionModels[i];
	const model = await new Model({id: revision.id})
		.fetch({
			require: false,
			withRelated: [
				'data.aliasSet.defaultAlias',
				'entity'
			]
		});

	if (model) {
		const modelJSON = model.toJSON();
		if (!modelJSON.dataId) {
			const bbid = modelJSON.bbid;
			// This feels a bit hacky, but shouldn't break: get e.g. 'edition_group' from 'bookbrainz.edition_group_revision' table name
			const modelType = model.tableName.match(/^bookbrainz.(\w+)_revision$/i)[1];
			const databaseModelName = _.upperFirst(_.camelCase(modelType));
			modelJSON.data.deleted = true;
			const parentAlias = await orm.func.entity.getEntityParentAlias(
				orm, databaseModelName, bbid
			);
			modelJSON.data.parentAlias = parentAlias;
		}

		const revisionId = revision.id;
		revision.editor = revision.author;
		delete revision.author;

		return {
			revisionId,
			...revision,
			...modelJSON
		};
	}
}
```

Note the withRelated: 'data.aliasSet.defaultAlias' ;  since we only need to load the alias, this nice shortcut using the ORM relationships saves us some hassle :)
See also the use of `orm.func.entity.getEntityParentAlias`

src/server/routes/index.js
```diff
@@ -45,11 +45,11 @@ router.get('/', async (req, res, next) => {
...
 
-	function render(entities) {
+	function render(results) {
 		const props = generateProps(req, res, {
 			homepage: true,
-			recent: _.take(entities, numRevisionsOnHomepage),
-			requireJS: Boolean(res.locals.user)
+			requireJS: Boolean(res.locals.user),
+			results: _.take(results, numRevisionsOnHomepage)
 		});
```
MonkeyDo: I still think that call to generateProps function is not useful. I think you only end up needing props.results, which is just as good as the original results array.
In any case, the rest of the props are incorrect: homepage shouldn't be true and I don't think you need requireJS.

src/server/helpers/utils.js
prabalsingh24: I was try to use that getEntityParentAlias using import method but that didn't work. It worked but was showing some error, so I just copy pasted the function in the file for. I will change it

prabalsingh24: orm.func.entity.getEntityParentAlias
I forgot I could use it this :P my bad

src/server/routes/index.js
prabalsingh24: I didn't change this part, it was like this before also. I just renamed it from recent to result

prabalsingh24:
Thanks for the review. I will make the appropriate changes :)

src/server/helpers/utils.js
prabalsingh24: Thanks sir. Yeah this is much better approach. I will restructure it accordingly.

MonkeyDo: Oh, no need for honorifics !
I am no sir but a simple monkey :)

src/server/routes/index.js
MonkeyDo: I think you can just remove the `homepage` and `requireJS` properties.

prabalsingh24: If homepage is not true, then it shows search bar in the page. And for index page we don't want that because there is already a search bar in it

src/server/helpers/utils.js
```diff
@@ -49,11 +50,80 @@ export function getEntityModels(orm: Object): Object {
...
+			.fetch({
+				withRelated: [
+					'data.aliasSet.defaultAlias'
+				]
+			});
+		if (model) {
+			const modelJSON = model.toJSON();
+			// eslint-disable-next-line no-await-in-loop
+			const entity = await new Entity({bbid: modelJSON.bbid}).fetch();
+			const entityJSON = entity ? entity.toJSON() : {};
```
prabalsingh24: I used 78, 79 instead of  `const modelType = model.tableName.match(/^bookbrainz.(\w+)_revision$/i)[1];
			                const databaseModelName = _.upperFirst(_.camelCase(modelType));`

This seems less hacky to me.

src/server/routes/index.js
```diff
@@ -45,11 +45,10 @@ router.get('/', async (req, res, next) => {
 	const {orm} = req.app.locals;
 	const numRevisionsOnHomepage = 9;
 
-	function render(entities) {
+	function render(results) {
 		const props = generateProps(req, res, {
 			homepage: true,
```
prabalsingh24: If I remove the homepage, it changes the UI, It determines if there should be searchbox in navbar or not.

src/server/helpers/utils.js
```diff
@@ -90,71 +60,63 @@ function getRevisionModels(orm) {
 		WorkRevision
 	];
 }
-async function getCompleteRevision(revision, orm) {
-	const {Entity, AliasSet} = orm;
+async function 	getCompleteRevision(revision, orm) {
```
MonkeyDo: So this was starting to work quite well, but with 100 results per page loading the next page was quite slow.
Looking at the way we're constructing the complete revision, we do a lot of requests:
5 (one for each entity type) * number of revisions * 1 or 2 (getEntityParentAlias might be called)

Instead, we can fetch once for each entity type all the entity_revisions that match all our revisions, in one go, and then for each fetch the information we need. As a quick note, even though fetching the Entity model is much cleaner than my table name parsing hack, it does add a request per entity, so it might be worth comparing the speed at which the page loads with both configurations, if we want to further optimize.

Here's what I ended up with:
```
/* eslint-disable no-await-in-loop */
/**
 * Fetches the entities affected by a revision, their alias
 * or in case of deleted entities their last know alias.
 * It then attaches the necessary information to each revisions's entities array
 *
 * @param {array} revisions - the array of revisions
 * @param {object} orm - the BookBrainz ORM, initialized during app setup
 * @returns {array} - The modified revisions array
 */
async function getCompleteRevision(revisions, orm) {
	const revisionIDs = revisions.map(({revisionId}) => revisionId);
	const RevisionModels = getRevisionModels(orm);
	const {Entity} = orm;
	for (let i = 0; i < RevisionModels.length; i++) {
		const EntityRevision = RevisionModels[i];
		const entityRevisions = await new EntityRevision()
			.query((qb) => {
				qb.whereIn('id', revisionIDs);
			})
			.fetchAll({
				require: false,
				withRelated: [
					'data.aliasSet.defaultAlias'
				]
			})
			.catch(EntityRevision.NotFoundError, (err) => {
				// eslint-disable-next-line no-console
				console.log(err);
			});
		if (entityRevisions && entityRevisions.length) {
			const entityRevisionsJSON = entityRevisions.toJSON();
			for (let index = 0; index < entityRevisionsJSON.length; index++) {
				const entityRevision = entityRevisionsJSON[index];
				const entity = await new Entity({bbid: entityRevision.bbid}).fetch();
				const type = entity.get('type');
				const bbid = entity.get('bbid');
				const entityProps = {bbid, type};
				if (entityRevision.data) {
					entityProps.defaultAlias = entityRevision.data.aliasSet.defaultAlias;
				}
				// Fetch the parent alias only if data property is nullish, i.e. it is deleted
				else {
					entityProps.parentAlias = await orm.func.entity.getEntityParentAlias(orm, type, bbid);
				}
				// Find the revision by id and attach the current entity to it
				const revisionIndex = revisions.findIndex(rev => rev.revisionId === entityRevision.id);
				revisions[revisionIndex].entities.push(entityProps);
			}
		}
	}
	return revisions;
}
/* eslint-enable no-await-in-loop */
```

src/client/components/pages/parts/revisions-table.js
```diff
@@ -0,0 +1,88 @@
...
+					<tr>
+						<th className="col-sm-2">Revision ID</th>
+						<th className="col-sm-5">Revision</th>
+						<th className="col-sm-3">Edited By</th>
+						<th className="col-sm-2">Date</th>
+
+					</tr>
+				</thead>
+				<tbody>
+					{
```
MonkeyDo: With my fetching method described in the other comments, the revisions table needs a tweak to display multiple potential entities for each revision.
There's what I have:

```
{
						results.map((revision, i) => (
							<tr key={i}>
								<td>
									<a href={`/revision/${revision.revisionId}`} >
										{revision.revisionId}
									</a>
								</td>
								<td>
									{revision.entities.map(entity => (
										<a href={`/${entity.type}/${entity.bbid}`} key={entity.bbid}>
											{genEntityIconHTMLElement(entity.type)}
											{getEntityLabel(entity)}
										</a>
									))}
								</td>
								<td>
									<a href={`/editor/${revision.editor.id}`} >
										{revision.editor.name}
									</a>
								</td>
								<td>{formatDate(new Date(revision.createdAt))}</td>
							</tr>
						))
					}
```

src/client/components/pages/index.js
```diff
@@ -216,25 +215,28 @@ class IndexPage extends React.Component {
...
 											<Row>
 												<Col md={2}>{`r${entity.revisionId}`}</Col>
-												<Col md={6}>
+												<Col md={5} >
 													{genEntityIconHTMLElement(entity.type)}
 													{getEntityLabel(entity)}
 												</Col>
-												<Col md={4}>
+												<Col md={3}>
+													{entity.editor.name}
```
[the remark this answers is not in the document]
prabalsingh24: Yeah I'll do that, and also I have to add see all revisions link

prabalsingh24:
@MonkeyDo This is how the table will look after making changes you mentioned. I think entities look much better on top of each other rather than a line, but what do you think of the timestamp? Is this better or only date is better?
Screenshot from 2020-01-28 03-24-50

src/client/components/pages/index.js
```diff
@@ -215,38 +212,12 @@ class IndexPage extends React.Component {
...
-								</ListGroup>
-							</Col>
-						</Row>
-					</div>
-				}
+				<RevisionsTable
+					results={this.props.results}
+				/>
+				<div>
+					<a className="text-right col-xs-12" href="/revisions">See all revisions</a>
```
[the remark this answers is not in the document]
prabalsingh24: You forgot to tell me which button you thought of :P

MonkeyDo:
> @MonkeyDo This is how the table will look after making changes you mentioned. I think entities look much better on top of each other rather than a line, but what do you think of the timestamp? Is this better or only date is better?

Looking nice !
I think the timestamp is useful information here.
If there's an outcry about it we can think of hiding it a bit (for example putting the full date-time as a title attribute so that it appears when you hover over the date), but I think this is mighty fine.

src/client/components/pages/index.js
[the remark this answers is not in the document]
MonkeyDo: I meant the link "See all revisions" should be a big button

src/client/components/pages/parts/revisions-table.js
```diff
@@ -0,0 +1,96 @@
...
+							<tr key={i}>
+								<td>
+									<a href={`/revision/${revision.revisionId}`} >
+										{revision.revisionId}
+									</a>
+								</td>
+								<td>
+									{revision.entities.map(entity => (
+										<div key={`${revision.revisionId}-${entity.bbid}`}>
+											<a href={`/${_kebabCase(entity.type)}/${entity.bbid}`} >
```
MonkeyDo: Actually, here you can use the `getEntityLink` utility (from src/server/helpers/utils) to remove some code duplication

src/client/components/pages/index.js
```diff
@@ -262,7 +235,7 @@ class IndexPage extends React.Component {
 IndexPage.displayName = 'IndexPage';
 IndexPage.propTypes = {
 	disableSignUp: PropTypes.bool,
-	recent: PropTypes.array.isRequired
+	results: PropTypes.array.isRequired
```
MonkeyDo: Originally when I asked you to rename 'recent' to 'results', I only had in mind the revisions page, not the index page.
Semantically, it makes a lot more sense to have a 'recent' or 'recentRevisions‘ on the index page.
On the revisions page, it's obvious that we have 'results' of a query (last X revisions, page Y), but on the index page what are they the results of?

This will need changing in a few places, and I now it's an annoying little change, but I keep looking at it and it bugs me :p

prabalsingh24: It's alright I'll change this :)

prabalsingh24:
This is how the table looks now. What do you think?

Screenshot from 2020-01-29 05-33-18

prabalsingh24:
Thanks @MonkeyDo , I learned alot :)

PR #341, opened 2020-01-14
fix(editor): Fix Identifiers validation not working - BB-390

sbvkrishna:
### Problem

[BB-390: Identifiers validation not working](https://tickets.metabrainz.org/browse/BB-390)

### Solution

Pass `identifiersTypes` parameter to `validate()` function

### Areas of Impact
src/client/entity-editor/submission-section/submission-section.js


sbvkrishna:
Thanks :) And yeah it took some time to understand what's going on in there!

PR #342, opened 2020-01-14
fix (editor) : Saving a revision with no changes BB 337

prabalsingh24:
### Problem
https://tickets.metabrainz.org/browse/BB-337
It shouldn't be possible to save a revision with no changes

### DIFF
--- a/src/server/routes/entity/author.js
+++ b/src/server/routes/entity/author.js
@@ -224,11 +224,13 @@ function transformNewForm(data) {
 		aliases,
 		beginAreaId: data.authorSection.beginArea &&
 			data.authorSection.beginArea.id,
-		beginDate: dateObjectToISOString(data.authorSection.beginDate),
+		beginDate: data.authorSection.beginDate ?
+			dateObjectToISOString(data.authorSection.beginDate) : '',
 		disambiguation: data.nameSection.disambiguation,
 		endAreaId: data.authorSection.endArea &&
 			data.authorSection.endArea.id,
-		endDate: data.authorSection.ended ? dateObjectToISOString(data.authorSection.endDate) : '',
+		endDate: data.authorSection.endDate ?
+			dateObjectToISOString(data.authorSection.endDate) : '',
 		ended: data.authorSection.ended,
 		genderId: data.authorSection.gender,
 		identifiers,
--- a/src/server/routes/entity/entity.js
+++ b/src/server/routes/entity/entity.js
@@ -20,10 +20,12 @@
 // @flow
 
 import * as achievement from '../../helpers/achievement';
+import * as error from '../../../common/helpers/error';
 import * as handler from '../../helpers/handler';
 import * as propHelpers from '../../../client/helpers/props';
 import * as search from '../../helpers/search';
 import * as utils from '../../helpers/utils';
+
 
 import type {$Request, $Response, NextFunction} from 'express';
 import type {
@@ -51,6 +53,7 @@ import WorkPage from '../../../client/components/pages/entities/work';
 import _ from 'lodash';
 import config from '../../../common/helpers/config';
 import target from '../../templates/target';
+
 
 type PassportRequest = $Request & {user: any, session: any};
 
@@ -127,6 +130,7 @@ export function displayEntity(req: PassportRequest, res: $Response) {
 							}
 							return unlockName;
 						})
+						// eslint-disable-next-line no-shadow
 						.catch((error) => {
 							log.debug(error);
 						})
@@ -267,8 +271,7 @@ export function handleDelete(
 		// Get the parents of the new revision
 		const revisionParentsPromise = newRevisionPromise
 			.then((revision) =>
-				revision.related('parents').fetch({transacting})
-			);
+				revision.related('parents').fetch({transacting}));
 
 		// Add the previous revision as a parent of this revision.
 		const parentAddedPromise =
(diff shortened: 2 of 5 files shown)


src/server/routes/entity/publisher.js
```diff
@@ -225,13 +224,13 @@ function transformNewForm(data) {
...
-
 	return {
 		aliases,
 		areaId: data.publisherSection.area && data.publisherSection.area.id,
-		beginDate: dateObjectToISOString(data.publisherSection.beginDate),
+		beginDate: data.publisherSection.beginDate ?
+			dateObjectToISOString(data.publisherSection.beginDate) : '',
 		disambiguation: data.nameSection.disambiguation,
-		endDate: data.publisherSection.ended ?
+		endDate: data.publisherSection.endDate ?
```
MonkeyDo: `endDate` is supposed to be an object of type `{day:string, month:string, year:string}`, although it sometimes is `false`. Not sure where this inconsistency comes from
This used to only try to parse and set the endDate if the ended checkbox is checked, which is probably still a useful check

prabalsingh24: There will be a case when ended is true but endDate is null. i.e. when you know a publisher company has dissolved but don't know when.
Earlier the code was if ended is true, pick the endDate, and in some cases endDate was null and was creating some problems.

prabalsingh24: Or we can have 2 checks ,'if ended is true and endDate is available', but I think having check for only endDate works as well

src/server/routes/entity/entity.js
```diff
@@ -482,9 +485,9 @@ async function getNextAnnotation(
 		id && new Annotation({id}).fetch({transacting})
 	);
 
-	return orm.func.annotation.updateAnnotation(
+	return body.annotation ? orm.func.annotation.updateAnnotation(
 		orm, transacting, oldAnnotation, body.annotation, revision
-	);
+	) : Promise.resolve(null);
 }
```
prabalsingh24: I did this because earlier if oldAnnotaion and newAnnotation both are undefined, updateAnnotation was thinking both are same and was just returning oldAnnotation

src/server/routes/entity/publisher.js
MonkeyDo: I agree. The redux state takes care of erasing the endDate if you uncheck "ended", so you shouldn't see any case where ended is false but there is an endDate.
I tested that successfully.

MonkeyDo: There is another underlying issue here. I've seen that endDate can either be a `{day, month, year}` object or a boolean. Not sure what are the conditions for there to be a boolean, but I doubt that it is correct.

src/server/routes/entity/entity.js
MonkeyDo: I think this is absolutely fine and a justified optimization

src/server/routes/entity/entity.js
```diff
@@ -364,6 +366,10 @@ async function processEditionSets(
 	);
 
 	const releaseEvents = _.get(body, 'releaseEvents') || [];
+	if (!releaseEvents[0].areaId) {
+		releaseEvents[0].areaId = null;
+	}
+
```
prabalsingh24: https://github.com/bookbrainz/bookbrainz-data-js/blob/9faa67b5631ea5816828e05680e579d11526515f/src/func/releaseEvent.js#L34
it was returning false when old and new areaId were 'null' and ;undefined.;
I guess better solution would be if we change the bookbrainz-data code accordingly. But this works for now.

src/server/routes/entity/entity.js
```diff
@@ -482,9 +488,9 @@ async function getNextAnnotation(
 		id && new Annotation({id}).fetch({transacting})
 	);
 
-	return orm.func.annotation.updateAnnotation(
+	return body.annotation ? orm.func.annotation.updateAnnotation(
```
prabalsingh24: if body.annotation was undefined, and when oldAnnotation was also undefined, old==new and it was returning old

src/server/routes/entity/author.js
```diff
@@ -224,11 +224,13 @@ function transformNewForm(data) {
...
 			data.authorSection.beginArea.id,
-		beginDate: dateObjectToISOString(data.authorSection.beginDate),
+		beginDate: data.authorSection.beginDate ?
+			dateObjectToISOString(data.authorSection.beginDate) : '',
 		disambiguation: data.nameSection.disambiguation,
 		endAreaId: data.authorSection.endArea &&
 			data.authorSection.endArea.id,
-		endDate: data.authorSection.ended ? dateObjectToISOString(data.authorSection.endDate) : '',
+		endDate: data.authorSection.endDate ?
+			dateObjectToISOString(data.authorSection.endDate) : '',
```
prabalsingh24: same as publisher.js

src/server/routes/entity/entity.js
prabalsingh24: The releaseEvents is always an array of size 1 right?

MonkeyDo: At the moment yes.
If you want that bit of code to be more error-proof, you could do something like
`if (_.isNil(_.get(releaseEvents,'[0].areaId')) {…`

_.get returns the value at the path safely (won't throw an error if there's no `releaseEvents[0]` for example), and _.isNil determines if the value is null or undefined (instead of falsy; for example in this case if areaId = 0 it will be considered falsy and be overwritten with null [not sure that will ever happen, but…])

MonkeyDo: I just created this as well as you suggested, if you want to review it:
https://github.com/bookbrainz/bookbrainz-data-js/pull/212

prabalsingh24: Now that you've made changes in bookbrainz-data, this is not needed. 
Should I remove it or leave it untill those changes you made are merged in the master?

MonkeyDo: Let's leave it, it doesn't hurt to be clean on both sides.

PR #344, opened 2020-01-21
fix(edition) : Cannot read property 'year' of undefined

prabalsingh24:
### Problem
https://tickets.metabrainz.org/browse/BB-395
Cannot read property 'year' of undefined

### Solution
Now it checks if data.editionSection.releaseDate exists or not

### DIFF
--- a/src/server/routes/entity/edition.js
+++ b/src/server/routes/entity/edition.js
@@ -339,7 +339,7 @@ function transformNewForm(data) {
 	);
 
 	let releaseEvents = [];
-	if (data.editionSection.releaseDate.year) {
+	if (data.editionSection.releaseDate && data.editionSection.releaseDate.year) {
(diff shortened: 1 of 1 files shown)


src/server/routes/entity/edition.js
```diff
@@ -339,7 +339,7 @@ function transformNewForm(data) {
 	);
 
 	let releaseEvents = [];
-	if (data.editionSection.releaseDate.year) {
+	if (data.editionSection.releaseDate && data.editionSection.releaseDate.year) {
```
prabalsingh24: I think this would've worked also `if(data.edition.releaseData){..` because year is an empty string by default and will always be present.
But I think it's better to be safe :)

MonkeyDo: Indeed. I think also if by any chance you get something other than a {year,month,day} object (like a date string), calling dateObjectToISOString with it will throw an error.

PR #348, opened 2020-01-26
React fontawesome #bb371

prabalsingh24:
### Problem
https://tickets.metabrainz.org/browse/BB-371
 Migrating React-FontAwesome : edition.js

### Areas of Impact 
src/client/components/pages/entities/edition.js

### DIFF
--- a/src/client/components/pages/entities/edition.js
+++ b/src/client/components/pages/entities/edition.js
@@ -109,7 +109,7 @@ function EditionDisplayPage({entity, identifierTypes}) {
 		editionGroupSection = (
 			<div className="margin-bottom-d15">
 				<a href={`/edition-group/${entity.editionGroup.bbid}`}>
-					<FontAwesomeIcon icon="external-link"/>
+					<FontAwesomeIcon icon="external-link-alt"/>
 					<span>&nbsp;See all similar editions</span>
 				</a>
 			</div>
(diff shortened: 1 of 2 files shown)


src/client/components/pages/entities/edition.js
```diff
@@ -109,7 +109,7 @@ function EditionDisplayPage({entity, identifierTypes}) {
 		editionGroupSection = (
 			<div className="margin-bottom-d15">
 				<a href={`/edition-group/${entity.editionGroup.bbid}`}>
-					<Icon name="external-link"/>
+					<FontAwesomeIcon icon="external-link"/>
```
MonkeyDo: The icon should be external-link-alt according to https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes

I know there is a shim less file, but let's try to do everything cleanly since we're here modifying things :)

prabalsingh24: Also @MonkeyDo  I noticed this icon and some other icons used it other PR's are not imported https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/helpers/setupIconLibrary.js#L35-L42 . Should I add these icons in their PR's or we'll be adding them in the end in a separate PR. Will it create conflicts if we add them in their individual PR's ?

MonkeyDo: Go ahead and add them in the PR related to that icon/page.
I'll sort out any conflicts, they'll be easy to fix (we keep both changes).

PR #357, opened 2020-01-29
BB-378 Migrating React-FontAwesome

kingpin24:
BB-378 on Jira

### DIFF
--- a/src/client/entity-editor/button-bar/alias-button.js
+++ b/src/client/entity-editor/button-bar/alias-button.js
@@ -17,7 +17,7 @@
  */
 
 import {Button} from 'react-bootstrap';
-import Icon from 'react-fontawesome';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


kingpin24:
@MonkeyDo would you like all the different subtasks of the Migrating ticket to be in one PR or separate ones?

MonkeyDo:
@kingpin24 Well, I'm sure it'll be easier for everyone to have the remaining tasks in one PR.
Three things to remember though:
1. Don't forget to assign the subtasks to yourself to show you're working on it
2. Some icon names have changed; check if they have in this list https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes
3. Each icon we use on the website needs to be imported into an icon library, so make sure any icon you encounter is there (https://github.com/bookbrainz/bookbrainz-site/blob/2b183c6e366e3bbb513b2f5dfc6a2ef3ab7a682a/src/client/helpers/setupIconLibrary.js)

Thanks for tackling this, good continuation ! :)

PR #358, opened 2020-01-31
Editor revisions table - #BB-397

prabalsingh24:
### Problem
https://tickets.metabrainz.org/browse/BB-397

### Solution
Added pagination and improved table in the editor/:id/revisions

Screenshot from 2020-01-31 09-17-21

### Areas of Impact
src/client/components/pages/editor-revision
src/routes/editor
src/client/components/pages/parts/revision-table

### DIFF
--- a/src/client/components/pages/editor-revision.js
+++ b/src/client/components/pages/editor-revision.js
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2019 Prabal Singh
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import React from 'react';
+import RevisionsTable from './parts/revisions-table';
+import request from 'superagent-bluebird-promise';
+
+
+const {Pager, Button, ButtonGroup, DropdownButton, MenuItem} = bootstrap;
+
+class EditorRevisionPage extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+			from: this.props.from,
+			nextEnabled: this.props.results.length === this.props.size,
+			results: this.props.results,
+			size: this.props.size
+		};
+		// React does not autobind non-React class methods
+		this.triggerSearch = this.triggerSearch.bind(this);
+		this.handleClickPrevious = this.handleClickPrevious.bind(this);
+		this.handleClickNext = this.handleClickNext.bind(this);
+		this.handleResultsPerPageChange = this.handleResultsPerPageChange.bind(this);
+	}
+
+	triggerSearch() {
+		const pagination = `size=${this.state.size}&from=${this.state.from}`;
+		request.get(`./revisions/revisions?${pagination}`)
+			.then((res) => JSON.parse(res.text))
+			.then((data) => {
+				this.setState(prevState => ({
+					nextEnabled: data.length >= prevState.size,
+					results: data
+				}));
+			});
+	}
+
+	handleClickPrevious() {
+		this.setState(prevState => ({from: Math.max(prevState.from - prevState.size, 0)}), this.triggerSearch);
+	}
+
+	handleClickNext() {
+		this.setState(prevState => ({from: prevState.from + prevState.size}), this.triggerSearch);
+	}
+
+	handleResultsPerPageChange(value) {
+		this.setState({size: parseInt(value, 10)}, this.triggerSearch);
+	}
+
+	render() {
+		return (
+			<div className="container">
+				<div id="EditorRevisionPage">
+
+					<RevisionsTable results={this.state.results}/>
+					{
+						this.state.results && this.state.results.length ?
+							<div>
+								<hr className="thin"/>
+								<Pager>
+									<Pager.Item
+										previous disabled={this.state.from <= 0}
+										href="#" onClick={this.handleClickPrevious}
+									>
+										&larr; Previous Page
+									</Pager.Item>
+									<ButtonGroup>
+										<Button disabled>Results {this.state.from + 1} —
+											{this.state.results.length < this.state.size ?
+												this.state.from + this.state.results.length :
+												this.state.from + this.state.size
+											}
+										</Button>
+										<DropdownButton
+											dropup bsStyle="info" id="bg-nested-dropdown"
+											title={`${this.state.size} per page`}
+											onSelect={this.handleResultsPerPageChange}
+										>
+											<MenuItem eventKey="10">10 per page</MenuItem>
+											<MenuItem eventKey="20">20 per page</MenuItem>
+											<MenuItem eventKey="35">35 per page</MenuItem>
+											<MenuItem eventKey="50">50 per page</MenuItem>
+											<MenuItem eventKey="100">100 per page</MenuItem>
+										</DropdownButton>
+									</ButtonGroup>
+									<Pager.Item
+										next disabled={!this.state.nextEnabled}
+										href="#" onClick={this.handleClickNext}
+									>
+										Next Page &rarr;
+									</Pager.Item>
+								</Pager>
+
+							</div> :
+							null
+					}
+				</div>
+			</div>
+		);
+	}
+}
+
+
+EditorRevisionPage.displayName = 'EditorRevisionPage';
+EditorRevisionPage.propTypes = {
+	from: PropTypes.number,
+	results: PropTypes.array,
+	size: PropTypes.number
+};
+EditorRevisionPage.defaultProps = {
+	from: 0,
+	results: [],
+	size: 20
+};
+
+export default EditorRevisionPage;
--- a/src/client/components/pages/editor-revision.js
+++ b/src/client/components/pages/editor-revision.js
@@ -32,8 +32,12 @@ class EditorRevisionPage extends React.Component {
 			from: this.props.from,
 			nextEnabled: this.props.results.length === this.props.size,
 			results: this.props.results,
-			size: this.props.size
+			showRevisionEditor: this.props.showRevisionEditor,
+			showRevisionNote: this.props.showRevisionNote,
+			size: this.props.size,
+			tableHeading: this.props.tableHeading
 		};
+
 		// React does not autobind non-React class methods
 		this.triggerSearch = this.triggerSearch.bind(this);
 		this.handleClickPrevious = this.handleClickPrevious.bind(this);
@@ -70,7 +74,13 @@ class EditorRevisionPage extends React.Component {
 			<div className="container">
 				<div id="EditorRevisionPage">
 
-					<RevisionsTable results={this.state.results}/>
+					<RevisionsTable
+						results={this.state.results}
+						showRevisionEditor={this.props.showRevisionEditor}
+						showRevisionNote={this.props.showRevisionNote}
+						tableHeading={this.props.tableHeading}
+					/>
+
 					{
 						this.state.results && this.state.results.length ?
 							<div>
@@ -123,12 +133,18 @@ EditorRevisionPage.displayName = 'EditorRevisionPage';
 EditorRevisionPage.propTypes = {
 	from: PropTypes.number,
 	results: PropTypes.array,
-	size: PropTypes.number
+	showRevisionEditor: PropTypes.bool,
+	showRevisionNote: PropTypes.bool,
+	size: PropTypes.number,
+	tableHeading: PropTypes.string
 };
 EditorRevisionPage.defaultProps = {
 	from: 0,
 	results: [],
-	size: 20
+	showRevisionEditor: false,
+	showRevisionNote: false,
+	size: 20,
+	tableHeading: 'Recent Activity'
 };
 
 export default EditorRevisionPage;
--- a/src/client/components/pages/editor-revision.js
+++ b/src/client/components/pages/editor-revision.js
@@ -16,113 +16,43 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import * as bootstrap from 'react-bootstrap';
+import PagerElement from './parts/pager';
 import PropTypes from 'prop-types';
 import React from 'react';
 import RevisionsTable from './parts/revisions-table';
-import request from 'superagent-bluebird-promise';
 
-
-const {Pager, Button, ButtonGroup, DropdownButton, MenuItem} = bootstrap;
 
 class EditorRevisionPage extends React.Component {
 	constructor(props) {
 		super(props);
 		this.state = {
-			from: this.props.from,
-			nextEnabled: this.props.results.length === this.props.size,
-			results: this.props.results,
-			showRevisionEditor: this.props.showRevisionEditor,
-			showRevisionNote: this.props.showRevisionNote,
-			size: this.props.size,
-			tableHeading: this.props.tableHeading
+			results: this.props.results
 		};
 
-		// React does not autobind non-React class methods
-		this.triggerSearch = this.triggerSearch.bind(this);
-		this.handleClickPrevious = this.handleClickPrevious.bind(this);
-		this.handleClickNext = this.handleClickNext.bind(this);
-		this.handleResultsPerPageChange = this.handleResultsPerPageChange.bind(this);
+		this.searchResultsCallback = this.searchResultsCallback.bind(this);
+		this.paginationUrl = './revisions/revisions?q=';
 	}
 
-	triggerSearch() {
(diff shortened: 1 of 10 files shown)


src/client/components/pages/editor-revision.js
```diff
@@ -0,0 +1,134 @@
...
+		return (
+			<div className="container">
+				<div id="EditorRevisionPage">
+
+					<RevisionsTable results={this.state.results}/>
+					{
+						this.state.results && this.state.results.length ?
+							<div>
+								<hr className="thin"/>
+								<Pager>
```
MonkeyDo: Unless I'm mistaken, the Pager element is the same on the various pages we use it.
It would be a good candidate for a refactoring PR.

I took the liberty to create a ticket for you (don't hesitate to unassign yourself if you don't want to do it :) ) : https://tickets.metabrainz.org/browse/BB-406

src/client/components/pages/parts/revisions-table.js
```diff
@@ -42,41 +42,62 @@ function RevisionsTable(props) {
...
+					{
+						results.length > 0 ?
+							<tr>
+								<th className="col-sm-2">Revision ID</th>
+								<th className="col-sm-5">Modified entities</th>
+								{
+									results[0].noteContent ?
+										<th className="col-sm-3">Note</th> :
+										<th className="col-sm-3">User</th>
+								}
```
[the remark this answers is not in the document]
MonkeyDo: I can't add this to the relevant line, so I'm doing it here:

Another useful optional component property would be a 'title' defaulting to 'Recent Activity'.
That way for an editor's revisions page, we can go back to the title 'Revision History' which is a bit more precise.

src/server/routes/editor.js
```diff
@@ -250,47 +251,93 @@ router.get('/:id', (req, res, next) => {
...
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+
+	const orderedRevisions = await getOrderedRevisionForEditorPage(from, size, req);
+
+	const editor = await new Editor({id: req.params.id}).fetch();
+	const editorJSON = await getEditorTitleJSON(editor.toJSON(), TitleUnlock);
+
+	const props = generateProps(req, res, {
+		editor: editorJSON,
```
MonkeyDo: This editor prop seems to be unused now (editor name was displayed on each revision previously).

If so, it could either be deleted entirely (no need to `getEditorTitleJSON` is that case), or used for the RevisionsPage title ("Revision History for {editorName}"). I'll let you have a look at what works out best.

src/client/components/pages/parts/revisions-table.js
```diff
@@ -42,41 +42,62 @@ function RevisionsTable(props) {
...
-						<th className="col-sm-2">Revision ID</th>
-						<th className="col-sm-5">Modified entities</th>
-						<th className="col-sm-3">User</th>
-						<th className="col-sm-2">Date</th>
-
-					</tr>
+					{
+						results.length > 0 ?
+							<tr>
+								<th className="col-sm-2">Revision ID</th>
```
MonkeyDo: I wonder about these column sizes.
I've tried removing all the col-… classes on the th elements, and there are positive and negative aspects:

Positive: Better use of space, for example rev id is as small as it can while giving space to the large notes

Negative: Depending on content the columns are not the same size, which doesn't look great when going back or forth a page.
Notes often contain urls, which have no spaces, and as such will take as much space as possible, making the 'modified entities' column wrap first. Not sure I like that.

I'm conflicted about this. Let's keep it as-is for now, but if you're interested, keep that issue in the back of your mind.
Maybe there is a solution out there that allows us to set min- or max-width on certain columns (useful for example for rev id column [small] and notes column [we don't want it to be wider than, say, 40%])

src/client/components/pages/editor-revision.js
prabalsingh24: I am doing this in a different PR :)

src/server/routes/editor.js
```diff
@@ -250,47 +251,93 @@ router.get('/:id', (req, res, next) => {
...
-		.catch(next);
+
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+
+	const orderedRevisions = await getOrderedRevisionForEditorPage(from, size, req);
+
+	const editor = await new Editor({id: req.params.id}).fetch();
+	const editorJSON = await getEditorTitleJSON(editor.toJSON(), TitleUnlock);
+
```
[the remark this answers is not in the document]
prabalsingh24: Thanks :)

prabalsingh24:
When a row has multiple notes, I find it better if we display in a list with bullets using `<ul>` but on the other hand, it there is only one note, having only a single note with a bullet do not look very good. 
What do you think? 
Also idk why, `<li>` tag is creating an extra indentation, will have to have change css for that i guess. Maybe left margin for `<li>` tag

src/server/routes/editor.js
prabalsingh24: editor props is used actually. Not in the table but we need editor.id and editor.name for url and heading.(Sprinter Monkey). You were right though, we don't need to fetch authors (withRelated) while getting revisions. :)

MonkeyDo:
> When a row has multiple notes, I find it better if we display in a list with bullets using `<ul>` but on the other hand, it there is only one note, having only a single note with a bullet do not look very good.
> What do you think?
> Also idk why, `<li>` tag is creating an extra indentation, will have to have change css for that i guess. Maybe left margin for `<li>` tag

I think you could get by with adding a the character "•" in front of each note conditionally if there's more than one note.

Or the css way :) There's the [:only-child](https://www.w3schools.com/cssref/sel_only-child.asp) selector you can use to target a single <li> and set `list-style-type: none;` on it to hide the bullet.

I think either will work. The css way seems a bit cleaner.

src/client/components/pages/editor-revision.js
```diff
@@ -0,0 +1,150 @@
...
+	showRevisionEditor: PropTypes.bool,
+	showRevisionNote: PropTypes.bool,
+	size: PropTypes.number,
+	tableHeading: PropTypes.string
+};
+EditorRevisionPage.defaultProps = {
+	from: 0,
+	results: [],
+	showRevisionEditor: false,
+	showRevisionNote: false,
```
MonkeyDo: Shouldn't the default be `true` here?

src/client/components/pages/parts/revisions-table.js
```diff
@@ -28,67 +28,107 @@ const {formatDate} = utilsHelper;
...
+														{revision.editor.name}
+													</a>
+												</td> : null
+										}
+										{
+											showRevisionNote ?
+												<td>
+													{revision.noteContent.map(note => (
+														// eslint-disable-next-line react/jsx-key
+														<div>
```
MonkeyDo: There should really be key here. I see that there aren't any great candidates, considering using the array index doesn't work.
Let's instead find something unique to the note: its id.
We don't have access to it here, so you'll have to change `noteContent` instead to a `notes` array of objects {id, authorId, content} (in `getOrderedRevisionForEditorPage`).
You can then safely use `note.id` as the `key`.

It also makes sense if we want to reuse this mechanism for the API, where I would expect that format.

src/server/routes/editor.js
```diff
@@ -21,17 +21,18 @@ import * as auth from '../helpers/auth';
 import * as error from '../../common/helpers/error';
 import * as handler from '../helpers/handler';
 import * as propHelpers from '../../client/helpers/props';
+import * as utilis from '../helpers/utils';
```
MonkeyDo: utilis -> utils

src/server/routes/editor.js
```diff
@@ -250,47 +251,92 @@ router.get('/:id', (req, res, next) => {
...
-			throw new error.NotFoundError('Editor not found', req);
-		})
-		.catch(next);
+
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+
+	const orderedRevisions = await getOrderedRevisionForEditorPage(from, size, req);
+
+	const editor = await new Editor({id: req.params.id}).fetch();
```
MonkeyDo: The mechanism should catch if there is no existing Editor here.
Try visiting an non-existing user's revisions: http://localhost:9099/editor/12789345/revisions
That's what the
```
.catch(Editor.NotFoundError, () => {
	throw new error.NotFoundError('Editor not found', req);
})
```
 block was doing. You'll have to recreate that.

I think you should  also do that before calling `getOrderedRevisionForEditorPage`, or do the check (does the user exist?) in `getOrderedRevisionForEditorPage` itself.

src/server/routes/editor.js
```diff
@@ -250,47 +251,92 @@ router.get('/:id', (req, res, next) => {
...
+	}));
 });
 
+// eslint-disable-next-line consistent-return
+router.get('/:id/revisions/revisions', async (req, res, next) => {
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+
+	try {
+		const orderedRevisions = await getOrderedRevisionForEditorPage(from, size, req);
```
MonkeyDo: Here as well you should check first that the Editor exists, and return a `new error.NotFoundError('Editor not found', req);` otherwise.
At the moment visiting for a non-existing user (http://localhost:9099/editor/12789345/revisions/revisions)  I get an empty array, which is better that a random error, but I would expect a 404.

I think it's better to change check for the Editor in `getOrderedRevisionForEditorPage` then.

Again, this will be useful for reusing these tools for the API.

src/server/routes/editor.js
prabalsingh24: Yeah sorry

src/client/components/pages/editor-revision.js
prabalsingh24: If default is `true` then I have to pass `false` in props for the other thing and vice versa. I set it to `false` to avoid the case when editor or note is true but it isn't present 
I can change switch it if you want but I don't think it makes any difference.

src/client/components/pages/parts/revisions-table.js
prabalsingh24: Oh the key is used for API, I didn't know that :P

MonkeyDo: No, I didn't explain well :D Sorry !

The key is for React to keep track of reusable DOM elements (it needs to know what elements need to be updated. [Here's a better explanation](https://blog.arkency.com/2014/10/react-dot-js-and-dynamic-children-why-the-keys-are-important/)).

The format of `revision.notes` is the part that matches what I would expect the API to return.
Currently there is no API endpoint to get editors or revisions, so it's a question of shaping that new getOrderedRevisionForEditorPage function to be easier to reuse.

prabalsingh24: Thankss :D

src/client/components/pages/editor-revision.js
MonkeyDo: I'm not sure I understand what you mean by "I have to pass false in props for the other thing and vice versa".

I assume that specifically for this EditorRevisionPage, we always want to display the notes and don't want to display the editor, so the default props should reflect that, regardless of what is passed to the EditorRevisionPage 
 when it is instantiated.

prabalsingh24: Oh I am so sorry I thought this was default props for revision table. My bad :(

prabalsingh24: Yeah you're correct. I will have to change it in `revisionPage` also I guess. I will fix it. Thanks :)

MonkeyDo: It's easy to get confused when looking at similar files for weeks :)
That's why pair programming and reviews are so useful — for everyone!

prabalsingh24:
We can have multiple notes for one revision. And that too from different editors. Do you think note's editor name should be displayed in note column or is that too much information for the table?

MonkeyDo:
> We can have multiple notes for one revision. And that too from different editors. Do you think note's editor name should be displayed in note column or is that too much information for the table?

Good question. I see two easy possibilities:
1. Put the author's name as a link inline with the note like this:
  "[...] end of my note. — Monkey"
2. Considering this isn't so important for this table view, maybe it would be enough to have the note author as a title attribute for the note DOM node, so that when you hover it for a second the name will appear. I'm thinking something like ```title={`note by ${myUsernameVar}`}```

prabalsingh24:
This is how the table is looking right now :)) 
Have a look when you're free :D 
Screenshot from 2020-02-06 01-28-48

MonkeyDo:
> This is how the table is looking right now :))
> Have a look when you're free :D
> Screenshot from 2020-02-06 01-28-48

Looking pretty nice ! I like the italic.
However I think you would save some space by attaching the name right after the note, rather than on the next line. What do you think?

prabalsingh24:
> > > This is how the table is looking right now :))
> > > Have a look when you're free :D
> > > Screenshot from 2020-02-06 01-28-48
> > 
> > 
> > Looking pretty nice ! I like the italic.
> > However I think you would save some space by attaching the name right after the note, rather than on the next line. What do you think?
> 
> I tried that. But for some cases, the editor name comes in next line and in the left. Because note takes up all the space

Wait I think it can be done.

prabalsingh24:
This is how the table looks now. What do you think? 
Screenshot from 2020-02-06 02-55-07

prabalsingh24:
Yeah this looks much better.

prabalsingh24:
I am confused :P You decide @MonkeyDo . Which one looks better?

src/client/components/pages/parts/revisions-table.js
```diff
@@ -28,67 +28,115 @@ const {formatDate} = utilsHelper;
...
+														<div key={note.id}>
+															{/* eslint-disable-next-line react/no-unescaped-entities */}
+															"{note.content}"
+															<a className="pull-right" href={`/editor/${note.author.id}`}>
+																<i>
+																	—{note.author.name}
+																</i>
+															</a>
+															<br/>
+															<br/>
```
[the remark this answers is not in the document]
prabalsingh24: I like this thin line between notes. Otherwise in some cases it was getting difficult to distinguish b/w two notes. That is why I added `<br/>` . and yeah it was creating an empty line in some cases, I should've tested it properly.
I don't know much css, I was struggling with it. Thanks :D
Also I personally would've liked if note-author is at the right side. Especially for the second note in this picture. What do you think?

prabalsingh24:
> This is already a solid good page, but as you might have guessed by now I'm a bit of a stickler :p
> I hope you don't mind me asking for a million small adjustments; i think you do a really good job to start with, and I'm happy to take the time to make it Just Right™ together.
> 
> I come from a graphic design and webdesign background (the reason why i'm always jumping to tiny details), and I've done a looooot of CSS, so there are a few tricks in my brain that aren't necessarily well known that can help us out.
> 
> Keep up the good work!
> I'm excited to release those improvements to the revision tables throughout, I think it's a really good addition to the website and once we've tested a bit more and we're satisfied, I'll want to publish to the main website and publicise it.

No I like that you're getting satisfied with okayish code and asking for changes even for very small things. One of the reasons I started open source - getting my code reviewed and learning :D

prabalsingh24:
> > This is already a solid good page, but as you might have guessed by now I'm a bit of a stickler :p
> > I hope you don't mind me asking for a million small adjustments; i think you do a really good job to start with, and I'm happy to take the time to make it Just Right™ together.
> > I come from a graphic design and webdesign background (the reason why i'm always jumping to tiny details), and I've done a looooot of CSS, so there are a few tricks in my brain that aren't necessarily well known that can help us out.
> > Keep up the good work!
> > I'm excited to release those improvements to the revision tables throughout, I think it's a really good addition to the website and once we've tested a bit more and we're satisfied, I'll want to publish to the main website and publicise it.
> 
> No I like that you're getting satisfied with okayish code and asking for changes even for very small things. One of the reasons I started open source - getting my code reviewed and learning :D

*not getting satisfied :P

src/client/components/pages/parts/revisions-table.js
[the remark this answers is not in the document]
MonkeyDo: I was wondering how that would work, and I was curious so i had a go :)

Here's the markup and CSS I ended up with:
```
<div className="revision-note" key={note.id}>
	<span className="note-content">
		{note.content}
		<a
			className="note-author pull-right" href={`/editor/${note.author.id}`}
		>
			—{note.author.name}
		</a>
	</span>
</div>
```

```
.revision-note {
	&:not(:last-child) {
		border-bottom: 1px solid @disabled-color;
		margin-bottom: 0.3em;
		padding-bottom: 0.3em;
	}
	.note-content {
		quotes: "\201C" "\201D";
		&::before {
				content: open-quote;
		}
		&::after {
				content: close-quote;
		}
	}
	.note-author {
		&:extend(.small);
		font-style: italic;
		white-space: nowrap;
		margin-left: 0.2em;
	}
}
```

prabalsingh24:
I will resolve these conflicts after `PagerElement` is finished and merged ( `PagerElement` will be used in this too)

MonkeyDo:
Apart form the remaining conflicts (#360 is now merged), I think this is pretty much ready to go !

prabalsingh24:
> Apart from the remaining conflicts (#360 is now merged), I think this is pretty much ready to go !

Yes, I will replace the pagination code with `pagerElement`

prabalsingh24:
Sorry about extra commits. First I merged master without updating my fork. Also while resolving conflicts I made couple of mistakes :P

MonkeyDo:
> Sorry about extra commits. First I merged master without updating my fork. Also while resolving conflicts I made couple of mistakes :P

That's fine, everyone makes mistakes, so it's expected there will be commits fixing them :)

If you look at the commits this PR would merge, it's still quite clean: https://github.com/bookbrainz/bookbrainz-site/pull/358/commits

PR #360, opened 2020-02-05
Refactor pagination component

prabalsingh24:
### Problem
https://tickets.metabrainz.org/browse/BB-406
Refactoring Pagination Component

The bootstrap Pager element is used across various pages (search, revisions page, editor's revisions page), each having a copy of the code.

This PR is about refactoring that pager code into a component that can be reused

### Solution

Added a new component - PagerElement and reused it for search-results and revision-table

### DIFF
--- a/src/client/components/pages/parts/pager.js
+++ b/src/client/components/pages/parts/pager.js
@@ -1,0 +1,134 @@
+/*
+ * Copyright (C) 2019 Prabal Singh
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import PropTypes from 'prop-types';
+import React from 'react';
+import request from 'superagent-bluebird-promise';
+
+
+const {Pager, Button, ButtonGroup, DropdownButton, MenuItem} = bootstrap;
+
+class PagerElement extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+			from: this.props.from,
+			nextEnabled: this.props.results.length === this.props.size,
+			results: this.props.results,
+			size: this.props.size
+		};
+		this.handleClickPrevious = this.handleClickPrevious.bind(this);
+		this.handleClickNext = this.handleClickNext.bind(this);
+		this.handleResultsPerPageChange = this.handleResultsPerPageChange.bind(this);
+		this.triggerSearch = this.triggerSearch.bind(this);
+	}
+
+	triggerSearch() {
+		const pagination = `&size=${this.state.size}&from=${this.state.from}`;
+		const collectionString = this.state.collection ? `&collection=${this.state.collection}` : '';
+
+		request.get(`${this.props.paginationUrl}${this.props.query}${collectionString}${pagination}`)
+			.then((res) => JSON.parse(res.text))
+			.then((data) => {
+				this.setState(prevState => ({
+					nextEnabled: data.length >= prevState.size,
+					results: data
+				}));
+				this.props.changeDataInTable(data);
+			});
+	}
+
+	handleClickPrevious() {
+		this.setState(prevState => ({from: Math.max(prevState.from - prevState.size, 0)}), this.triggerSearch);
+	}
+
+	handleClickNext() {
+		this.setState(prevState => ({from: prevState.from + prevState.size}), this.triggerSearch);
+	}
+
+	handleResultsPerPageChange(value) {
+		this.setState({size: parseInt(value, 10)}, this.triggerSearch);
+	}
+
+	render() {
+		return (
+			<div id="PagerElement">
+				{
+					this.state.results && this.state.results.length ?
+						<div>
+							<hr className="thin"/>
+							<Pager>
+								<Pager.Item
+									previous disabled={this.state.from <= 0}
+									href="#" onClick={this.handleClickPrevious}
+								>
+									&larr; Previous Page
+								</Pager.Item>
+								<ButtonGroup>
+									<Button disabled>Results {this.state.from + 1} —
+										{this.state.results.length < this.state.size ?
+											this.state.from + this.state.results.length :
+											this.state.from + this.state.size
+										}
+									</Button>
+									<DropdownButton
+										dropup bsStyle="info" id="bg-nested-dropdown"
+										title={`${this.state.size} per page`}
+										onSelect={this.handleResultsPerPageChange}
+									>
+										<MenuItem eventKey="10">10 per page</MenuItem>
+										<MenuItem eventKey="20">20 per page</MenuItem>
+										<MenuItem eventKey="35">35 per page</MenuItem>
+										<MenuItem eventKey="50">50 per page</MenuItem>
+										<MenuItem eventKey="100">100 per page</MenuItem>
+									</DropdownButton>
+								</ButtonGroup>
+								<Pager.Item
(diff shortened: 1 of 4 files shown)


src/client/components/pages/search.js
```diff
@@ -73,32 +67,19 @@ class SearchPage extends React.Component {
...
-					nextEnabled: data.length >= prevState.size,
-					results: data.filter(result => !isNil(result))
-				}));
-			});
-	}
 
-	handleClickPrevious(event) {
-		this.setState(prevState => ({from: Math.max(prevState.from - prevState.size, 0)}), this.triggerSearch);
+		const url = `./search/search?q=${this.state.query}${collectionString}${pagination}`;
+		// TODO redirect to url;
```
prabalsingh24: This is left to do. I can't make it work for now. I will try again. Also the url is wrong. There should be only one search in it. :P

prabalsingh24: Right now you cannot use search bar when you're in search-result table page. The pager is working though

src/client/components/pages/parts/pager.js
```diff
@@ -0,0 +1,134 @@
...
+										dropup bsStyle="info" id="bg-nested-dropdown"
+										title={`${this.state.size} per page`}
+										onSelect={this.handleResultsPerPageChange}
+									>
+										<MenuItem eventKey="10">10 per page</MenuItem>
+										<MenuItem eventKey="20">20 per page</MenuItem>
+										<MenuItem eventKey="35">35 per page</MenuItem>
+										<MenuItem eventKey="50">50 per page</MenuItem>
+										<MenuItem eventKey="100">100 per page</MenuItem>
+									</DropdownButton>
```
MonkeyDo: There's also a nagging styling issue that was here before. My bad !
Now that it's all in one place it's easier to fix :)

This is what I see:
Capture d’écran 2020-02-06 à 12 59 20

Whatever I did doesn't look good ! :)
I think it is due to some css rules I added willy-nilly. Theres a `.pager li > a, .pager li > span` rule that was probably meant for the 'previous page' and 'next page' buttons only but instead is applied to the dropdown menu elements too.
That can be fixed by making sure we target only `li` elements that are direct children of `.pager`:
`.pager > li > a, .pager > li > span`

src/client/components/pages/search.js
```diff
@@ -116,46 +97,14 @@ class SearchPage extends React.Component {
...
-									href="#" onClick={this.handleClickNext}
-								>
-									Next Page &rarr;
-								</Pager.Item>
-							</Pager>
-						</div> :
-						null
-				}
+				<PagerElement
+					changeDataInTable={this.changeDataInTable}
```
MonkeyDo: This name suggests to me that calling it will change the data.
Instead it's a callback; I would suggest another name to reflect that (everywhere it is used): `searchResultsCallback`

prabalsingh24: Yeah I didn't give much thought while naming the function. Thanks :D

MonkeyDo: It's always a tricky exercise, naming variables and functions !

src/client/components/pages/parts/pager.js
prabalsingh24: > I think it is due to some css rules I added willy-nilly. Theres a `.pager li > a, .pager li > span` rule that was probably meant for the 'previous page' and 'next page' buttons only but instead is applied to the dropdown menu elements too.
> That can be fixed by making sure we target only `li` elements that are direct children of `.pager`:
> `.pager > li > a, .pager > li > span`

In stlye.less, there was #searchPage, I renamed it to #pageWithPagination. Is that okay? I think by making it a genric name we can reuse it. Otherwise we would have to add for revisionPage again. And also for editor-page-revision.

prabalsingh24:
I am not doing url history thing in this PR. I'll do it in differnet PR later. I want to complete revisions table first. Only refactoring is done in this PR.

MonkeyDo:
> I am not doing url history thing in this PR. I'll do it in differnet PR later. I want to complete revisions table first. Only refactoring is done in this PR.

Yes, I think that's a smart move. Separation of concerns ! :)

> there was #searchPage, I renamed it to #pageWithPagination. Is that okay? […]

Ah, yes, good find !

prabalsingh24:
Sorry about that entity dropdown issue. It was working in my pc but before pushing the changes I remembered you asked not to use collection in `pagerElement` and to just pass `fullQuery`. I made mistake while changing it. :P

prabalsingh24:
Also Thanks I didn't know the use of componentDidUpdate properly. It's sad though, I spent alot of time making that ref thing work haha :P

prabalsingh24:
https://reactjs.org/docs/react-component.html#componentdidupdate
`You may call setState() immediately in componentDidUpdate() but note that it must be wrapped in a condition `
But still I had to use `// eslint-disable-next-line react/no-did-update-set-state`

MonkeyDo:
> Also Thanks I didn't know the use of componentDidUpdate properly. It's sad though, I spent alot of time making that ref thing work haha :P

Well, to be honest there are times where you need a reference to a component, and now that you know how to do it it'll be very useful for next time you encounter such an issue :D

Also useful to understand what's going on if you run into it in a codebase ;)  => https://github.com/bookbrainz/bookbrainz-site/search?q=ref%3D%7B%28ref%29+%3D%3E&unscoped_q=ref%3D%7B%28ref%29+%3D%3E

src/client/components/pages/search.js
```diff
@@ -43,18 +39,15 @@ class SearchPage extends React.Component {
 
 		this.state = {
 			from: 0,
-			nextEnabled: this.props.initialResults.length === this.props.resultsPerPage,
 			query: this.props.query,
 			results: this.props.initialResults,
 			size: this.props.resultsPerPage
```
MonkeyDo: I think for any page that uses the PagerElements, you don't need to store the `size` in the state. You can pass it directly to PagerElement as `this.props.size`.

src/client/components/pages/parts/pager.js
```diff
@@ -0,0 +1,140 @@
...
+		this.handleClickPrevious = this.handleClickPrevious.bind(this);
+		this.handleClickNext = this.handleClickNext.bind(this);
+		this.handleResultsPerPageChange = this.handleResultsPerPageChange.bind(this);
+		this.triggerSearch = this.triggerSearch.bind(this);
+	}
+
+	componentDidUpdate(prevProps, prevState) {
+		if (prevProps.query !== this.props.query) {
+			// eslint-disable-next-line react/no-did-update-set-state
+			this.setState(this.props, this.triggerSearch);
```
MonkeyDo: I guess you should only set the query prop here (`{query: this.props.query}`), even though React is optimized and should not change the other properties if they are the same value.

It'll make the code clearer to understand though, and I think it might allow you to remove the `react/no-did-update-set-state` comment. (I haven't checked, don't quote me on that :p )

src/client/components/pages/search.js
MonkeyDo: And I think the same is true for the `from` property. It should only change when `triggerSearch` is called and the query has changed.

src/client/components/pages/parts/pager.js
prabalsingh24: Yeah alright. I'll have to update `from : 0` then.
Also should the `size` be updated? If I'm searching something and I've set size to 100, I'd want the new results table with size 100 only right?

MonkeyDo: For `from` I think you don't need to react to any update, considering only the PagerElement changes `from`.
We do want to pass it from the parent element, but only so that we can for example add `?from=150`  to the url and have that be passed down to the pager.

Unless I'm mistaken, the same thing applies for the `size`: you only modify it in PagerElement, so there's no need to react to props changes.

src/client/components/pages/search.js
prabalsingh24: I'm on my phone right now. I'll have a look at this in the evening or maybe tomorrow. I've got classes right now :P

src/client/components/pages/parts/pager.js
prabalsingh24: I'll have a look later

prabalsingh24:
The last commit is for https://tickets.metabrainz.org/browse/BB-416
Earlier it was setState->triggerSearch->setState . Now I've changed it to triggerSearch->setState
What do you think?

MonkeyDo:
> The last commit is for https://tickets.metabrainz.org/browse/BB-416
> Earlier it was setState->triggerSearch->setState . Now I've changed it to triggerSearch->setState
> What do you think?

Nice catch !

MonkeyDo:
This is now updated on beta.bookbrainz.org, if you want to test it there :)

PR #370, opened 2020-02-09
BB-417 fix(revision-table) : remove unneccesary div tag and use 'pull-right'

prabalsingh24:
### Problem
https://tickets.metabrainz.org/projects/BB/issues/BB-417
For the revisions table, right margin was lesser than left. It was looking unsymmetrical 
In the main page, `see all revisions` button was showing on the left side for smaller screens( mobile)

### Solution
Removed an unnecessary `<div id=container>` (made it same as search result table)
For the main page, used `pull-right` for `see all revisions` button 

Screenshot from 2020-02-09 18-07-31

Screenshot from 2020-02-09 18-07-39

### Areas of Impact
 src/client/components/pages/index.js
 src/client/components/pages/revisions.js

### DIFF
--- a/src/client/components/pages/index.js
+++ b/src/client/components/pages/index.js
@@ -212,20 +212,17 @@ class IndexPage extends React.Component {
 						Register!
 					</Button>
 				</div>
-				<RevisionsTable
-					results={this.props.recent}
-				/>
 				<div>
-					<Row>
-						<Col mdOffset={10}>
-							<Button
-								bsStyle="primary"
-								href="/revisions"
-							>
-								See all revisions
-							</Button>
-						</Col>
-					</Row>
+					<RevisionsTable
+						results={this.props.recent}
+					/>
+					<Button
+						bsStyle="primary"
+						className="pull-right"
+						href="/revisions"
+					>
+						See all revisions
+					</Button>
 				</div>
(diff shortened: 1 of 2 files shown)


src/client/components/pages/index.js
```diff
@@ -212,20 +212,17 @@ class IndexPage extends React.Component {
...
-								See all revisions
-							</Button>
-						</Col>
-					</Row>
+					<RevisionsTable
+						results={this.props.recent}
+					/>
+					<Button
+						bsStyle="primary"
+						className="pull-right"
```
MonkeyDo: I think the button would be better if it was centered.
On mobile it works OK on the right, but on desktop size I feel it's not visible enough and I find myself looking for it.

You can wrap a `<div className="text-center">` around the button to center it.
What do you think?

prabalsingh24: Like this?

Screenshot from 2020-02-12 08-02-24

MonkeyDo: I think that works better yes. My eye naturally finds it at the end of the table

prabalsingh24: > I think that works better yes. My eye naturally finds it at the end of the table

I'll make the change

PR #373, opened 2020-02-13
fixed margin issue of brand logo in mobile view

endurance21:
### Problem
https://tickets.metabrainz.org/browse/BB-420

### before 
issuenavbar2

### after
image

### Areas of Impact
src/client/sytylesheets/style.less

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -353,7 +353,7 @@ hr.wide {
 	margin-top: 40px;
     margin-bottom: 40px;
 }
-#searchPage .pager .dropdown-menu {
+#pageWithPagination .pager .dropdown-menu {
 	li {
 	  display: initial;
 	  > a,
@@ -376,5 +376,36 @@ hr.wide {
 @media only screen and (max-width: 600px) {
 	#searchpage-button-group > * {
 		width: 100%;
+	}
+}
+
+.BookBrainz > .container-fluid > .navbar-header{
+	margin-left: 0px;
+	margin-right: 0px;
+	>.navbar-toggle{
+		margin-right: 0px;
+	}
+
+}
+.revision-note {
+	&:not(:last-child) {
+		border-bottom: 1px solid @disabled-color;
+		margin-bottom: 0.3em;
+		padding-bottom: 0.3em;
+	}
+	.note-content {
+		quotes: "\201C" "\201D";
+		&::before {
+			content: open-quote;
+		}
+		&::after {
+			content: close-quote;
+		}
+	}
+	.note-author {
+		&:extend(.small);
+		font-style: italic;
+		white-space: nowrap;
+		margin-left: 0.2em;
 	}
 }
--- a/DEPENDENCIES_MANUAL_INSTALL.md
+++ b/DEPENDENCIES_MANUAL_INSTALL.md
@@ -49,10 +49,15 @@ Then, uncompress the `latest.sql.bz2` file, using the bzip2 command:
     bzip2 -d latest.sql.bz2
 
 This will give you a file that you can restore into PostgreSQL, which will
-set up data identical to the data we have on the bookbrainz.org website. To do
+set up data identical to the data we have on the bookbrainz.org website. First, you must create the necessary role and database with these two commands:
+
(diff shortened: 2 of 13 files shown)


src/client/stylesheets/style.less
```diff
@@ -378,3 +378,9 @@ hr.wide {
 		width: 100%;
 	}
 }
+
+@media only screen and(max-width:768px){
+.navbar-header{
+	margin-left:-13px !important;
```
prabalsingh24: Hi. Using `!important` works but I think it's not the ideal way to do. Have a look at this comment
https://github.com/bookbrainz/bookbrainz-site/pull/340#discussion_r366814767

endurance21: @prabalsingh24 
yes ! you are right. 
i have made the necessary changes and i'm about to change the PR,
thank you reviewing it

endurance21: @MonkeyDo 
i have done the changes and i think this would be done with merging .

src/client/stylesheets/style.less
```diff
@@ -378,3 +378,11 @@ hr.wide {
 		width: 100%;
 	}
 }
+
+
+@media only screen and (max-width: 768px){
+	.BookBrainz > .container-fluid > .navbar-header{
+		margin-left: -13px;
+	}
```
MonkeyDo: This seems to work, but I think it would be  better to target the `navbar-brand` element instead.
The idea is to keep the `navbar-header` as full width, and rely on the `container-fluid`'s padding, the same that sets the space of the hamburger button on the right side.
Here's what your proposal looks like if I visualize the margins/padding:
Capture d’écran 2020-02-16 à 12 44 27
You can see there's an empty space on the left.

Instead, I propose to set the `navbar-brand` element's margin-left to 0 (currently it is set to -26px), which would look like this:
Capture d’écran 2020-02-16 à 12 45 07

The padding for the container-fluid and the navbar-brand would collapse (they count as one)
Capture d’écran 2020-02-16 à 12 44 58


The navbar-brand's negative margin is currnetly targetted with these CSS selectors: `body.BookBrainz .navbar .navbar-brand.logo`. Adding `> .container-fluid` in there like you did will works nicely :)

endurance21: hey @monkeyDo !

i are right!!!

but i got  one more implementation for the solution that is more i robust and intuitive as i think.( almost borrowed from your idea);


right now, we have following elements structure 

container-fluid

here , "container-fluid" maintains 21px distance from left as well as right which is ok to make things symmetric!

so "navbar-header" should not be given any negative margin after that , otherwise our structure won't look symmetric anymore!

so i will remove all negative margin from "navbar-header" , only problem is with "toggle-button"  which will have extra distance from right because of "container-fluid" and its own margin
21px + 21px  = 42px in total!!!  that really sucks!!!! 

extramarginfortogglebutton





:::::::::::::::::::::::::::::::::::::::::::::::::::::::::::now the solution is :::::::::::::::::::::::::::::::::::::::::::::::::::::::

 1. i will remove every negative margin on "navbar-header".
2. i will remove extra 21px margin from "toggle-button" so that it would be only at a distance of 21px from right which is due to "container-fluid" !!
that's it

endurance21: @MonkeyDo 
 i have done the promised changes .
and now it looks very neat and clean with "NO NEGATIVE STUFFS"
and it is more symmetric than before!!!!

endurance21:
@MonkeyDo
I didn't get you, how can I resolve merge conflicts
On your branch?.

prabalsingh24:
You have to update your master and merge it with this branch. You'll have to resolve conflicts you get that time.

prabalsingh24:
Did you get it?

MonkeyDo:
Here's more information

A good explanation: https://www.git-tower.com/learn/git/ebook/en/command-line/advanced-topics/merge-conflicts

For GitHub: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/resolving-a-merge-conflict-on-github

endurance21:
@MonkeyDo
I have resolved the conflicts. 
I think it is ready to go!!!

PR #376, opened 2020-02-20
changed styling of button-bar

endurance21:
### Problem
https://tickets.metabrainz.org/projects/BB/issues/BB-421?filter=allopenissues

### Solution
before ::::
buttons (1)

after::::
solvedbuttonbar

### Areas of Impact
bookbrainz-site/src/client/entity-editor/button-bar/alias-button.js
bookbrainz-site/src/client/entity-editor/button-bar/button-bar.js
bookbrainz-site/src/client/entity-editor/button-bar/identifier-button.js

### DIFF
--- a/src/client/entity-editor/button-bar/button-bar.js
+++ b/src/client/entity-editor/button-bar/button-bar.js
@@ -25,6 +25,7 @@ import {
 import {validateAliases, validateIdentifiers} from '../validators/common';
 
 import AliasButton from './alias-button';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import IdentifierButton from './identifier-button';
 import PropTypes from 'prop-types';
 import React from 'react';
@@ -67,15 +68,15 @@ function ButtonBar({
 	return (
 		<div>
 			<form>
-				<Row className="margin-top-1">
-					<Col className="text-center" md={4}>
+				<Row >
+					<Col className="text-center margin-top-1" md={4} >
 						<AliasButton
 							aliasesInvalid={aliasesInvalid}
 							numAliases={numAliases}
 							onClick={onAliasButtonClick}
 						/>
 					</Col>
-					<Col className="text-center" md={4}>
+					<Col className="text-center margin-top-1" md={4}>
 						<IdentifierButton
 							identifiersInvalid={identifiersInvalid}
 							numIdentifiers={numIdentifiers}
@@ -84,12 +85,13 @@ function ButtonBar({
 					</Col>
 					{
 						!disambiguationVisible &&
-						<Col className="text-center" md={4}>
+						<Col className="text-center margin-top-1" md={4}>
 							<Button
-								bsStyle="link"
+								bsStyle="success"
 								onClick={onDisambiguationButtonClick}
 							>
-								Add disambiguation…
+								<FontAwesomeIcon icon="plus"/>
+								<span>&nbsp; Add disambiguation… </span>
 							</Button>
 						</Col>
 					}
--- a/src/client/entity-editor/button-bar/alias-button.js
+++ b/src/client/entity-editor/button-bar/alias-button.js
@@ -50,9 +50,10 @@ function AliasButton({
 		<FontAwesomeIcon className="margin-right-0-5 text-danger" icon="times"/>;
 
 	return (
-		<Button bsStyle="link" {...props}>
+		<Button bsStyle="success" {...props}>
 			{iconElement}
-			{text}
+			<FontAwesomeIcon icon="plus"/>
+			<span>&nbsp; {text} </span>
 		</Button>
 	);
 }
--- a/src/client/entity-editor/button-bar/identifier-button.js
+++ b/src/client/entity-editor/button-bar/identifier-button.js
@@ -50,9 +50,10 @@ function IdentifierButton({
 		<FontAwesomeIcon className="margin-right-0-5 text-danger" icon="times"/>;
 
 	return (
-		<Button bsStyle="link" {...props}>
+		<Button bsStyle="success" {...props}>
 			{iconElement}
-			{text}
+			<FontAwesomeIcon icon="plus"/>
+			<span>&nbsp;{text}</span>
 		</Button>
 	);
 }
--- a/src/client/entity-editor/edition-section/edition-section.js
+++ b/src/client/entity-editor/edition-section/edition-section.js
@@ -398,10 +398,11 @@ function EditionSection({
 				<Row>
 					<Col className="text-center" md={4} mdOffset={4}>
 						<Button
-							bsStyle="link"
+							bsStyle="success"
 							onClick={onPhysicalButtonClick}
 						>
-							Add physical data…
+							<FontAwesomeIcon icon="plus"/>
+							<span>&nbsp;Add physical data…</span>
 						</Button>
 					</Col>
 				</Row>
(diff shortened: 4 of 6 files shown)


endurance21:
@prabalsingh24 
Thanks for the branch take 😅, I will keep in mind from next time.

Buttons are indeed in same row, I just presented the mobile view. Now, in mobile view buttons will have to be like that only, i.e in different rows.. In Web view it is in same row.. 

And as far as icon is concerned, I was thinking of doing that too. Thanks for the reference to your  fontAwesome icons.

I am doing the icon changes right there.

prabalsingh24:
Oh sorry about that. I didn't test in my local server. I just saw the picture. It looked like web browser :p.
You go ahead and add icons here. :)

endurance21:
> Oh sorry about that. I didn't test in my local server. I just saw the picture. It looked like web browser :p.
> You go ahead and add icons here. :)

added them

endurance21:
@prabalsingh24 is it okay now?

src/client/entity-editor/button-bar/button-bar.js
```diff
@@ -17,6 +17,7 @@
  */
 
 import {Button, Col, Row} from 'react-bootstrap';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
```
prabalsingh24: This will come at the end. Multiple imports should be before single imports.

prabalsingh24:
Also rewriting commits was not needed, and you should avoid using `git push --force`. You're working alone in this so it's fine but when working in a team it's better to have an extra commit than using --force
https://stackoverflow.com/questions/18357511/git-remove-committed-file-after-push/22041320#22041320
` while potentially running into a conflict with your colleague who may have pulled it already, and who will grow grey hair and lose lots of time trying to reconcile his local branch head with the central one:`

endurance21:
@prabalsingh24 
thanks for the report , 
i have modified the pr and is passing the eslint tests now, i actually  forgot some extra trailing lines and ordering of imports.

endurance21:
> Also rewriting commits was not needed, and you should avoid using `git push --force`. You're working alone in this so it's fine but when working in a team it's better to have an extra commit than using --force
> https://stackoverflow.com/questions/18357511/git-remove-committed-file-after-push/22041320#22041320
> ` while potentially running into a conflict with your colleague who may have pulled it already, and who will grow grey hair and lose lots of time trying to reconcile his local branch head with the central one:`

ya , it's a good practise,
thanks .

prabalsingh24:
Your last commit didn't fix 

> @prabalsingh24
> thanks for the report ,
> i have modified the pr and is passing the eslint tests now, i actually forgot some extra trailing lines and ordering of imports.
 Umm no tests are failing?..

endurance21:
@MonkeyDo are we okay, with it?..

prabalsingh24:
Tests are still failing :P

endurance21:
updated !

endurance21:
@MonkeyDo
Have a look over this. Thanks.

prabalsingh24:
> @MonkeyDo
> Have a look over this. Thanks.

I remember that he said he's little busy this week. I am sure he'll review this when he gets free, even one of my PR's review is pending XD

endurance21:
@prabalsingh24
Ya i remember that.
I just marked the final say of our conversation, by that comment... 😅😅

endurance21:
i have also changed the button on add edition page 
"ADD PHYSICAL DATA" button

prabalsingh24:
> i have also changed the button on add edition page
> "ADD PHYSICAL DATA" button

It would've been better if you made a new commit for this. Easy to review :)

endurance21:
@prabalsingh24
Actually i thought, changing styles of button is same type of work, either done on button bars or on edition page. 
So why should I explicitly mention that.
But ya for the sake of understanding, your say is appreciable...

endurance21:
@MonkeyDo 
thanks for reviewing .
1. your first concern about fitness of buttons 
    >> this is only the problem in devices other than mobile devices .  so i only need to make the solution for devices other than mobile . i am working on it.
2.your second concern about icons to display need of rectification .
  >> for this i will just make the icon element dynamic , and implement a state to entitty-editor component that is just a bool that will store if rectification needed or not .
as per the input , the state of entity-editor component  will change  (with setstate) and so as the dynamic icon will be update there .

MonkeyDo:
After thinking about it for a bit, I think there are some considerations that will solve a few issues:

1. I think the disambiguation should always be there, meaning we won't need a button
2. The identifiers should be presented on the page without having to open a popup. There's a ticket open for that: https://tickets.metabrainz.org/browse/BB-343

I think those should be done as separate PRs, but they will definitely change things for this current PR.


Regarding your proposal for the rectification icon, the issue is not only the icon itself, but more so the fact that it is not very visible. Changing the "+" into an "x" will not solve that issue, and maybe even make it even less visible.
Would you need to turn the whole button orange or red, Or some other solution?

endurance21:
@MonkeyDo
That means I should,
1. Remove all pop-up feature?.
2.make available all inputs(add disambiguation, add aliases) on same page 
 On which particular entity editor is present. 
That will solve BB343 as well.
Am I getting it right?
And should I make another PR. For all these changes 
And what will happen to current pr?.

endurance21:
sorry , i have deleted the some commit by mistake from master , that's why pr was closed

endurance21:
@MonkeyDo 
 i have done part 1,
deleted disambiguation button , and added disambiguation field by default ,
should i push in this PR or  i have another PR for that ,cause the title of this PR is not matching the work i have done.

MonkeyDo:
Good job on that first PR !

You are correct about the rest: the goal is to move the alias and identifiers popups content into the page.
It might be helpful to do a quick mockup beforehand, to see how everything will fit on the page

endurance21:
@MonkeyDo
I have already made the design, I was waiting for your responses.. 
Here it is
https://xd.adobe.com/view/36b49154-f1ed-4803-4a44-c8897dcca694-4085/

endurance21:
hey  @MonkeyDo  !
what you say about my design ?
https://xd.adobe.com/view/36b49154-f1ed-4803-4a44-c8897dcca694-4085/
should i start building them ?

prabalsingh24:
Don't you think it's better if you just add those two - alias editor and identifier editor - in the current design. I mean what you suggested would involve complete redesign of the website, which is alot of work!? 
In any case have a look at this too. 
https://tickets.metabrainz.org/projects/BB/issues/BB-236

endurance21:
@prabalsingh24
I appreciate your concern about a lot of work! 😅
But what matters here is what we give to the end user.
And for me, making that happen is the goal I decided at the first place!

And for the record, I may include this in my proposal for gsoc, if it is needed to be done.!!! 

@MonkeyDo
I just want to hear, is this design can be a basis to think about applying in UNIFIED CREATION FORM for gsoc?.

MonkeyDo:
@endurance21 At this point I don't think we should do an entire overhaul of the design. That's not a priority, and would require a lot more preparation, discussion, designs, etc.

The mockups I'm more interested in should be based on the existing layout, but modified as we've discussed to move the alias and identifier editors into the main body of the page.

endurance21:
@MonkeyDo
Okay! I get it
I will keep the existing layout!
One more thing!
As I have removed the disambiguation button, should I know change the buttons too?. I mean for the purpose this current pr was made?.
Will have to make another PR?

MonkeyDo:
I think it would be better to do a PR each for the alias editor and for the identifier editor, and abandon this PR, as I feel the problem would be solved another way.

MonkeyDo:
I am closing this in favor of creating two separate PRs for moving both aliases and identifiers in the main flow of the page, as described in ticket https://tickets.metabrainz.org/browse/BB-343

PR #377, opened 2020-02-22
Docker postgres no password error

prabalsingh24:
### Problem
https://github.com/docker-library/postgres/issues/681
Due to a recent change in postgres, an empty password was throwing an error.

### Solution
Modified the environment to allow empty password

### Areas of Impact

 docker-compose.yml

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
       - postgres
     volumes:
       - "./config/config.json:/home/bookbrainz/bookbrainz-site/config/config.json:ro"
-  
+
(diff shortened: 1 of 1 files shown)


prabalsingh24:
I was setting up bookbrainz on a different pc, It took me soo long to figure this out, I thought it was some issue with the ubuntu version first lol. I was so frustated that I could not setup the project LOL. Apparently it's a very recent change.
Another way to fix this would be to change `postgres` version to `9.5.20`. It works that way also. Right now installing `9.5 `installs `9.5.21`

PR #378, opened 2020-02-23
BB-426: fix text overflow on buttons on small screens

endurance21:
### Problem
https://tickets.metabrainz.org/browse/BB-426?filter=-2

### Solution
no more text overflow on device having width below "451px"'
solved-text-overflow

### Areas of Impact
src/client/entity-editor/edition-section/edition-section.js

### DIFF
--- a/src/client/entity-editor/edition-section/edition-section.js
+++ b/src/client/entity-editor/edition-section/edition-section.js
@@ -271,7 +271,6 @@ function EditionSection({
 								<Button
 									block
 									bsStyle="primary"
-									id="existing-edition-group-search-button"
 									onClick={onEditionGroupButtonClick}
 								>
 									Click here to search for an existing one instead
--- a/src/client/entity-editor/edition-section/edition-section.js
+++ b/src/client/entity-editor/edition-section/edition-section.js
@@ -271,6 +271,7 @@ function EditionSection({
 								<Button
 									block
 									bsStyle="primary"
+									id="existing-edition-group-search-button"
 									onClick={onEditionGroupButtonClick}
 								>
 									Click here to search for an existing one instead
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -409,3 +409,10 @@ hr.wide {
 		margin-left: 0.2em;
 	}
 }
+
+@media only screen and (max-width: 451px){
+	#existing-edition-group-search-button{
+		font-size: calc(3px + 1.6vw);
+    padding: 6px 5px;
+	}
+}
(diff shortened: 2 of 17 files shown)


endurance21:
@MonkeyDo
 Is it okay that, I had used I'd css selector rather than making hierarchy of class css selector.?.

src/client/entity-editor/edition-section/edition-section.js
```diff
@@ -271,6 +271,7 @@ function EditionSection({
 								<Button
 									block
 									bsStyle="primary"
+									id="existing-edition-group-search-button"
```
MonkeyDo: I'm not sure about the use of an id here.
It would be much better if the solution was generic and applied to other buttons that could potentially have that issue

src/client/stylesheets/style.less
```diff
@@ -409,3 +409,10 @@ hr.wide {
 		margin-left: 0.2em;
 	}
 }
+
+@media only screen and (max-width: 451px){
+	#existing-edition-group-search-button{
+		font-size: calc(3px + 1.6vw);
```
MonkeyDo: This makes it quite difficult to read. 
I would prefer if the button style was amended to allow a multiline text if there isn't enough space on one line.

endurance21: @MonkeyDo 


ya ! i think  multiline text is good idea .

src/client/entity-editor/edition-section/edition-section.js
endurance21: @MonkeyDo 
i have possibly checked whole website , i didn't
 find this bug anywhere else , that's why i did'nt stressed much about generic solution at the first place .

src/client/stylesheets/style.less
endurance21: @MonkeyDo 
its as simple as that ,
the white-space property is set to "no-wrap"
that allows our text to ovreflow 


I HAVE set the property to "normal"
and wola ! it works!

endurance21:
@MonkeyDo 
done and dusted ! 
have a look!

src/client/stylesheets/style.less
```diff
@@ -409,3 +409,9 @@ hr.wide {
 		margin-left: 0.2em;
 	}
 }
+
+.btn{
+	white-space: normal;
+
+
```
MonkeyDo: Please remove the extra blank lines here

endurance21:
@MonkeyDo
Sorry for the extra lines 😅😅, I am on the wayyyy😁

endurance21:
@MonkeyDo 
done and dusted !

MonkeyDo:
Thanks !

prabalsingh24:
@endurance21 Can I do this?

endurance21:
@prabalsingh24 ya sure.

PR #384, opened 2020-03-08
HTML TITLE FIX

chinmay-kothari:
### Problem

HTML page wasn't showing the title properly

### Solution

I have added titles to the page to make it more readable

### Areas of Impact

HTML code

### DIFF
--- a/src/client/helpers/entity.js
+++ b/src/client/helpers/entity.js
@@ -154,7 +154,7 @@ export function showEntityEditions(entity) {
 	);
 }
 
-export function getEntityLabel(entity, returnHTML = true, unNamedHTML = true) {
+export function getEntityLabel(entity, returnHTML = true) {
 	if (entity.defaultAlias) {
 		return `${entity.defaultAlias.name} `;
 	}
@@ -168,12 +168,12 @@ export function getEntityLabel(entity, returnHTML = true, unNamedHTML = true) {
 		if (returnHTML) {
 			return <span className="text-muted deleted" title={`Deleted ${entity.type}`}>{deletedEntityName}</span>;
 		}
-		return `Deleted ${deletedEntityName}`;
+		return `${deletedEntityName}`;
 	}
-	if (unNamedHTML) {
+	if (returnHTML) {
 		return <span title={`Unnamed ${entity.type} ${entity.bbid}`}>(unnamed)</span>;
 	}
-	return 'unNamed';
+	return 'Unnamed';
 }
 
 export function getEditionReleaseDate(edition) {
--- a/src/client/helpers/entity.js
+++ b/src/client/helpers/entity.js
@@ -154,7 +154,7 @@ export function showEntityEditions(entity) {
 	);
 }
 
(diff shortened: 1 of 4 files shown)


src/server/routes/entity/entity.js
```diff
@@ -161,7 +161,8 @@ export function displayEntity(req: PassportRequest, res: $Response) {
 				markup,
 				page: entityName,
 				props: escapeProps(props),
-				script: '/js/entity/entity.js'
+				script: '/js/entity/entity.js',
+				title: `${getEntityLabel(props.entity)}: ${_.upperFirst(entityName)}`
```
MonkeyDo: If you look in `getEntityLabel`, you'll see that in certain cases it will return an html fragment.
That's not going to work in the title, it'll look like "[object Object] Work" for example.

You could change `getEntityLabel` to add an optional boolean parameter `returnHTML` (defaulting to true) and return a string or html fragment according to that.

src/server/routes/editor.js
```diff
@@ -245,7 +244,8 @@ router.get('/:id', (req, res, next) => {
 				markup,
 				page: 'profile',
 				props: escapeProps(props),
-				script: '/js/editor/editor.js'
+				script: '/js/editor/editor.js',
+				title: `Profile: ${props.editor.name}`
```
MonkeyDo: I think I would prefer `${props.editor.name}'s Profile` instead, and same format for the other two editor routes.

src/server/routes/entity/entity.js
MonkeyDo: I also would prefer the entity name to be in parentheses instead of the semicolon.
Like this "Aldous Huxley (Author)".

What do you think?

src/server/routes/editor.js
chinmay-kothari: I also think writing title that way would be a better choice.

src/server/routes/entity/entity.js
chinmay-kothari: I have created an extra boolean function for deleted entities. And also changed the title slightly as you asked for.

src/client/helpers/entity.js
```diff
@@ -165,10 +165,15 @@ export function getEntityLabel(entity) {
 		if (entity.parentAlias) {
 			deletedEntityName = entity.parentAlias.name;
 		}
-		return <span className="text-muted deleted" title={`Deleted ${entity.type}`}>{deletedEntityName}</span>;
+		if (returnHTML) {
+			return <span className="text-muted deleted" title={`Deleted ${entity.type}`}>{deletedEntityName}</span>;
+		}
+		return `Deleted ${deletedEntityName}`;
```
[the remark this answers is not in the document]
chinmay-kothari: I also think that simply using  ' $(deletedEntityName) '  would be a better option in case of multiple deleted entities tabs.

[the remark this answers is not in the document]
chinmay-kothari: I am committing the said changes, let me know if there's any change is still required.

Issue #51, opened 2015-04-07
Allow using spacebar to select dropdown menu item


johtso:

LordSputnik:
Are you referring to any particular select dropdown, or all of them? I'm not sure how possible this is, or if it's something we can do with select2... @leftmostcat

LordSputnik:
Closing - added to JIRA in http://tickets.musicbrainz.org/browse/BB-143

PR #385, opened 2020-03-09
Refactoring revision components

prabalsingh24:
### Problem
This PR is about having all the functions required for RevisionsPage, EditorRevisions, EntityRevisions inside routes/helper/revisions
And also write unit tests for these functions

### DIFF
--- a/src/server/helpers/utils.js
+++ b/src/server/helpers/utils.js
@@ -51,17 +51,6 @@ export function getEntityModels(orm: Object): Object {
 	};
 }
 
-export function getRevisionModels(orm) {
-	const {AuthorRevision, EditionRevision, EditionGroupRevision, PublisherRevision, WorkRevision} = orm;
-	return [
-		AuthorRevision,
-		EditionGroupRevision,
-		EditionRevision,
-		PublisherRevision,
-		WorkRevision
-	];
-}
-
 export function getDateBeforeDays(days) {
 	const date = new Date();
 	date.setDate(date.getDate() - days);
--- a/src/server/helpers/utils.js
+++ b/src/server/helpers/utils.js
@@ -50,7 +50,8 @@ export function getEntityModels(orm: Object): Object {
 		Work
 	};
 }
-function getRevisionModels(orm) {
+
+export function getRevisionModels(orm) {
 	const {AuthorRevision, EditionRevision, EditionGroupRevision, PublisherRevision, WorkRevision} = orm;
 	return [
 		AuthorRevision,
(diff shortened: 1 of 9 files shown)


prabalsingh24:
Tests are incomplete right now. I am writing more tests for this

src/server/helpers/utils.js
```diff
@@ -50,7 +50,8 @@ export function getEntityModels(orm: Object): Object {
 		Work
 	};
 }
-function getRevisionModels(orm) {
+
+export function getRevisionModels(orm) {
```
MonkeyDo: Shouldn't this function also jump over to src/server/helpers/revisions.js ?

prabalsingh24: I didn't write this function so I just assumed it's being used for other things too. I guess I am wrong, I will have a look and change if that isn't the case.

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,57 @@
...
+			// revisions are same if their revisionId are equal
+			expect(orderedRevisions[i].revisionId).to.be.equal(allRevisions[i + from].revisionId);
+			expect(orderedRevisions[i]).to.have.keys(
+				'authorId',
+				'createdAt',
+				'editor',
+				'entities',
+				'revisionId'
+			);
+		}
```
MonkeyDo: Considering you're already testing for the keys in the test above, let's not redo the check here.

That also allows you to simplify the test and replace the `for` loop with this: 
```
const allRevisionsSubset = allRevisions.slice(from, from + size);
expect(orderedRevisions).to.deep.equal(allRevisionsSubset);
```

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,57 @@
...
+				'createdAt',
+				'editor',
+				'entities',
+				'revisionId'
+			);
+		});
+		expect(orderedRevisions).to.be.descendingBy('createdAt');
+	});
+
+	it('offset, when having higher value, should be working properly ', async () => {
```
MonkeyDo: "it should return the expected subset of revisions when passed an offset (from)"

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,57 @@
...
+chai.use(require('chai-sorted'));
+
+
+describe('getOrderedRevisions', () => {
+	before(async () => {
+		await createMultipleRevisions(1000);
+	});
+	after(truncateEntities);
+
+	it('ordered revisions should be sorted and have the required keys', async () => {
```
MonkeyDo: it('should return sorted revisions with the expected keys'

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,57 @@
...
+				'createdAt',
+				'editor',
+				'entities',
+				'revisionId'
+			);
+		}
+		expect(orderedRevisions).to.be.descendingBy('createdAt');
+	});
+
+	it('ordredRevisions should be empty if offset is higher than total revisions', async () => {
```
MonkeyDo: it('should return no results if offset is higher than total revisions'

prabalsingh24:
I've changed it `before` and `after` to `beforeEach` and `afterEach` as you mentioned. But I didn't get why there could be an interference b/w tests?

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,415 @@
...
+				authorId: editorJSON.id,
+				entities: [],
+				revisionId
+			}
+		];
+		const revisionsWithEntities = await getAssociatedEntityRevisions(formattedRevision, orm);
+
+		expect(revisionsWithEntities.length).to.be.equal(1);
+		expect(revisionsWithEntities[0].entities.length).to.be.equal(1);
+		revisionsWithEntities[0].entities.forEach((entity) => {
```
MonkeyDo, line 131: Instead of a `forEach` here, considering you test the length just above you can use `revisionsWithEntities[0].entities[0].defaultAlias.id` which is a bit clearer.

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,415 @@
...
+				expect(entity.defaultAlias.id).to.be.equal(expectedDefaultAliasId);
+			});
+		});
+	});
+});
+
+describe('getOrderedRevisionForEditorPage', () => {
+	// eslint-disable-next-line one-var
+	let editorJSON, req;
+	beforeEach(async () => {
```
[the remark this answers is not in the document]
prabalsingh24: I've modified one revision in one of the tests. I created anextra revision with multiple notes. I will get interference on the later tests.

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,415 @@
...
+describe('getOrderedRevisions', () => {
+	before(async () => {
+		const editor = await createEditor();
+		const editorJSON = editor.toJSON();
+		const revisionAttribs = {
+			authorId: editorJSON.id,
+			id: 1
+		};
+		const promiseArray = [];
+		for (let id = 1; id <= 1000; id++) {
```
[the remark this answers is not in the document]
prabalsingh24: Umm I made 1000 tests so that the `created_at` column is different in some rows.
I've changed it to 50 revisions and used `fake.date.recent()` to get a recent random date

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,415 @@
...
+		const noteAttrib = {
+			authorId: editorJSON2.id,
+			content: 'note content',
+			id: 1,
+			revisionID
+		};
+
+		// Creating 100 notes for revision#1001
+		const promiseArray = [];
+		for (let id = 1; id <= 100; id++) {
```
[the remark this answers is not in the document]
prabalsingh24: same as above

test/src/server/helpers/revisions.js
```diff
@@ -0,0 +1,418 @@
...
+		await new Revision({
+			authorId: editorJSON2.id,
+			createdAt: date.recent(),
+			id: revisionID
+		}).save(null, {method: 'insert'});
+
+		const noteAttrib = {
+			authorId: editorJSON2.id,
+			content: 'note content',
+			id: 1,
```
MonkeyDo, line 370: Maybe safer to use a random.number() here and in the for loop below? What do you think?

test/src/server/helpers/revisions.js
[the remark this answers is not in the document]
MonkeyDo: > I've modified one revision in one of the tests.

I don't see it. Am I missing something?
You do create a new revision (# 51) with notes, but it's not reused in any other test.

As far as I can tell, you're not testing against any property that you may have changed in another test, so you should be in the clear with `before` (compared to `getAssociatedEntityRevisions` tests where you change the entities' aliases in tests then test against the `defautlAliasId` in following tests).

prabalsingh24:
In the very last test I am assuming there are only 50 Revision, But I created #51. I think I can change the `from` to `52` or a higher value?

prabalsingh24:
or maybe move the test creating Revision#51 to the bottom?

MonkeyDo:
> In the very last test I am assuming there are only 50 Revision, But I created #51. I think I can change the `from` to `52` or a higher value?
>or maybe move the test creating Revision#51 to the bottom?

I see, I did miss that :p

Yes, you can even give yourself a margin of error and test for an offset of 60.
Or moving it to the top would also work.

PR #386, opened 2020-03-10
BB-50: Add Editor activity graphs

prabalsingh24:
### Problem
https://tickets.metabrainz.org/projects/BB/issues/BB-50
This PR adds a new feature(?) which shows number of revisions performed by an editor in a graph

### Solution

Screenshot from 2020-03-10 07-40-46


prabalsingh24:
Tests will most likely fail, because I didn't change package-lock.json, they failed in other PR also because of same reason. 
You asked me to change the package-lock but there were so many changes(933 differences) in that file, and I was feeling reluctant to add that file. And most of them those changes not because of these two package installed. 
Also I was confused between line graph and bar graph. What do you think?
I could've avoided installing 'binary-search-bound' package by writing upper-bound, lower-bound code by myself but I was feeling lazy and chose the easy way :P

MonkeyDo:
> Tests will most likely fail, because I didn't change package-lock.json, they failed in other PR also because of same reason.
> You asked me to change the package-lock but there were so many changes(933 differences) in that file, and I was feeling reluctant to add that file. And most of them those changes not because of these two package installed.

I confirm you do have to commit the package-lock file, or the tests complain, as you've discovered :)
I suppose it's not ideal to have all those changes in the file, but don't worry, it has no ill effects.

src/server/routes/editor.js
```diff
@@ -193,13 +193,47 @@ function getIdEditorJSONPromise(userId, req) {
...
+
+		const revisionsThisMonth = utils.numbersBetweenToAndFrom(revisionDates, thisMonth, nextMonth);
+		const monthName = `${thisMonth.toLocaleString('en-us', {month: 'short'})}-${thisMonth.getFullYear() - 2000}`;
+
+		editorActivity.push({month: monthName, revisions: revisionsThisMonth});
+
+		thisMonth.setMonth(thisMonth.getMonth() + 1);
+	}
+	return editorActivity;
+}
```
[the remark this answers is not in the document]
MonkeyDo: Hmm, that doesn't fill up all the months (for example here there's no entry for 'Feb-12') and the charting library doesn't fill that up automatically.

Here's a succinct solution that works with date-fns to fix that .
It does require to update date-fns, but I think that's a good thing in any case, we're currently using a beta version and we'll need some of the newer features (intervals) soon enough [`npm i --save date-fns@2.10`]:

```
import {eachMonthOfInterval, format} from 'date-fns';
    […]
	const revisionJSON = revisions.toJSON();
	const revisionDates = revisionJSON.map((revision) => format(new Date(revision.createdAt), 'LLL-yy'));
	const revisionsCount = _.countBy(revisionDates);

	const allMonthsInInterval = eachMonthOfInterval({
		end: new Date(revisionJSON[revisionJSON.length - 1].createdAt),
		start: new Date(revisionJSON[0].createdAt)
	  })
		.map(month => format(new Date(month), 'LLL-yy'))
		.reduce((accumulator, month) => {
		  accumulator[month] = 0;
		  return accumulator;
		}, {});

	return {...allMonthsInInterval, ...revisionsCount};
```

Then you can use the result like this in `renderActivityGraph`
```
import {keys, values} from 'lodash';
	[...]
	const months = keys(activityData);
	const numberOfRevisions = values(activityData);
```

prabalsingh24:
Screenshot_2020-04-02 BookBrainz – The Open Book Database

prabalsingh24:
Which graph do you think looks better?

prabalsingh24:
Also I think we had this conversation before also, I forgot what you said, do you think the graph should be for past 1 year ? and an option to change the year from the dropdown?

src/server/routes/editor.js
```diff
@@ -193,13 +193,47 @@ function getIdEditorJSONPromise(userId, req) {
 		});
 }
 
+async function getEditorActivity(editorId, startDate, Revision) {
+	const revisions = await new Revision()
+		.query('where', 'author_id', '=', editorId)
```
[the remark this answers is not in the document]
prabalsingh24: here startDate is the date when user joined BookBrainz. i don't think this is required because all the revisions will be after that date only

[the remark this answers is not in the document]
MonkeyDo: I was thinking about the future and the possibility to implement a date range.
However, let's keep it at that for now; I don't think there's a huge need for more granularity at the moment.

MonkeyDo:
How would you feel about working on adding activity graphs like that but for the overall statistic on the page of the same name?
I'm thinking each entity type in its own dataset but on the same graph (and maybe a 'total'?)

I'm not suggesting you do that now or in priority, but just floating the idea…

prabalsingh24:
I set the column width to `<Col md ={9} > `. When I set it to 12, the graph looked little big on my screen. Is that fine or should I change it to 12?

src/client/components/pages/parts/editor-profile.js
```diff
@@ -188,12 +191,58 @@ class EditorProfileTab extends React.Component {
...
+					hoverBackgroundColor: 'rgba(235,116,59,0.4)',
+					hoverBorderColor: 'rgba(235,116,59,1)',
+					label: 'Revisions'
+				}
+			],
+			labels: months
+		};
+
+		return (
+			<div className="editor-activity-graph">
```
MonkeyDo, line 219: This class can be removed

prabalsingh24:
> How would you feel about working on adding activity graphs like that but for the overall statistic on the page of the same name?
> I'm thinking each entity type in its own dataset but on the same graph (and maybe a 'total'?)
> 
> I'm not suggesting you do that now or in priority, but just floating the idea…

Yeah this seems good. We can maybe have a pie chart too? Representing percentage of each type of entity ?

src/client/components/pages/parts/editor-profile.js
```diff
@@ -188,12 +191,58 @@ class EditorProfileTab extends React.Component {
...
 	render() {
 		return (
 			<Row>
 				<Col md={12}>
 					{this.renderBasicInfo()}
+				</Col>
+				<Col md={12}>
 					{this.renderStats()}
+				</Col>
+				<Col md={9}>
```
MonkeyDo, line 242: I think this size works fine.
I have a suggestion:  Making the stats column 3 wide and putting it after the graph. That way they look like a single row, which makes sense to me as they are related.
What do you think?

FireShot Capture 045 - Monkey's Profile – BookBrainz - localhost

prabalsingh24: Or maybe have stats on the left and the graph to the right?

prabalsingh24: Screenshot_2020-04-08 prabalsingh24's Profile – BookBrainz(1)

src/client/components/pages/parts/editor-profile.js
prabalsingh24: Oh yeah I forgot. :P

src/client/components/pages/parts/editor-profile.js
prabalsingh24: I think this looks better. Without any heading 
Screenshot_2020-04-08 prabalsingh24's Profile – BookBrainz(2)

MonkeyDo:
If you're feeling like it, you can write some unit tests for `getEditorActivity`, but I am going to merge this PR now as I want to see this in the next website update :)

PR #391, opened 2020-03-15
BB-436: Revision Note Newlines fix

prabalsingh24:
…in revision notes

### Problem
https://tickets.metabrainz.org/projects/BB/issues/BB-436
revision note does not respect newlines

### Solution
used `white-space: pre-wrap` to handle white spaces 

Screenshot_2020-03-15 BookBrainz – The Open Book Database(1)
Screenshot_2020-03-15 BookBrainz – The Open Book Database

### Areas of Impact

 src/client/components/pages/revision.js

### DIFF
--- a/src/client/components/pages/revision.js
+++ b/src/client/components/pages/revision.js
@@ -162,8 +162,8 @@ class RevisionPage extends React.Component {
 				<ListGroupItem
 					key={note.id}
 				>
-					<div>
-						<p>{note.content}</p>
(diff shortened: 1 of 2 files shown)


prabalsingh24:
I was able to re-use `revision-note` class we made earlier.
I didn't add `note-author` class for the author in the revision page because it would make the font smaller and I don't think that is required. ( There is enough space in the revision page )

PR #393, opened 2020-03-19
Elasticsearch APIversion fix to 5.6

chinmay-kothari:
### Problem

Elasticsearch does not work as the older version is getiing installed 

### Solution

Changes the elasticsearch api version to .56

### Areas of Impact

Changes 2 config files


prabalsingh24:
Shouldn't this change be made in `test.json` as well? @MonkeyDo

PR #396, opened 2020-03-20
fix(achievement): UI improvement in achievement tab

prabhleenkaur1007:
noticed a little discontinuity under the achievement tab in the profile page. The background behind the update button is little discontinuous which causes a trouble for eyes while scrolling. Just changed some margins to make the content a little more clear.

### Problem

fixing ticket: BB-447

### DIFF
--- a/src/client/components/pages/parts/editor-achievements.js
+++ b/src/client/components/pages/parts/editor-achievements.js
@@ -125,7 +125,9 @@ class EditorAchievementTab extends React.Component {
 						<DragAndDrop name="rank3"/>
 					</div>
 					<div className="form-group">
-						<span>
+						<span style={{
+							marginLeft: '30px'
+						}}>
 							<button
 								className="btn btn-default"
 								type="submit"
@@ -158,6 +160,7 @@ class EditorAchievementTab extends React.Component {
 										const updatedStyle = {
 											...style,
 											background: 'white',
+											borderBottom: '1px solid',
 											flex: '1',
 											marginTop: STICKY_TOP_MARGIN,
 											zIndex: 10
@@ -170,7 +173,7 @@ class EditorAchievementTab extends React.Component {
 									}
 								}
 							</Sticky>
-							<div style={{zIndex: 1}}>
+							<div style={{zIndex: 1, marginLeft: '21px', marginRight: '21px'}}>
 								<div className="h1">Unlocked Achievements</div>
 								{achievements}
 								<div className="h1">Locked Achievements</div>
@@ -197,3 +200,4 @@ EditorAchievementTab.propTypes = {
 };
 
 export default EditorAchievementTab;
+
--- a/src/client/components/pages/parts/editor-achievements.js
+++ b/src/client/components/pages/parts/editor-achievements.js
@@ -125,21 +125,11 @@ class EditorAchievementTab extends React.Component {
 						<DragAndDrop name="rank3"/>
 					</div>
 					<div className="form-group">
-						<span style={{
-							marginLeft: '30px'
-						}}>
-							<button
-								className="btn btn-default"
-								type="submit"
-							>
+						<span style={{ marginLeft: '30px'}}>
+							<button className="btn btn-default" type="submit" >
 								Update
 							</button>
-							<p
-								style={{
-									display: 'inline-block',
-									marginLeft: '10px'
-								}}
-							>
+							<p style={{ display: 'inline-block', marginLeft: '10px'}} >
 								click badge to unset
 							</p>
 						</span>
--- a/src/client/components/pages/parts/editor-achievements.js
+++ b/src/client/components/pages/parts/editor-achievements.js
@@ -173,7 +173,7 @@ class EditorAchievementTab extends React.Component {
 									}
 								}
 							</Sticky>
-							<div style={{zIndex: 1, marginLeft: '21px', marginRight: '21px'}}>
+							<div style={{marginLeft: '21px', marginRight: '21px', zIndex: 1}}>
 								<div className="h1">Unlocked Achievements</div>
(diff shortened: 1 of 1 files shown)


prabhleenkaur1007:
issue: BB-447

MonkeyDo:
Was this closed on purpose @prabhleenkaur1007 ?

prabhleenkaur1007:
I am sorry, just rechecked. It was not intentional. :-)

endurance21:
@prabhleenkaur1007 
 i think making changes using inline style is just less appropriate than making use of  css selectors in bookbrainz-site/src/client/stylesheets/style.less

prabhleenkaur1007:
hi! @endurance21,
I thought the css selectors for the parent tags may not be appropriate, but now i think that adding a new id for each may not harm, and it will not add the unnecessary changes to others.  
Will surely do the required soon. :-))

src/client/components/pages/parts/editor-achievements.js
```diff
@@ -125,19 +125,11 @@ class EditorAchievementTab extends React.Component {
 						<DragAndDrop name="rank3"/>
 					</div>
 					<div className="form-group">
```
MonkeyDo: I've had a good result by removing  the `row` and `form-group` classes  from the preceding div (keeping only `dnd-container`), and deleting this div (moving the contents [the `<button>` and `<p>`] to be direct children of the `<form>`.

If you then add `padding-bottom-1` to the containing `<form>` element, the result is looking quite good !

src/client/components/pages/parts/editor-achievements.js
```diff
@@ -125,19 +125,11 @@ class EditorAchievementTab extends React.Component {
...
-						<span>
-							<button
-								className="btn btn-default"
-								type="submit"
-							>
+						<span style={{marginLeft: '30px'}}>
```
MonkeyDo: With the above change, this margin shouldn't be necessary anymore

src/client/components/pages/parts/editor-achievements.js
```diff
@@ -125,19 +125,11 @@ class EditorAchievementTab extends React.Component {
...
+							<button className="btn btn-default" type="submit">
 								Update
 							</button>
-							<p
-								style={{
-									display: 'inline-block',
-									marginLeft: '10px'
-								}}
-							>
+							<p style={{display: 'inline-block', marginLeft: '10px'}}>
```
MonkeyDo: On top of the modification above (moving `<button>` and `<p>` one level up), you can change that `<p>` into
`<span className="margin-left-1">….</span>`

src/client/components/pages/parts/editor-achievements.js
```diff
@@ -158,6 +150,7 @@ class EditorAchievementTab extends React.Component {
 										const updatedStyle = {
 											...style,
 											background: 'white',
+											borderBottom: '1px solid',
```
MonkeyDo: I find the black border a bit harsh; maybe you could give it the same border color as the surrounding badges (#ebe2df), or some other color?

src/client/components/pages/parts/editor-achievements.js
```diff
@@ -125,19 +125,11 @@ class EditorAchievementTab extends React.Component {
 						<DragAndDrop name="rank3"/>
 					</div>
 					<div className="form-group">
-						<span>
-							<button
-								className="btn btn-default"
-								type="submit"
-							>
+						<span style={{marginLeft: '30px'}}>
+							<button className="btn btn-default" type="submit">
```
MonkeyDo: I think this would benefit from being a `btn-primary`.

src/client/components/pages/parts/editor-achievements.js
prabhleenkaur1007: Thanks a lot! I tried these, and it is indeed good.
Will surely add in a few hours

src/client/components/pages/parts/editor-achievements.js
prabhleenkaur1007: I tried this, but then the button  would just be a little more left than the above container. So, if you want to make it completely left align, than i can add a margin of just 2-3 px to make sure it is in absolute line. :-)

src/client/components/pages/parts/editor-achievements.js
prabhleenkaur1007: I don't think that it look goods with the change you suggested.

image

src/client/components/pages/parts/editor-achievements.js
prabhleenkaur1007: #ebe2df looks subtle and good. Thanks!

src/client/components/pages/parts/editor-achievements.js
prabhleenkaur1007: well... `btn-default` goes very well the color-theme of bookbrainz (brown, orange).
But if we want to highlight it here, we can go with `btn-success`, same as the search button.
Let me know.

src/client/components/pages/parts/editor-achievements.js
MonkeyDo: It looks to me like the class `margin-left-1` is missing on "click badge to unset"

To clarify, what I meant was `<span className="margin-left-1">click badge to unset</span>` after the button.

src/client/components/pages/parts/editor-achievements.js
MonkeyDo: Sure, let's make it fit perfectly !

src/client/components/pages/parts/editor-achievements.js
MonkeyDo: Yes `btn-success` is the right choice.

PR #397, opened 2020-03-23
BB-435 (test): Allow creation of multiple test entities

MonkeyDo:
### Problem
Errors are thrown in various places in create-entities test helpers when trying to create multiple test entities because of duplicate IDs.

### Solution
Instead of a static id, create a random ID every time where needed.

Also cleaned up unused function;

### Areas of Impact
test/test-helpers/create-entities.js

### DIFF
--- a/test/test-helpers/create-entities.js
+++ b/test/test-helpers/create-entities.js
@@ -252,8 +252,9 @@ export async function createEdition(optionalBBID) {
 
 	await new Entity({bbid, type: 'Edition'})
 		.save(null, {method: 'insert'});
-	await new Edition({...entityAttribs, bbid})
+	const edition = await new Edition({...entityAttribs, bbid})
 		.save(null, {method: 'insert'});
+	return edition;
 }
 
 export async function createWork(optionalBBID) {
@@ -270,8 +271,9 @@ export async function createWork(optionalBBID) {
 	};
 	await new WorkType({id: workAttribs.typeId, label: `Work Type ${workAttribs.typeId}`})
 		.save(null, {method: 'insert'});
-	await new Work({...entityAttribs, ...workAttribs})
+	const work = await new Work({...entityAttribs, ...workAttribs})
 		.save(null, {method: 'insert'});
+	return work;
 }
 
 export async function createEditionGroup(optionalBBID) {
@@ -285,8 +287,9 @@ export async function createEditionGroup(optionalBBID) {
 		.save(null, {method: 'insert'});
 	await new Entity({bbid, type: 'EditionGroup'})
 		.save(null, {method: 'insert'});
-	await new EditionGroup({...entityAttribs, ...editionGroupAttribs})
+	const editionGroup = await new EditionGroup({...entityAttribs, ...editionGroupAttribs})
 		.save(null, {method: 'insert'});
+	return editionGroup;
 }
 
 export async function createAuthor(optionalBBID) {
@@ -313,8 +316,9 @@ export async function createAuthor(optionalBBID) {
 		.save(null, {method: 'insert'});
 	await new Entity({bbid, type: 'Author'})
 		.save(null, {method: 'insert'});
-	await new Author({...entityAttribs, ...authorAttribs})
+	const author = await new Author({...entityAttribs, ...authorAttribs})
 		.save(null, {method: 'insert'});
+	return author;
 }
 
 export async function createPublisher(optionalBBID) {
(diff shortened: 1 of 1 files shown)


MonkeyDo:
@prabalsingh24 I finally got up to making a PR for this code.
Care to have a look and try it out, see if it works for you?

prabalsingh24:
> @prabalsingh24 I finally got up to making a PR for this code.
> Care to have a look and try it out, see if it works for you?

Wow, using random number. Very clever solution I must say!  Thanks. I will have a look at this.

test/test-helpers/create-entities.js
```diff
@@ -242,15 +261,14 @@ export async function createWork(optionalBBID) {
...
 	const languageSetId = await createLanguageSet();
 
 	const workAttribs = {
 		bbid,
 		languageSetId,
-		typeId: setData.id
+		typeId: random.number()
 	};
-	await new WorkType({...setData, label: 'Work Type 1'})
+	await new WorkType({id: workAttribs.typeId, label: 'Work Type 1'})
```
prabalsingh24: `label` has to be unique too. Maybe replace this with ```  label: `Work Type ${workAttribs.typeId}` ```

test/test-helpers/create-entities.js
```diff
@@ -274,23 +292,24 @@ export async function createEditionGroup(optionalBBID) {
...
-		genderId: setData.id,
-		typeId: setData.id
+		genderId: editorAttribs.genderId,
+		typeId: random.number()
 	};
-	await new Area({...setData, gid: uuidv4(), name: 'Rlyeh'})
+	await new Area({gid: uuidv4(), id: areaId, name: 'Rlyeh'})
 		.save(null, {method: 'insert'});
-	await new AuthorType({...setData, label: 'Author Type 1'})
+	await new AuthorType({id: authorAttribs.typeId, label: 'Author Type 1'})
```
prabalsingh24: Here also ``` label: `Author Type ${authorAttribs.typeId}` ```

test/test-helpers/create-entities.js
```diff
@@ -311,11 +330,11 @@ export async function createPublisher(optionalBBID) {
...
 		endYear: 2012,
 		ended: true,
-		typeId: setData.id
+		typeId: random.number()
 	};
-	await new Area({...setData, gid: uuidv4(), name: 'Rlyeh'})
+	await new Area({gid: uuidv4(), id: publisherAttribs.areaId, name: 'Rlyeh'})
 		.save(null, {method: 'insert'});
-	await new PublisherType({...setData, label: 'Publisher Type 1'})
+	await new PublisherType({id: publisherAttribs.typeId, label: 'Publisher Type 1'})
```
prabalsingh24: ``` label: `Publisher Type ${publisherAttribs.typeId} ```

test/test-helpers/create-entities.js
```diff
@@ -261,9 +279,9 @@ export async function createEditionGroup(optionalBBID) {
 	await createEntityPrerequisites();
 	const editionGroupAttribs = {
 		bbid,
-		typeId: setData.id
+		typeId: random.number()
 	};
-	await new EditionGroupType({...setData, label: 'Edition Group Type 1'})
+	await new EditionGroupType({id: editionGroupAttribs.typeId, label: 'Edition Group Type 1'})
```
prabalsingh24: Replace `label` with ``` label: `Edition Group Type ${editionGroupAttribs.typeId}` ```

test/test-helpers/create-entities.js
```diff
@@ -242,15 +261,14 @@ export async function createWork(optionalBBID) {
...
 	const workAttribs = {
 		bbid,
 		languageSetId,
-		typeId: setData.id
+		typeId: random.number()
 	};
-	await new WorkType({...setData, label: 'Work Type 1'})
+	await new WorkType({id: workAttribs.typeId, label: `Work Type ${workAttribs.typeId}`})
 		.save(null, {method: 'insert'});
 	await new Work({...entityAttribs, ...workAttribs})
```
prabalsingh24:
```
return new Work({...entityAttribs, ...workAttribs})
		.save(null, {method: 'insert'}); 
```

MonkeyDo: I don't think that's a bad idea at all, and could be useful i some cases.

MonkeyDo:
@prabalsingh24
Thanks for the suggestiong.
I want to point out that in some cases, it'll be better to fetch the newly created entity to get its data rather that use the entity returned by createWork, createAuthor, etc.
That way you ensure that what you expect is in the database.
There are cases however where we might need to compare attributes between the created entity and the one fetched from the database after creation, so it's useful to return it.

Anything else you think might be useful or missing?

prabalsingh24:
Yeah thanks :D
No this is it I guess. 
I am able to make changes I want using this. I should be able to write tests without any problem now (hopefully) xD

PR #398, opened 2020-03-29
Fix(aliases): add legal and pseudonym label in aliases

prabhleenkaur1007:
BB-284 

Earlier there was no way that a user can know about the legal or pseudonym of an alias

### Solution

I changed the primary label to legal . If the checkbox is checked then append "(legal)" with the name, otherwise "(pseudonym)".


MonkeyDo:
This won't be solve that easily; as I commented above the terminology only works for Authors, but the components you modified are used for all entities.
Should I close this PR for now?

prabhleenkaur1007:
Yeah i got your point... Yes, you can close for now.

PR #400, opened 2020-03-30
fix(achievement): unnecessary badge GSoC 2016

prabhleenkaur1007:
I tried to solve issue #BB-445 by deleting rows from the database using SQL statements in bookbrainz.sql :
>DELETE FROM bookbrainz.achievement_type WHERE id =27;
>DELETE FROM bookbrainz.achievement_unlock WHERE achievement_id =27;
 but got no results.
Hence, i set the condition to return null when id =27 (for achievement badge GSoC 2016), for time being.


MonkeyDo:
The best way to do it would be in a new .sql file in its own folder in `sql/migrations/`

PR #402, opened 2020-04-01
Invalid bbid error

prabalsingh24:
### Problem
https://beta.bookbrainz.org/work/123123
It'd be better if this page shows `Work not found error` rather than `Page not found`

### Solution

Throw error even when `bbid` is invalid

### Areas of Impact

 src/server/helpers/middleware.js


prabalsingh24:
Sure I'll have a look at this

prabalsingh24:
I created a error class in [342](https://github.com/bookbrainz/bookbrainz-site/pull/342/files), I realized it wasn't really needed, could've used class `FormSubmissionError` there. 
Renamed that to `BadRequest` class which can be used for other errors too.

PR #404, opened 2020-04-07
Fix entity /create route

MonkeyDo:
### Problem
Following PR #402 , the create route for entities (i.e: `/work/create`) throws an "Invalid BBID" error

### Solution
Modify the entity loading middleware to accept the `/create` route path, since we have no good way to differentiate a mistyped BBID and a valid route. 

Also added basic tests for the entity routes to make sure we don't break them as easily in the future.

### DIFF
--- a/src/server/helpers/middleware.js
+++ b/src/server/helpers/middleware.js
@@ -139,7 +139,10 @@ export function makeEntityLoader(modelName, additionalRels, errMessage) {
 
 	return async (req, res, next, bbid) => {
 		const {orm} = req.app.locals;
-		if (commonUtils.isValidBBID(bbid)) {
+		if (req.path === '/create') {
+			return next('route');
+		}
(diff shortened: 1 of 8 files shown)


src/server/helpers/middleware.js
```diff
@@ -139,7 +139,10 @@ export function makeEntityLoader(modelName, additionalRels, errMessage) {
 
 	return async (req, res, next, bbid) => {
 		const {orm} = req.app.locals;
-		if (commonUtils.isValidBBID(bbid)) {
+		if (req.path === '/create') {
```
prabalsingh24, line 142: wouldn't it be better if the path is first converted to small letters? I think /work/CREATE will fail here

MonkeyDo: Good point. i'll modify it tomorrow.
I'm currently struggling with how to mock user authentication to test the /create and /edit routes…

PR #405, opened 2020-04-08
feature(navbar): Added menu item pointing to revisions page

MonkeyDo:
### Problem
The revisions page is quite an important new page, and it should be accessible on every page.

### Solution
Add new menu item to navbar

Capture d’écran 2020-04-08 à 14 37 34

### DIFF
--- a/src/client/containers/layout.js
+++ b/src/client/containers/layout.js
@@ -193,6 +193,12 @@ class Layout extends React.Component {
 						{' Statistics '}
 					</NavItem>
 				</Nav>
+				<Nav pullRight>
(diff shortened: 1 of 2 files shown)


MonkeyDo:
@prabalsingh24 Any thoughts on this?
I wanted the page to be more prominent, and not having to know about that button at the very bottom of the homepage…

prabalsingh24:
> @prabalsingh24 Any thoughts on this?
> I wanted the page to be more prominent, and not having to know about that button at the very bottom of the homepage…

Looks niceee :D

PR #408, opened 2020-04-09
Add test case for routes

chinmay-kothari:
### Problem

Unit test case

### Solution

Add tests to check routes in entity,editor, revision etc.

### Areas of Impact

test directory


prabalsingh24:
@MonkeyDo  I think tests failed because of `random.number()` giving same number...

MonkeyDo:
> @MonkeyDo I think tests failed because of `random.number()` giving same number...

Yes, it looks like it. Oddly enough, it's the same test that failed for me (creating EG data for `getOrderedRevisionsForEntityPage` tests.
In any case, I think we'll have to have a global id counter that we use and increment everywhere an id is needed…

chinmay-kothari:
> While these tests are definitely a good starting point, i would love to see more elaborate testing for each route, which can be achieved by setting up an integration testing and/or end-to-end testing.
> Here's a useful read on the subject: https://medium.com/expedia-group-tech/integration-testing-in-react-21f92a55a894

Thanks for the suggestions. I also think doing it with end-to-end testing would be a good option. Though I need to learn more about it :). Thanks for the read it helps.

MonkeyDo:
So the goal of these end to end tests is to ensure that the pages are working properly. That is indeed very vague, but by visiting each page you can think about what you’re expecting to see, and how to test it.
The goal is to instruct a headless browser to do the same inspection we would do to check everything is working fine: target an element on the page, and compare the content to what I am expecting to see there; compare the page title to what I'm expecting

Here are a few examples:

Editor page:
For that page, you’ll want to create an Editor with the relevant info so that you can visit their page.
I should see the information about the editor ("Basic info”): MB name, display name, area, etc.
Also in the “stats” section I should see a number of total revisions (you’ll need to create some revisions beforehand with that editor as the revision author)
Let’s skip the badges section and achievements tab for now, they’re not critical.
However, visiting the revisions tab, I should see a table with X revisions. Each revision should have the appropriate info (id, entity, note, date.)

Revisions page:
With the tests that we did before, we created some revisions to test the editor page.
Which means  we can visit the /revisions page and expect to see a table with the same revisions that we tested for (we can reuse some test code here, which is handy)
You’ll want to create another editor and more revisions (let’s say 25). Than way we can test that the user column shows the right user.
Now, you can test if clicking on the “next” pagination button will show the following revisions (21 to 25).
You can also test if visiting the page with /revisions?size=30 will show 25 results in the table.

For the entities pages, I’ll take the example of Author pages but the same applies for each entity type.
You’ll need to create an Author with entity with multiple revisions.
With that, you can test that the /author page show the correct information for that entity, and that clicking the buttons at the bottom will bring you the the relevant pages (/author/revisions and /author/delete).
As above, the /author/revisions page should show a table with the relevant revisions. No need to test pagination here (clicking on next, etc.) as we did it above for the editor page (the code is the same)
The page title should also reflect the name fo the author.

The entity edit pages we can talk about another day, they’ll be a whole chunk by themselves and introduce the complexity of being signed in.

I hope this helps in giving a general direction. Don't hesitate if you have more specific questions.
You might still want to read more examples of end-to-end tests to get a better idea of what people usually do. I don't have a good example at hand but I'm sure you'll be able to find plenty.

chinmay-kothari:
Thanks for explaining the whole blueprint of things. It certainly clears the aim to use end to end testing for. Nice explained :)

MonkeyDo:
I'll also add that an invaluable tool when developing and also when testing is a debugger.
While you run your code, you're able to stop execution where you want it, inspect what the values are, and walk through your code everywhere.
If you're not very familiar with it yet I highly recommend you read this intro: https://developers.google.com/web/tools/chrome-devtools/javascript/

The BookBrainz repo is set up to easily work with the editor VSCode that includes a debugger. https://code.visualstudio.com/Docs/editor/debugging
To run a single test file in debug mode you can set the file path here then launch the "Webserver Mocha Tests" configuration: https://github.com/bookbrainz/bookbrainz-site/blob/master/.vscode/launch.json#L25-L26

MonkeyDo:
I've recently made some changes to the test files and helpers, so that I could test routes properly.
There is now a way to create a fake editor and log them in, so that we can test auth is working.

If you're interested in picking up this task again, I can help you get set up again.

If you're not interested, that's not an issue, but please let me know so that I can close the PR :)

chinmay-kothari:
> I've recently made some changes to the test files and helpers, so that I could test routes properly.
> There is now a way to create a fake editor and log them in, so that we can test auth is working.
> 
> If you're interested in picking up this task again, I can help you get set up again.
> 
> If you're not interested, that's not an issue, but please let me know so that I can close the PR :)

I am interested in picking it up again!
Recently I have been very caught up with my academic so couldn't get time for it. But sure I am interested in starting it again :)

MonkeyDo:
> I am interested in picking it up again!
> Recently I have been very caught up with my academic so couldn't get time for it. But sure I am interested in starting it again :)

I completely understand, it's complicated times !
I'm glad you're still interested in it, that's great ! :)

I'll leave the PR open; there's no rush.
When you have the time, you can have a look at the tests of the /merge routes here, they'll give a decent idea of how I've tested those routes:
https://github.com/bookbrainz/bookbrainz-site/blob/master/test/src/server/routes/merge.js

The mechanism for logging in as fake editor "123456" is like this: https://github.com/bookbrainz/bookbrainz-site/blob/master/test/src/server/routes/merge.js#L19-L40
1. create editor 123456
2. GET the `/cb` route to log in the fake editor
3. re-use the chai-http `agent` we created (think of it like an instance of the app where the editor is logged in)
4. after all tests, close the agent.


I started implementing that in the entity routes too, but didn't go very far at all…

chinmay-kothari:
Hey @MonkeyDo 
Recently I have been very caught up with some work not getting time to work on this PR. I'll be closing this PR for the same.

Though I'm working on generating automated JSDoc documentation. Will make a PR soon. Thanks

PR #409, opened 2020-04-09
GetEditorActivity Test

chinmay-kothari:
### Problem

Test case for GetEditorActivity function

### Solution

Test

### Areas of Impact

Test dir


test/src/server/routes/editor.js
```diff
@@ -0,0 +1,63 @@
...
+		}
+		await Promise.all(RevisionArray);
+	});
+	after(truncateEntities);
+
+	it('should return values for plotting graph', async () => {
+		const startDate = new Date();
+		startDate.setFullYear(2019, 1, 1);
+		const ActivityData = await getEditorActivity(editorJSON.id, startDate, Revision);
+		const MonthArray = ['Jan-19', 'Feb-19', 'Mar-19', 'Apr-19', 'May-19', 'Apr-19', 'May-19',
```
MonkeyDo, line 42: There's an extra `'Apr-19', 'May-19'` here, which points to an issue with the test: it doesn't test what's returned.

You can create an object with exactly the values you expect to get back from the utility, then use `expect(ActivityData).to.deep.equal(myExpectedResultsObject)`.
That way, you can test at the same time both if you have the right number of results *and* if you have one per month *and* if the month string and activityData are formatted correctly.

Also, by convention, variables and functions names use camelCase notation, while starting with an uppercase is used to represent class constructors, types and a few other special cases.

chinmay-kothari:
@MonkeyDo 
I have added another test as you mentioned for the condition when some months would be having multiple revisions while some would be with zero.

I made an `expectedResultObject` which mocks how we expect the result to be. Then compared it with the returned value. 

Let me know what you think about the changes :)

test/src/server/routes/editor.js
```diff
@@ -35,29 +36,83 @@ describe('getEditorActivity', () => {
...
+			// eslint-disable-next-line sort-keys
+			'Jan-19': 1, 'Feb-19': 1, 'Mar-19': 1, 'Apr-19': 1, 'May-19': 1, 'Jun-19': 1, 'Jul-19': 1,
+			// eslint-disable-next-line sort-keys
+			'Aug-19': 1, 'Sep-19': 1, 'Oct-19': 1, 'Nov-19': 1, 'Dec-19': 1
+		};
+
+		// Adds remaining months upto the present month
+		const expectedStartDate = new Date();
+		expectedStartDate.setFullYear(2020, 0, 1);
+		const monthsUptoNow = eachMonthOfInterval({
```
prabalsingh24, line 55: Don't do like this. This will always give correct result because this is exactly the way it's been fetched in the `getEditorActivity` function.
Rather create an `expectedResult` object with hard coded values

test/src/server/routes/editor.js
```diff
@@ -35,29 +36,83 @@ describe('getEditorActivity', () => {
...
+			// eslint-disable-next-line sort-keys
+			'Jan-19': 2, 'Feb-19': 1, 'Mar-19': 2, 'Apr-19': 1, 'May-19': 2, 'Jun-19': 1, 'Jul-19': 2,
+			// eslint-disable-next-line sort-keys
+			'Aug-19': 1, 'Sep-19': 2, 'Oct-19': 1, 'Nov-19': 2, 'Dec-19': 1, 'Jan-20': 1, 'Feb-20': 0, 'Mar-20': 1
+		};
+
+		// Adds remaining months upto the present month
+		const expectedStartDate = new Date();
+		expectedStartDate.setFullYear(2020, 3, 1);
+		const monthsUptoNow = eachMonthOfInterval({
```
prabalsingh24, line 103: Same as above

test/src/server/routes/editor.js
```diff
@@ -0,0 +1,118 @@
...
+		for (let i = 0; i < 15; i += 2) {
+			const tempDate = new Date();
+			tempDate.setFullYear(2019, i, 1);
+			revisionAttribs.id = random.number();
+			revisionAttribs.createdAt = tempDate;
+			RevisionArray.push(
+				new Revision(revisionAttribs).save(null, {method: 'insert'})
+			);
+		}
+		await Promise.all(RevisionArray);
```
prabalsingh24, line 86: RevisionArray => revisionsArray (Camel Case)

test/src/server/routes/editor.js
```diff
@@ -0,0 +1,118 @@
...
+
+
+chai.use(chaiHttp);
+const {expect} = chai;
+
+
+describe('getEditorActivity', () => {
+	const {Revision} = orm;
+	let editorJSON;
+	const RevisionArray = [];
```
prabalsingh24, line 18: Instead of having a globally defined array, you can create a local array where it is used. Reusing the same array multiple times might cause some interferance

prabalsingh24: Also camel case.. and this is storing promises so maybe `revisionsPromiseArray` sounds better?

test/src/server/routes/editor.js
chinmay-kothari: The problem with creating hardcoded values is that our `getEditorActivity` function returns values till the present date.

So suppose if I do hardcode it till `Apr-20` then next month we'll need to add `May-20` so that the test does not fail next month. I personally don't think hard coding the values would be a good option. 

What else we can do here is pass an `endDate` into the `getEditorActivity` function default to the `Date.now()`. Then I think we can hardcode here with a precise end date. What do you think?

prabalsingh24: Oh yeah I didn't think of this. Nice catch!
Yeah I think passing an endDate would better.

chinmay-kothari:
@prabalsingh24 
I have passed `endDate` into the `GetEditorActivity` with a default value of the present date. Also did the changes you mentioned. Let me know how it looks now :)

prabalsingh24:
As @MonkeyDo mentioned, Test 1 and 2 can be merged.
Also how about having a test with an editor having 0 revisions? Just to make sure the function don't break in such cases

test/src/server/routes/editor.js
MonkeyDo: I agree, passing and endDate is the way to go, not only just for the purpose of testing but also for more granular use ("fetch my activity from january of this year")

src/server/routes/editor.js
```diff
@@ -207,7 +207,7 @@ async function getEditorActivity(editorId, startDate, Revision) {
 	const revisionsCount = _.countBy(revisionDates);
 
 	const allMonthsInInterval = eachMonthOfInterval({
-		end: Date.now(),
+		end: endDate,
```
MonkeyDo, line 210: Addition of endDate is great.
I suggest one change that you can implement using TDD (test-driven development): first create a new test where you pass a string "bob" (or foo or bar or whatever :p) as the endDate.
I'm not sure what will happen with the current code, but you'll find out :)
Then, modify the test to expect it to throw a helpful error if the date is invalid.
Then, modify `getEditorActivity ` to act accordingly.
Finally, run the test again to ensure it's all working as expected.

It would be a good idea to do the same for startDate, now that I think of it.

You can use date-fns utilities to ensure the date is valid: https://date-fns.org/v2.12.0/docs/isValid#examples

chinmay-kothari:
> As @MonkeyDo mentioned, Test 1 and 2 can be merged.
> Also how about having a test with an editor having 0 revisions? Just to make sure the function don't break in such cases

I guess he said that for the previous two tests that were made, I have already merged those two. I also think having a test for zero revisions would be useful.

test/src/server/routes/editor.js
```diff
@@ -0,0 +1,83 @@
...
+		for (let i = 0; i < 12; i += 2) {
+			const tempDate = new Date();
+			tempDate.setFullYear(2020, i, 1);
+			revisionAttribs.id = random.number();
+			revisionAttribs.createdAt = tempDate;
+			revisionsPromiseArray.push(
+				new Revision(revisionAttribs).save(null, {method: 'insert'})
+			);
+		}
+		await Promise.all(revisionsPromiseArray);
```
MonkeyDo, line 66: If you modify fixtures created in `before` in a test, it's best to make sure it can't affect the tests that are run after it.
The easiest way to do that is to recreate the fixtures before each test by changing `before` and `after` functions to `beforeEach` and `afterEach`.

That way, for the next test you write, you don't have to keep track of the changes you might have made across the previous tests to know exactly what to test against. More deterministic is always better for tests.

chinmay-kothari:
@MonkeyDo,
For invalid dates, I have added error to be thrown and added their respective tests for checking the same. 
Let me know what you think of it. :)

chinmay-kothari:
> @MonkeyDo,
> For invalid dates, I have added errors to be thrown and added their respective tests for checking the same.
> Let me know what you think of it. :)

I guess it failed because of `random.number()` giving the same number..

test/src/server/routes/editor.js
```diff
@@ -0,0 +1,142 @@
...
+			'Jan-20': 1, 'Feb-20': 0, 'Mar-20': 1, 'Apr-20': 0, 'May-20': 1, 'Jun-20': 0, 'Jul-20': 1,
+			// eslint-disable-next-line sort-keys
+			'Aug-20': 0, 'Sep-20': 1, 'Oct-20': 0, 'Nov-20': 1, 'Dec-20': 0
+		};
+
+		expect(activity).to.deep.equal(expectedResult);
+	});
+
+	it('should throw an error for an invalid startDate', async () => {
+		const startDate = parseISO('bob');
```
MonkeyDo, line 103: parseIso will return specifically an Invalid Date object in this case.
Although it will not change the test or its results, I think startDate should be simply a string ("bob")

test/src/server/routes/editor.js
```diff
@@ -0,0 +1,142 @@
...
+		catch (err) {
+			const expectedError = new Error('Start date is invalid');
+			expect(err.message).to.equal(expectedError.message);
+		}
+	});
+
+	it('should throw an error for an invalid endDate', async () => {
+		const startDate = new Date();
+		startDate.setFullYear(2020, 0, 1);
+		const endDate = parseISO('bob');
```
MonkeyDo, line 118: Same as above, let's keep it simple and pass a string

test/src/server/routes/editor.js
```diff
@@ -0,0 +1,142 @@
...
+		await Promise.all(revisionsPromiseArray);
+
+		const startDate = new Date();
+		startDate.setFullYear(2020, 0, 1);
+		const endDate = new Date();
+		endDate.setFullYear(2020, 5, 1);
+		const activity = await getEditorActivity(editorJSON.id, startDate, Revision, endDate);
+
+		const expectedResult = {
+			// eslint-disable-next-line sort-keys
```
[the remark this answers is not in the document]
chinmay-kothari: It would certainly clean up the code very well. Thanks

chinmay-kothari:
@MonkeyDo,
 I have done the changes you mentioned. Added `afterEach`, removed the comments, and passed the dates simply as a string. :p

MonkeyDo:
🎉 🚀

PR #410, opened 2020-04-10
BB-444: Revision Display Page Improvement (WIP)

prabalsingh24:
WIP
PR not ready for review rn

### DIFF
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -16,7 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-
+import * as auth from '../helpers/auth';
 import * as baseFormatter from '../helpers/diffFormatters/base';
 import * as entityFormatter from '../helpers/diffFormatters/entity';
 import * as entityRoutes from './entity/entity';
@@ -169,12 +169,14 @@ function diffRevisionsWithParents(revisions) {
 				.then(
 					(parent) => Promise.props({
 						changes: revision.diff(parent),
-						entity: revision.related('entity')
+						entity: revision.related('entity'),
+						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']})
 					}),
 					// If calling .parent() is rejected (no parent rev), we still want to go ahead without the parent
 					() => Promise.props({
 						changes: revision.diff(null),
-						entity: revision.related('entity')
+						entity: revision.related('entity'),
+						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']})
 					})
 				)
 	));
@@ -223,7 +225,6 @@ router.get('/:id', (req, res, next) => {
 	const editionGroupDiffsPromise = _createRevision(EditionGroupRevision);
 	const publisherDiffsPromise = _createRevision(PublisherRevision);
 	const workDiffsPromise = _createRevision(WorkRevision);
-
 	Promise.join(
 		revisionPromise, authorDiffsPromise, editionDiffsPromise,
 		workDiffsPromise, publisherDiffsPromise, editionGroupDiffsPromise,
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -16,7 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-import * as auth from '../helpers/auth';
+
 import * as baseFormatter from '../helpers/diffFormatters/base';
 import * as entityFormatter from '../helpers/diffFormatters/entity';
 import * as entityRoutes from './entity/entity';
@@ -170,13 +170,15 @@ function diffRevisionsWithParents(revisions) {
 					(parent) => Promise.props({
 						changes: revision.diff(parent),
 						entity: revision.related('entity'),
-						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']})
+						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']}),
+						isNew: !parent
 					}),
 					// If calling .parent() is rejected (no parent rev), we still want to go ahead without the parent
 					() => Promise.props({
 						changes: revision.diff(null),
 						entity: revision.related('entity'),
-						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']})
+						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']}),
+						isNew: true
 					})
 				)
 	));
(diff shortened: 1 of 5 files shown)


prabalsingh24:
This isn't ready for a review right now. 

What other changes should be made @MonkeyDo ?

Screenshot from 2020-04-10 15-43-34

I think the way relationship is shown here is not very good. I added four relationships to "God Emperor of Dune" and the bbids in the table are of this work itself. I think it would be better if those bbids were of the other entity whom which this work is related to.

Even though "Public Affairs" and "Wonder Woman" were revised, their changes are not shown ( a bug maybe).

What do you think?

chinmay-kothari:
@prabalsingh24 
Writing AliasName there may not look good when we have long AliasNames like: https://beta.bookbrainz.org/edition/4e4788aa-b139-47e3-8674-dad226ddc9aa

What do you think?

MonkeyDo:
It definitely looks better. Maybe it would be good to also show the BBID? maybe under the title in a `<small>` tag?

I agree the way relationships are displayed is terrible. There's lots of low hanging fruits to make that page better.
To display relationships, in an ideal world I would love something like the diagram in the [help page](https://bookbrainz.org/help), but that might be for another day.

As for @chinmay-kothari 's concern about long names, you'll have to make sure the name wraps correctly, but I don't think we have many other options.

MonkeyDo:
> Even though "Public Affairs" and "Wonder Woman" were revised, their changes are not shown ( a bug maybe).

Possibly. How did you make changes to the relationships of God Emperor of Dune as well as Public Affairs and Wonder Woman at the same time?
In any case, it looks like it would be worth investigating.

prabalsingh24:
> > Even though "Public Affairs" and "Wonder Woman" were revised, their changes are not shown ( a bug maybe).
> 
> Possibly. How did you make changes to the relationships of God Emperor of Dune as well as Public Affairs and Wonder Woman at the same time?
> In any case, it looks like it would be worth investigating.

Screenshot_2020-04-10 BookBrainz – The Open Book Database

I think it's a bug. Here in this revision. I added the relationship - "New Author" wrote "New Work2" and it didn't show up in both. 
I am having a look

prabalsingh24:
I was editing "God Emperor of Dune" and I kept adding relationships.

prabalsingh24:
`formatNewRelationshipSet` function will be used when you add relationships to an entity w/o any previous relationships.

prabalsingh24:
The `formatRemoveRelationships` function isn't perfect yet...
It's having bug in following case.
Let's say the old relationship set is having following `ids``[100 , 200, 300, 400, 500, 600]` and the new relationship set is `[100, 200, 400, 500, 600]` 
Deep-Diff library which is used to find the changes will show that elements at 2nd, 3rd, 4th index are edited. and the element in 5th index (600) is deleted but in reality element 300 was deleted..

The `formatAddRelationships` is working fine because every time a new relationship is created, it's relationshipId will always greater than the older ids (I think?) and because of it, it will always be at the last index ( because we are sorting before finding the diffs)

prabalsingh24:
Screenshot_2020-04-13 BookBrainz – The Open Book Database

Instead of showing bbid of the entity itself, I changed it to show bbid of the other entity. the one it's related to. (check table)
I made a mistake in column name, I am gonna fix it in next commit. Don't wanna create an extra commit for a typo :(

prabalsingh24:
@MonkeyDo Tagging you because the option to ask for a review is not showing up :P

prabalsingh24:
Screenshot_2020-04-16 BookBrainz – The Open Book Database



In this revision, I removed `MusicBrainz ID`, but this shows that I removed `WikiData` and then 
added `WikiData` inplace of `MusicBrainz ID`.

`Deep-Diff` isn't working properly. Should it be replaced w some different library?

MonkeyDo:
> Screenshot_2020-04-16 BookBrainz – The Open Book Database
> 
> In this revision, I removed `MusicBrainz ID`, but this shows that I removed `WikiData` and then
> added `WikiData` inplace of `MusicBrainz ID`.
> 
> `Deep-Diff` isn't working properly. Should it be replaced w some different library?

Same as before, then. That's unfortunate.
I wonder if there's a way to sort the items by id before doing the comparison, and whether that would help.

Would it help to print out the data of two specific revisions (before/after) to be able to test it in a sandbox?

prabalsingh24:
We ARE sorting it by ids. https://github.com/bookbrainz/bookbrainz-data-js/blob/1b3bf5c12326a8419401f283e25e014aca917347/src/util.js#L158

how about we compare relationshipSet and IdentifierSet by brute force. Picking up one relationship from the old set and check if it's present in the new set? Because we are facing problem only in sets (in the array). Lemme try this once. I think I can get it done.

MonkeyDo:
> We ARE sorting it by ids. https://github.com/bookbrainz/bookbrainz-data-js/blob/1b3bf5c12326a8419401f283e25e014aca917347/src/util.js#L158
> 
> how about we compare relationshipSet and IdentifierSet by brute force. Picking up one relationship from the old set and check if it's present in the new set? Because we are facing problem only in sets (in the array). Lemme try this once. I think I can get it done.

Hmmmm. It's been a while I looked into this in detail :p
I wonder then if on the contrary you should try not sorting the array and letting the new library handle it.
That being said, I don't know if we can assume the sets will be returned in the same order every time by the ORM…

MonkeyDo:
Looking at [the sorting code](https://github.com/bookbrainz/bookbrainz-data-js/blob/1b3bf5c12326a8419401f283e25e014aca917347/src/util.js#L124) again, it looks like we're not sorting by ID but rather by value and type.label, which might explain why it's all over the place.
You could try sorting by id or by type.id, that might do the trick.

prabalsingh24:
@MonkeyDo I'll be afk for about a week. I'll continue this later. Is that alright?

MonkeyDo:
> @MonkeyDo I'll be afk for about a week. I'll continue this later. Is that alright?

Of course! You certainly do not need my permission !
Take care!

src/server/routes/revision.js
```diff
@@ -16,7 +16,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
-
+import * as auth from '../helpers/auth';
```
MonkeyDo, line 19: This is not used

src/server/routes/revision.js
```diff
@@ -169,12 +169,14 @@ function diffRevisionsWithParents(revisions) {
 				.then(
 					(parent) => Promise.props({
 						changes: revision.diff(parent),
-						entity: revision.related('entity')
+						entity: revision.related('entity'),
+						entityAlias: revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias']})
```
MonkeyDo, line 173: Instead of fetching here, you can add `'data.aliasSet.defaultAlias'` to withRelated in `_createRevision` and just use `revision.related('data')` as we do for `entity` just above.

Should be replicated for the identical promise below.

MonkeyDo:
You can Ignore the previous review.
The rel formatting issue I was seeing was due to my local copy of bookbrainz-data that I was doing experiments with, didn't catch that at the time.
As for the suggestion to fetch 'data.aliasSet.defaultAlias', that doesn't seem to work either.

PR #411, opened 2020-04-10
BB-430: Add 'Copy' button for Sort Name field

chinmay-kothari:
### Problem

Copying the content of name field.

### Solution

Adds a copy button next to guess button

### Areas of Impact

sort-name-field.js


src/client/entity-editor/common/sort-name-field.js
```diff
@@ -148,9 +158,12 @@ function SortNameField({
 	const guessButton =
 		<Button bsStyle="primary" onClick={handleGuessClick}>Guess</Button>;
 
+	const copyButton =
+		<Button bsStyle="success" onClick={handleCopyClick}>Copy</Button>;
```
MonkeyDo, line 162: I see no reason for this button's color to be different from the other one.
Button colors have specific meaning rather than being only decorative.

If you are worried same-color buttons would be an issue because they look like one button, you'll have to find another solution.
Unfortunately we're using Bootstrap 3 which is lacking a few nice options like outline buttons from v4 that could be a good solution.

chinmay-kothari:
@MonkeyDo ,
I have changed the Copy button copy to the primary. While for spacing between the buttons since we don't have outline in bootstrap 3 I have provided an extra space `buttonAfter={[guessButton, ' ', copyButton]}` here. Though not the best way to do but it solved the problem of two buttons looking like one.

Let me know what you think of it.

PR #412, opened 2020-04-14
BB-451 Delete Relationships during handle delete for all entities

anirudhjain75:


src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
+		const newEntityRevisionPromise = await new RevisionModel({
+			bbid: entity.bbid,
+			dataId: null,
+			id: newRevisionPromiseRevision.get('id')
+		}).save(null, {
+			method: 'insert',
+			transacting
+		});
+
+		const entityRevision = newEntityRevisionPromise;
```
MonkeyDo, line 440: This should be `const entityRevision =  await new RevisionModel({…`:
- the name `newEntityRevisionPromise` is misleading: `new RevisionModel` returns a Promise, but when you `await` a promise, the result is a value, not a promise.
- the `const entityRevision = newEntityRevisionPromise;` is redundant.

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
 	const editorJSON = req.session.passport.user;
 	const {body}: {body: any} = req;
 
-	const entityDeletePromise = bookshelf.transaction((transacting) => {
+	const entityDeletePromise = bookshelf.transaction(async (transacting) => {
+		const otherEntities = await deleteRelationships(orm, transacting, entity);
+
 		const editorUpdatePromise =
 			utils.incrementEditorEditCountById(orm, editorJSON.id, transacting);
```
MonkeyDo, line 418: Incrementing the editor count should be taken care of by `saveEntitiesAndFinishRevision`, so I think at the moment you're incrementing it twice.

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
 
 		const newRevisionPromise = new Revision({
 			authorId: editorJSON.id
 		}).save(null, {transacting});
```
MonkeyDo, line 422: Here you could simplify like this: `const newRevision = await new Revision({…`  and remove the line `const newRevisionPromiseRevision = await newRevisionPromise;` below

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
+		});
+
+		const entityRevision = newEntityRevisionPromise;
+
+		const entityHeaderPromise = await new HeaderModel({
+			bbid: entity.bbid,
+			masterRevisionId: entityRevision.get('id')
+		}).save(null, {transacting});
+
+		const newRevision = await newRevisionPromise;
```
MonkeyDo, line 447: Looks like you are awaiting newRevisionPromise twice (lines 424 and 447).
With the change I commented above, you can just remove this line 447.

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
+			method: 'insert',
+			transacting
+		});
+
+		const entityRevision = newEntityRevisionPromise;
+
+		const entityHeaderPromise = await new HeaderModel({
+			bbid: entity.bbid,
+			masterRevisionId: entityRevision.get('id')
+		}).save(null, {transacting});
```
MonkeyDo, line 445: Same naming observation as above: `const entityHeader = await new HeaderModel({…`

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
+		const newRevision = await newRevisionPromise;
+
+		const mainEntity = await fetchOrCreateMainEntity(
+			orm, transacting, false, entity, entity.type
+		);
+
+		await saveEntitiesAndFinishRevision(orm, transacting, false, newRevision, mainEntity, otherEntities, editorJSON.id, body.note);
 
 		const searchDeleteEntityPromise = search.deleteEntity(entity)
 			.catch(err => { log.error(err); });
```
MonkeyDo, line 456: Here, I previously implemented a way to make sure that if search is unavailable, we don't fail the whole transaction: `handler.sendPromiseResult` below accepts an optional callback.

So you can remove these two line from here, and  instead below at line 462 you can `return handler.sendPromiseResult(res, entityDeletePromise, search.deleteEntity);` <- note our callback passed as the third argument.

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
+			masterRevisionId: entityRevision.get('id')
+		}).save(null, {transacting});
+
+		const newRevision = await newRevisionPromise;
+
+		const mainEntity = await fetchOrCreateMainEntity(
+			orm, transacting, false, entity, entity.type
+		);
+
+		await saveEntitiesAndFinishRevision(orm, transacting, false, newRevision, mainEntity, otherEntities, editorJSON.id, body.note);
```
MonkeyDo, line 453: Here you want something like `const savedMainEntity = await saveEntitiesAndFinishRevision(…`, because we want to return that savedMainEntity at the end.

src/server/routes/entity/entity.js
```diff
@@ -283,59 +411,52 @@ export function handleDelete(
...
 		const searchDeleteEntityPromise = search.deleteEntity(entity)
 			.catch(err => { log.error(err); });
-		return Promise.join(
-			editorUpdatePromise, newRevisionPromise, notePromise,
-			newEntityRevisionPromise, entityHeaderPromise, parentAddedPromise,
-			searchDeleteEntityPromise
-		);
+		return [editorUpdatePromise, newRevisionPromise,
+			newEntityRevisionPromise, entityHeaderPromise,
+			searchDeleteEntityPromise];
```
MonkeyDo, line 459: Here, you'll only need to `return savedMainEntity;` (see comment above)

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+	]);
+
+	return savedEntities.find(entityModel => entityModel.get('bbid') === mainEntity.get('bbid')) || mainEntity;
+}
+
+async function deleteRelationships(orm, transacting, mainEntity) {
+	const mainBBID = mainEntity.bbid;
+	const {RelationshipSet} = orm;
+	const {relationshipSet} = mainEntity;
+	const otherBBID = [];
```
MonkeyDo, line 340: `otherBBIDs` with an 's', since it's an array.

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+	const otherEntities = [];
+
+	if (relationshipSet) {
+		// Create a list of BBID's that is related to the deleted entity
+		relationshipSet.relationships.map((relationship) => {
```
MonkeyDo, line 345: array.forEach seems more appropriate than array.map since you're not using the result of .map
You won't need to `return` anything in the forEach

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+			}
+			return null;
+		});
+
+		// Loop over the BBID's of other entites related to deleted entity
+		if (otherBBID.length) {
+			await Promise.all(otherBBID.map(async (entityBbid) => {
+				const otherEntity = await getEntityByBBID(orm, transacting, entityBbid);
+
+				await otherEntities.push(otherEntity);
```
MonkeyDo, line 360: No need to `await` an array.push

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+				otherEntityRelationships = otherEntityRelationships.filter(({sourceBbid, targetBbid}) =>
+					mainBBID !== sourceBbid && mainBBID !== targetBbid);
+
+				const newRelationshipSet = new RelationshipSet();
+
+				await newRelationshipSet.save(null, {transacting});
+
+				await orm.func.relationship.updateRelationshipSets(
+					orm, transacting, newRelationshipSet, otherEntityRelationships
+				);
```
MonkeyDo, line 378: `updateReationshipSets` will create and return a new RelationshipSet for you, so you can instead do
`const newRelationshipSet = await orm.func.relationship.updateRelationshipSets(…`

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+			else if (relationship.targetBbid === mainBBID) {
+				otherBBID.push(relationship.sourceBbid);
+			}
+			return null;
+		});
+
+		// Loop over the BBID's of other entites related to deleted entity
+		if (otherBBID.length) {
+			await Promise.all(otherBBID.map(async (entityBbid) => {
+				const otherEntity = await getEntityByBBID(orm, transacting, entityBbid);
```
MonkeyDo, line 358: There are cases where otherEntity can be null (for example for a lingering relationship with a deleted entity — the very problem this PR aims to solve :D), so you should prematurely `return null` in such a case.

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+async function deleteRelationships(orm, transacting, mainEntity) {
+	const mainBBID = mainEntity.bbid;
+	const {RelationshipSet} = orm;
+	const {relationshipSet} = mainEntity;
+	const otherBBIDs = [];
+	const otherEntities = [];
+
+	if (relationshipSet) {
+		// Create a list of BBID's that is related to the deleted entity
+		relationshipSet.relationships.foreach((relationship) => {
```
MonkeyDo, line 345: That's `forEach` with a capital E

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+				otherBBIDs.push(relationship.sourceBbid);
+			}
+		});
+
+		// Loop over the BBID's of other entites related to deleted entity
+		if (otherBBIDs.length) {
+			if (otherBBIDs.length == 0) {
+				return null;
+			}
+			await otherBBIDs.map(async (entityBbid) => {
```
MonkeyDo, line 359: Unfortunately, we can't await an array of async/promises like that (I wish!)
You have to do `await Promise.all(otherBBIDs.map(async (entityBbid) => {…`
In short, await an array of promises, since .map returns an array.

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,135 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+
+				
+
+				const newRelationshipSet = await orm.func.relationship.updateRelationshipSets(
+					orm, transacting, otherEntityRelationshipSet, otherEntityRelationships
+				);
+
+				otherEntity.set(
+					'relationshipSetId',
+					newRelationshipSet.get('id')
```
MonkeyDo, line 382: I was wrong about how this works: `updateRelationshipSets` returns an object with the involved entities' BBIDs as keys and the RelationshipSets as value.

So you'll need to do `newRelationshipSet[entityBbid].get('id')` instead to target the correct RelationshipSet

src/server/routes/entity/entity.js
MonkeyDo: I was wrong !
`otherEntity` won't be null. However, for a deleted entity, `await otherEntity.relationshipSet()` below will return null and `otherEntityRelationshipSet.related('relationships').toJSON()`will throw an error.
So it's `otherEntityRelationshipSet` you should check for nullishness (`if(_.isNil(otherEntityRelationshipSet)) {…`)

src/server/routes/entity/entity.js
```diff
@@ -274,6 +273,136 @@ export function addNoteToRevision(req: PassportRequest, res: $Response) {
...
+			if (otherBBIDs.length === 0) {
+				return [];
+			}
+			await Promise.all(otherBBIDs.map(async (entityBbid) => {
+				const otherEntity = await getEntityByBBID(orm, transacting, entityBbid);
+
+				const otherEntityRelationshipSet = await otherEntity.relationshipSet()
+					.fetch({require: false, transacting, withRelated: 'relationships'});
+
+				if (_.isNull(otherEntityRelationshipSet)) {
```
MonkeyDo, line 364: `_.isNil` allows to check in one go if something is null OR undefined.
I usually use that instead of `isNull` because javascript is pretty loose when it comes to what it returns.
Although in the cases I've tried isNull seemed to work, Id suggest erring on the safer side, since an `undefined` otherEntityRelationshipSet would also throw an error below.

PR #413, opened 2020-04-14
BB-256: Build client-side code with Webpack

MonkeyDo:
https://tickets.metabrainz.org/browse/BB-256

Transpile, bundle and minify client-side code using Webpack instead of browserify.

This is only for client-side code and not for the server-side code.
I haven't yet seen any good reason to replace the current server-side code transpilation that simply uses the babel executable.

This setup also still allows to  run in dev mode with hot module replacement

Also solves https://tickets.metabrainz.org/browse/BB-255 (minify react code for production)


MonkeyDo:
@anirudhjain75 can you help me test this when you find some time?
Hopefully the local development setup should be fixed. Let me know if it works again for you.

P.S: you'll have to run `npm install` after pulling this PR.

anirudhjain75:
Sure :)

anirudhjain75:
@MonkeyDo Yes I can confirm that it's working fine now :). Thanks

PR #415, opened 2020-04-17
BB-455: Update Flow Version

chinmay-kothari:
### Problem

Upgrading the flow version from the 0.69.0 to the latest, 0.123.0

### Solution

I first removed the current flow version with `sudo rm -rf path` then `yarn global add flow-bin@latest` then after that `npm install flow-bin@latest`.

### Areas of Impact

package.json
package-lock.json

I haven't changed the source codebase because on doing `yarn create flow-upgrade --all` it did following changes in around 103 files disturbing the eslint-sort:
`import React from 'react';` changed to `import * as React from 'react';`

`class Achievement extends React.Component {` to  `class Achievement extends React.Component<$FlowFixMeProps, $FlowFixMeState> {`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,15 +10,15 @@
       "integrity": "sha512-XXLgAm6LBbaNxaGhMAznXXaxtCWfuv6PIDJ9Alsy9JYTOh+j2jJz+L/162kkfU1j/pTSxK1xGmlwI4pdIMkoag==",
       "dev": true,
       "requires": {
-        "chokidar": "2.1.8",
-        "commander": "4.1.1",
-        "convert-source-map": "1.7.0",
-        "fs-readdir-recursive": "1.1.0",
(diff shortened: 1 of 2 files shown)


chinmay-kothari:
@MonkeyDo 
On running the local server using npm, the server is getting started with many warnings which I guess are associated with the previous flow version. 
I may be totally wrong in this so can you please take a look into this. Thanks

MonkeyDo:
@chinmay-kothari Not to worry, the warnings are from ESLint, not with Flow, and they were not introduced with your PR.

However, something I haven't done in a while is run `npm run flow` which shows a lot of flow type errors.
If you feel like this is something you would like to tackle, you can open a new PR to try to fix some of those; it will likely be a little-by-little sort of job (just like the ESLint warnings you saw).

In any case, this PR is ready to be merged. Thanks :)

PR #418, opened 2020-04-19
BB-445: Remove GSoC 2016 as achievement

chinmay-kothari:
### Problem

Achievements section has GSoC 2016 as an achievement which didn't even have an image.

### Solution

Deletes GSoC 2016 as an achievement from the database.

### Areas of Impact

Database: sql/migration


paramsingh:
What's the point of doing this? Gsoc 2016 was when we added achievements and the person who wrote the code has the achievement, I think we should keep acknowledging them. :)

Also, maybe there's an idea here to add achievements for the rest of the GSoC students?

chinmay-kothari:
> What's the point of doing this? Gsoc 2016 was when we added achievements and the person who wrote the code has the achievement, I think we should keep acknowledging them. :)
> 
> Also, maybe there's an idea here to add achievements for the rest of the GSoC students?

chinmay-kothari:
@paramsingh 
Yeah true they actually not creating any problem or so. I liked your idea of rather adding GSoC as an achievement not just for 2016 but for rest of the people also. That would certainly make GSoC students much more happier XD. 
What do you think about this @MonkeyDo ?

MonkeyDo:
After some thought, I suggest changing the scope of the achievement to "Participated in GSOC" without a specific year.
That way we both maintain the achievement and also extend it so that more than one single person is able to claim it.

What do you think @chinmay-kothari and @paramsingh?

paramsingh:
that sounds great to me!

chinmay-kothari:
@MonkeyDo , I have added GSoC as an achievement in general not year specific as you mentioned. Let me know how does it look?

chinmay-kothari:
> While this would most likely work as expected, why not simply modify the fields in the existing badge?

I thought about that and used updating the row rather than deleting and adding. I did this few days back but forgot to commit :p

PR #419, opened 2020-04-20
Add tests for deleteRelationships

anirudhjain75:


test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
+	afterEach(truncateEntities);
+	// let author = await createAuthor();s
+	// const workUUID = getRandomUUID();
+	// const authorBbid = author.get('bbid');
+	// await createRelationshipSet(authorBbid, workUUID, 'Work');
+	// author = await author.refresh({withRelated: 'relationshipSet.relationships'});
+	// const authorJSON = author.toJSON();
+	// const authorRelationships = authorJSON.relationshipSet.relationships;
+	it('it should return an empty array if there are no relationships to delete', async () => {
+		const affectedEntities = await deleteRelationships(orm, null, author);
```
MonkeyDo, line 107: This test is currently not working properly.
There are two issues:
1. you're passing the author model instead of a JSON object
2. you haven't fetched the relationships before passing it.

With those two things changed, the test still passes even though `createAuthor` creates a relationship. The reason is the same as the test where you have a relationship to a deleted entity: the other entity create by `createAuthor` doesn't have a relationshipSet to modify.

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
+
+		// Fetch Refreshed Entities from database
+		author = await new Author({bbid: author.get('bbid')}).fetch({withRelated: 'relationshipSet.relationships'});
+		work = await new Work({bbid: work.get('bbid')}).fetch({withRelated: 'relationshipSet.relationships'});
+
+		// Get Updated Work
+		// fetchedWork = await new Work({bbid: work.get('bbid')}).fetch();
+		const affectedEntities = await deleteRelationships(orm, null, author.toJSON());
+
+		expect(affectedEntities).to.have.length(1);
```
MonkeyDo, line 279: You're missing an expectation, looking at the test title:
`expect(workRelationshipSetJSON.id).to.not.equal(affectedEntities[0].get('relationshipSetId')`

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
+			id: 1,
+			label: 'test label',
+			linkPhrase: 'test phrase',
+			reverseLinkPhrase: 'test reverse link phrase',
+			sourceEntityType: 'Author',
+			targetEntityType: 'Work'
+		};
+		relationshipTypeData.id = random.number();
+		await new RelationshipType(relationshipTypeData)
+			.save(null, {method: 'insert'});
```
MonkeyDo, line 132: This can be created once in the `before` function and be reused (store the id in a relationshipTypeId variable)

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
 	});
 });
+
+describe('deleteRelationships', () => {
+	const {Author, Work, Revision, WorkRevision, WorkHeader, Relationship, RelationshipSet, RelationshipType, bookshelf} = orm;
+	let author;
+	let editorJSON;
+
+	beforeEach(async () => {
+		const editor = await createEditor();
```
MonkeyDo, line 94: The editor you can create once and reuse (you'll need to move it to a `before`  function).

For that to work, you'll need to change the `afterEach` below to an `after`.

I think this should be fine since you overwrite the `author` variable  before each test.

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
+		work.set('revisionId', revisionId);
+		work.set('relationshipSetId', workRelationshipSetJSON.id);
+
+		// Save the Entities and Update the database
+		await author.save(null, {method: 'update'});
+		await work.save(null, {method: 'update'});
+
+		// Fetch Refreshed Entities from database
+		author = await new Author({bbid: author.get('bbid')}).fetch({withRelated: 'relationshipSet.relationships'});
+		work = await new Work({bbid: work.get('bbid')}).fetch({withRelated: 'relationshipSet.relationships'});
```
MonkeyDo, line 177: No need to fetch these guys here

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
+		await workA.save(null, {method: 'update'});
+		await workB.save(null, {method: 'update'});
+
+		// Get updated Entity
+		workA = await new Work({bbid: workA.get('bbid')}).fetch({withRelated: 'relationshipSet.relationships'});
+
+		// Get list of Affected Entities
+		const affectedEntities = await deleteRelationships(orm, null, workA.toJSON());
+
+		expect(affectedEntities).to.have.length(1);
```
MonkeyDo, line 385: Here let's also ensure it's the right affected entity bbid  you end up with (the author)

You also want to make sure the author's new relationship set 1. has a different id than before and 2. that it still contains a relationship to workB

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,436 @@ describe('getDefaultAliasIndex', () => {
...
+			dataId: null,
+			id: revisionId
+		}).save(null, {
+			method: 'insert'
+		});
+
+		await new WorkHeader({
+			bbid: work.get('bbid'),
+			masterRevisionId: revisionId
+		}).save(null);
```
MonkeyDo, line 196: This should make its way to your `deleteEntity` helper that currently sits unused :)

test/src/server/routes/entity/entity.js
MonkeyDo: There's one item that you did not address:
You also want to make sure the author's new relationship set has a different id than before:
authorRelationshipSetId should not be equal to affectedEntityRelationshipSet.get('id')

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,391 @@ describe('getDefaultAliasIndex', () => {
...
+
+		// Relate Relationship Data with Relationship Type
+		relationshipDataA.typeId = relationshipTypeData.id;
+		relationshipDataA.id = random.number();
+
+		relationshipDataB.typeId = relationshipTypeData.id;
+		relationshipDataB.id = random.number();
+
+		relationshipDataC.typeId = relationshipTypeData.id;
+		relationshipDataC.id = random.number();
```
MonkeyDo, line 399: Why not set those directly in the objects above?
Just a suggestion to try and reduce the number of lines.
(this could apply to the other tests as well)

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,391 @@ describe('getDefaultAliasIndex', () => {
...
+		const workCRelationshipSetId = random.number();
+
+		// Create Relationship sets
+		let authorRelationshipSet = await new RelationshipSet({id: authorRelationshipSetId})
+			.save(null, {method: 'insert'})
+			.then(
+				(model) =>
+					model.relationships().attach([relationshipA, relationshipB, relationshipC]).then(() => model)
+			);
+		authorRelationshipSet = await authorRelationshipSet.refresh({withRelated: 'relationships'});
```
MonkeyDo, line 424: There's a need to refresh any of the newly created relationship sets. The reason is that we don't actually use use —in this case— authorRelationshipSet anywhere after its creation (apart from getting the id, which will work without the refresh, and could also be done by using the authorRelationshipSetId variable).

The reason you had to refresh earlier was for debugging purposes so that you could see which relationships were in the set, which isn't needed anymore.
Removing all the calls to `.refresh` in this test file will shave off a bit of time when running the tests.

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,391 @@ describe('getDefaultAliasIndex', () => {
...
+
+		// Get list of Affected Entities
+		const affectedEntities = await deleteRelationships(orm, null, workA.toJSON());
+
+		const affectedEntityRelationshipSet = await new RelationshipSet({
+			id: affectedEntities[0].get('relationshipSetId')
+		}).fetch({withRelated: 'relationships'});
+
+		expect(affectedEntities[0].get('bbid')).to.equal(author.get('bbid'));
+		expect(affectedEntityRelationshipSet.toJSON().relationships[0].targetBbid).to.be.equal(workB.get('bbid'));
```
MonkeyDo, line 354: Here you can serialize only the relationships: `affectedEntityRelationshipSet.related('relationships').toJSON()[0].targetBbid`

test/src/server/routes/entity/entity.js
```diff
@@ -80,3 +84,391 @@ describe('getDefaultAliasIndex', () => {
...
+	});
+	it('it should return an empty array if the other entity is a deleted entity (no dataId)', async () => {
+		// Create Entities
+		let work = await createWork();
+
+		// Create Relation and its data between two entites
+		const relationshipData = {
+			id: 1,
+			sourceBbid: author.toJSON().bbid,
+			targetBbid: work.toJSON().bbid,
```
MonkeyDo, line 128: Wherever you can in this file, replacing call like `.toJSON().property` with `.get('property')` will save a bit of time, since serializing to JSON is an expensive operation, however convenient it may seem.

The only instances where toJSON is needed is to pass an entity to `deleteRelationships`.

anirudhjain75:
@MonkeyDo I've made the recommended changes

PR #423, opened 2020-05-08
search: add 'editor' type in the search result

prabalsingh24:
### Problem
Added `Editor` in the search result

### Solution
Screenshot_2020-05-08 Search Results – BookBrainz

### DIFF
--- a/src/server/helpers/search.js
+++ b/src/server/helpers/search.js
@@ -161,7 +161,7 @@ async function _processEntityListForBulk(entityList) {
 	await Promise.all(indexOperations);
 }
 
-export function autocomplete(orm, query, collection) {
+export function autocomplete(orm, query, type) {
 	let queryBody = null;
 
 	if (commonUtils.isValidBBID(query)) {
@@ -189,12 +189,12 @@ export function autocomplete(orm, query, collection) {
 		index: _index
 	};
 
-	if (collection) {
-		if (Array.isArray(collection)) {
-			dslQuery.type = collection.map(_.snakeCase);
+	if (type) {
+		if (Array.isArray(type)) {
+			dslQuery.type = type.map(_.snakeCase);
 		}
 		else {
-			dslQuery.type = _.snakeCase(collection);
+			dslQuery.type = _.snakeCase(type);
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


prabalsingh24:
TODO: 
1. I haven't changed `collection` to `type` in the routes yet.
2. If you update your username, it is not getting updated in the navbar. (it is getting updated in the search result though)

prabalsingh24:
You'll have to do a re-indexing first.

MonkeyDo:
It does look like after changing the username it's reindexed properly and I find it in the search results.

Do you want to change collection to type in this PR, or in a new one?

prabalsingh24:
> It does look like after changing the username it's reindexed properly and I find it in the search results.

But it's not getting updated in the NavBar. Right top corner.

prabalsingh24:
sessions are stored in redis right?

prabalsingh24:
I'll make the change in this PR only.

MonkeyDo:
> > It does look like after changing the username it's reindexed properly and I find it in the search results.
> 
> But it's not getting updated in the NavBar. Right top corner.

Huh, that's right. Sorry I read too fast :p
I just checked on test.bb and it looks like the issue is not linked to this PR at all, so not a blocker. I'll create a ticket.

Sessions are indeed stored on redis. Do you think they might need to be invalidated or refreshed for the username to appear correctly?

prabalsingh24:
I'll have a look. I will change `All Entity` --> `All Entities` and `collection` --> `type`

src/server/helpers/search.js
```diff
@@ -358,16 +371,36 @@ export async function generateIndex(orm) {
...
+	/** To index names, we use aliasSet.aliases.name and bbid, which Editors don't have.
+	 * We massage the editor to return a similar format as BB entities
+	 */
+	const processedEditors = editors.map((editor) => new Object({
+		aliasSet: {
+			aliases: [
+				{name: editor.name}
+			]
+		},
+		bbid: editor.id,
```
prabalsingh24, line 389: Do you approve of this?
By adding `bbid` here,  I avoided adding a special case for `editor` here. [search-table-row-link](https://github.com/bookbrainz/bookbrainz-site/blob/a8aa31b49e8cb9e63f9ddd71f9bf37fef528e350/src/client/components/pages/parts/search-results.js#L77)
The same can be done for collections too..

prabalsingh24:
I changed it. I think this can be merged now :)

MonkeyDo:
I don't see the changes to `src/client/entity-editor/name-section/actions.js`. Maybe you forgot to push?

prabalsingh24:
> I don't see the changes to `src/client/entity-editor/name-section/actions.js`. Maybe you forgot to push?

Oops. Yeah I forgot to push :P

prabalsingh24:
I'll have to change some tests too :)

PR #424, opened 2020-05-11
Usercollection add tables

prabalsingh24:
### Problem
Adding tables for UserCollection

### DIFF
--- a/sql/migrations/user-collection/up.sql
+++ b/sql/migrations/user-collection/up.sql
@@ -5,7 +5,7 @@ CREATE TABLE IF NOT EXISTS bookbrainz.user_collection (
 	owner_id INT NOT NULL,
 	name VARCHAR(80) NOT NULL CHECK (name <> ''),
 	description TEXT NOT NULL DEFAULT '',
-	type bookbrainz.entity_type NOT NULL,
+	entity_type bookbrainz.entity_type NOT NULL,
 	public BOOLEAN NOT NULL DEFAULT FALSE,
 	created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now()),
 	last_modified TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
--- a/sql/migrations/user-collection/up.sql
+++ b/sql/migrations/user-collection/up.sql
@@ -2,26 +2,26 @@ BEGIN;
 
 CREATE TABLE IF NOT EXISTS bookbrainz.user_collection (
 	id UUID PRIMARY KEY DEFAULT public.uuid_generate_v4(),
-	editor_id INT,
+	owner_id INT NOT NULL,
 	name VARCHAR(80) NOT NULL CHECK (name <> ''),
 	description TEXT NOT NULL DEFAULT '',
 	type bookbrainz.entity_type NOT NULL,
 	public BOOLEAN NOT NULL DEFAULT FALSE,
 	created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now()),
-   	last_modified TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
+	last_modified TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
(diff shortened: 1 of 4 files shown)


prabalsingh24:
Should I change `editor_id` --> `owner_id` in the `user_collection` table? Because we have owner and collaborators?

sql/migrations/user-collection/up.sql
```diff
@@ -0,0 +1,30 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS bookbrainz.user_collection (
+	id UUID PRIMARY KEY DEFAULT public.uuid_generate_v4(),
+	editor_id INT,
+	name VARCHAR(80) NOT NULL CHECK (name <> ''),
+	description TEXT NOT NULL DEFAULT '',
+	type bookbrainz.entity_type NOT NULL,
```
MonkeyDo, line 8: What do you think about `entity_type` instead of `type`?

sql/migrations/user-collection/up.sql
```diff
@@ -0,0 +1,30 @@
...
+	public BOOLEAN NOT NULL DEFAULT FALSE,
+	created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now()),
+   	last_modified TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
+);
+ALTER TABLE bookbrainz.user_collection ADD FOREIGN KEY (editor_id) REFERENCES bookbrainz.editor (id);
+
+CREATE TABLE IF NOT EXISTS bookbrainz.user_collection_item (
+	collection_id UUID,
+	bbid UUID,
+    added_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
```
MonkeyDo, line 18: Small nitpick: this line isn't aligned

sql/migrations/user-collection/up.sql
```diff
@@ -0,0 +1,30 @@
...
+
+CREATE TABLE IF NOT EXISTS bookbrainz.user_collection (
+	id UUID PRIMARY KEY DEFAULT public.uuid_generate_v4(),
+	editor_id INT,
+	name VARCHAR(80) NOT NULL CHECK (name <> ''),
+	description TEXT NOT NULL DEFAULT '',
+	type bookbrainz.entity_type NOT NULL,
+	public BOOLEAN NOT NULL DEFAULT FALSE,
+	created_at TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now()),
+   	last_modified TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
```
MonkeyDo, line 11: Small nitpick: this line isn't aligned

sql/migrations/user-collection/up.sql
```diff
@@ -0,0 +1,30 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS bookbrainz.user_collection (
+	id UUID PRIMARY KEY DEFAULT public.uuid_generate_v4(),
+	editor_id INT,
```
MonkeyDo, line 5: I agree with changing this to owner_id

sql/migrations/user-collection/up.sql
prabalsingh24: I prefer `type` here. `entity_type` would be fine too. Okay with either of them :P what do you say? should I change it to `entity_type`

sql/migrations/user-collection/up.sql
prabalsingh24: i was using spaces instead of tab here :P

sql/migrations/user-collection/up.sql
prabalsingh24: i was using spaces instead of tab here :P

sql/migrations/user-collection/up.sql
MonkeyDo: I think entity_type is clearer. Otherwise it suggests this column is "collection type" which could be interpreted different ways (public/private, for example?).
It's a bit of a nitpick though and both would work.

prabalsingh24: yeah true, I didn't think of public/private. I'll change it :)

MonkeyDo:
user_collections

PR #425, opened 2020-05-12
User collections create/edit route

prabalsingh24:
### Problem
Added Collection-Create/Edit Route

### DIFF
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -27,6 +27,7 @@ import ReactSelect from 'react-select';
 import SelectWrapper from '../input/select-wrapper';
 import classNames from 'classnames';
 import request from 'superagent';
+import {uniqBy} from 'lodash';
 
 
 const {Alert, Button, Col} = bootstrap;
@@ -85,15 +86,11 @@ class UserCollectionForm extends React.Component {
 		request.post(submissionURL)
 			.send(data)
 			.then((res) => {
-				if (res.status === 200) {
-					window.location.href = `/collection/${res.body.id}`;
-				}
-				else {
-					this.setState(prevState => ({
-						collection: prevState.collection,
-						errorText: `Internal Error: ${res.body.error}`
-					}));
-				}
+				window.location.href = `/collection/${res.body.id}`;
+			}, (error) => {
+				this.setState({
+					errorText: 'Internal Error'
+				});
 			});
 	}
 
@@ -102,7 +99,7 @@ class UserCollectionForm extends React.Component {
 	}
 
 	getCleanedCollaborators() {
-		const cleanedCollaborators = this.state.collaborators.filter(collaborator => collaborator && collaborator.id !== null);
+		const cleanedCollaborators = uniqBy(this.state.collaborators.filter(collaborator => collaborator && collaborator.id !== null), 'id');
 		return cleanedCollaborators;
 	}
 
@@ -171,7 +168,7 @@ class UserCollectionForm extends React.Component {
 						mdOffset={2}
 					>
 						<form
-							className="form-horizontal"
+							className="padding-sides-0"
 							onSubmit={this.handleSubmit}
 						>
 							<CustomInput
@@ -216,7 +213,7 @@ class UserCollectionForm extends React.Component {
 											type="button"
 											onClick={() => this.handleRemoveCollaborator(index)}
 										>
-											Remove
+											<FontAwesomeIcon icon="times"/>&nbsp;Remove
 										</Button>
 									);
 									return (
@@ -254,7 +251,7 @@ class UserCollectionForm extends React.Component {
 								</Button>
 							</div>
 							<div className={errorAlertClass}>
-								<Alert bsStyle="danger">Submission Error: {errorText}</Alert>
+								<Alert bsStyle="danger">Error: {errorText}</Alert>
 							</div>
 						</form>
 					</Col>
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -110,12 +110,9 @@ class UserCollectionForm extends React.Component {
 	}
 
 	handleRemoveCollaborator(index) {
-		// eslint-disable-next-line react/no-access-state-in-setstate
-		const newCollaborators = this.state.collaborators;
-		newCollaborators.splice(index, 1);
-		this.setState({
-			collaborators: newCollaborators
-		});
+		this.setState(prevState => ({
+			collaborators: prevState.collaborators.splice(index, 1) && prevState.collaborators
+		}));
 	}
 
 	handleChangeCollaborator(newCollab, index) {
@@ -129,11 +126,13 @@ class UserCollectionForm extends React.Component {
 		else {
 			newCollaborator = newCollab;
 		}
-		// eslint-disable-next-line react/no-access-state-in-setstate
-		const newCollaborators = this.state.collaborators;
-		newCollaborators[index] = newCollaborator;
-		this.setState({
-			collaborators: newCollaborators
+
+		this.setState((prevState) => {
+			const newCollaborators = prevState.collaborators;
+			newCollaborators[index] = newCollaborator;
+			return {
+				collaborators: newCollaborators
+			};
 		});
 	}
 
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -1,6 +1,5 @@
 /*
- * Copyright (C) 2015       Ben Ockmore
- *               2015-2016  Sean Burke
+ * Copyright (C) 2020 Prabal Singh
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -16,6 +15,7 @@
  * with this program; if not, write to the Free Software Foundation, Inc.,
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
+
 
 import * as bootstrap from 'react-bootstrap';
 import CustomInput from '../../input';
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -20,22 +20,25 @@
 import * as bootstrap from 'react-bootstrap';
 import CustomInput from '../../input';
 import EntitySearchFieldOption from '../../entity-editor/common/entity-search-field-option';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactSelect from 'react-select';
 import SelectWrapper from '../input/select-wrapper';
 import classNames from 'classnames';
-import request from 'superagent-bluebird-promise';
+import request from 'superagent';
 
 
-const {Alert, Button, Col, Grid, Row} = bootstrap;
+const {Alert, Button, Col} = bootstrap;
 
 class UserCollectionForm extends React.Component {
 	constructor(props) {
 		super(props);
 
+		const {collaborators, ...collection} = props.collection;
 		this.state = {
-			collection: props.collection,
+			collaborators,
+			collection,
 			errorText: null
 		};
 
@@ -52,10 +55,9 @@ class UserCollectionForm extends React.Component {
 		evt.preventDefault();
 
 		if (!this.isValid()) {
-			this.setState(prevState => ({
-				collection: prevState.collection,
+			this.setState({
 				errorText: 'Incomplete Form'
-			}));
+			});
 			return;
 		}
 
@@ -82,7 +84,6 @@ class UserCollectionForm extends React.Component {
 		}
 		request.post(submissionURL)
 			.send(data)
-			.promise()
 			.then((res) => {
 				if (res.status === 200) {
 					window.location.href = `/collection/${res.body.id}`;
@@ -101,35 +102,23 @@ class UserCollectionForm extends React.Component {
 	}
 
 	getCleanedCollaborators() {
-		const cleanedCollaborators = this.state.collection.collaborators.filter(collaborator => collaborator && collaborator.id !== null);
+		const cleanedCollaborators = this.state.collaborators.filter(collaborator => collaborator && collaborator.id !== null);
 		return cleanedCollaborators;
 	}
 
 	handleAddCollaborator() {
-		const newCollaborators = this.state.collection.collaborators;
-		newCollaborators.push({
-			id: null,
-			name: ''
-		});
 		this.setState(prevState => ({
-			collection: {
-				...prevState.collection,
-				collaborators: newCollaborators
-			},
-			errorText: prevState.errorText
+			collaborators: [...prevState.collaborators, {id: null, name: ''}]
 		}));
 	}
 
 	handleRemoveCollaborator(index) {
-		const newCollaborators = this.state.collection.collaborators;
+		// eslint-disable-next-line react/no-access-state-in-setstate
+		const newCollaborators = this.state.collaborators;
 		newCollaborators.splice(index, 1);
-		this.setState(prevState => ({
-			collection: {
-				...prevState.collection,
-				collaborators: newCollaborators
-			},
-			errorText: prevState.errorText
-		}));
+		this.setState({
+			collaborators: newCollaborators
+		});
 	}
 
 	handleChangeCollaborator(newCollab, index) {
@@ -143,19 +132,16 @@ class UserCollectionForm extends React.Component {
 		else {
 			newCollaborator = newCollab;
 		}
-		const newCollaborators = this.state.collection.collaborators;
+		// eslint-disable-next-line react/no-access-state-in-setstate
+		const newCollaborators = this.state.collaborators;
 		newCollaborators[index] = newCollaborator;
-		this.setState(prevState => ({
-			collection: {
-				...prevState.collection,
-				collaborators: newCollaborators
-			},
-			errorText: prevState.errorText
-		}));
+		this.setState({
+			collaborators: newCollaborators
+		});
 	}
 
 	render() {
-		if (this.state.collection.collaborators.length === 0) {
+		if (this.state.collaborators.length === 0) {
 			this.handleAddCollaborator();
 		}
 
@@ -172,18 +158,13 @@ class UserCollectionForm extends React.Component {
 		const {errorText} = this.state;
 		const errorAlertClass =
 			classNames('text-center', 'margin-top-1', {hidden: !errorText});
-		const submitLabel = this.props.collection.name ? 'UPDATE' : 'CREATE';
+		const submitLabel = this.props.collection.name ? 'Update collection' : 'Create collection';
 
 		/* eslint-disable react/jsx-no-bind */
 		return (
-			<Grid>
-				<h1>Create a collection</h1>
-				<Row>
-					<Col md={12}>
-						<p className="lead">Create Your Collection</p>
-					</Col>
-				</Row>
-				<Row>
+			<div>
+				<h1>Create your collection</h1>
+				<div>
 					<Col
 						id="collectionForm"
 						md={8}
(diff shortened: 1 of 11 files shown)


prabalsingh24:
This won't work rn. I made some changes in the bookshelf data and forgot to update it my project.

src/server/helpers/middleware.js
```diff
@@ -171,15 +171,14 @@ export function makeCollectionLoader() {
 		if (commonUtils.isValidBBID(collectionId)) {
 			try {
 				const collection = await new UserCollection({id: collectionId}).fetch({
-					require: false,
-					withRelated: ['collaborators.collaboratorDetail']
+					require: true,
+					withRelated: ['collaborators.collaboratorDetails']
```
MonkeyDo, line 175: Looking at naming for other models, I'd expect 'collaborators.details' here

prabalsingh24:
Thanks for the super quick review @MonkeyDo (I thought you were busy w LB stuff :P )

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,266 @@
...
+		}
+		request.post(submissionURL)
+			.send(data)
+			.promise()
+			.then((response) => {
+				window.location.href = `/collection/${response.body.id}`;
+			});
+	}
+
+	valid() {
```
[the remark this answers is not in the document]
prabalsingh24: Wow, this is a new feature right? to make changes directly from here.
Screenshot_2020-05-13  WIP  User collections create route by prabalsingh24 · Pull Request #425 · bookbrainz bookbrainz-site

src/server/helpers/middleware.js
prabalsingh24: https://github.com/bookbrainz/bookbrainz-data-js/blob/fc02949c3426a7eb213ad68df6c544045709fbf9/src/models/userCollectionCollaborator.js#L24
I think `UserCollection.collaborators.collaboratorDetails` is better because there is a `editor.collaboratesOn.collectionDetails` as well

Or Maybe `editor.collaboratesOn.collection` and `UserCollection.collaborators.details` sounds better..
What do you think?

src/server/helpers/collectionRouteUtils.js
```diff
@@ -0,0 +1,93 @@
...
+		);
+	});
+	removedCollaborators.forEach((collaborator) => {
+		collaboratorPromises.push(
+			new UserCollectionCollaborator({})
+				.where('collection_id', newCollection.get('id'))
+				.where('editor_id', collaborator.id)
+				.destroy()
+		);
+	});
```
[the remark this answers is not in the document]
prabalsingh24: ``` This method (along with Collection#detach and Collection#updatePivot) are mixed in to a Collection when returned by a belongsToMany relation. ```

Here I used `hasMany` relation. [UserCollectionModel](https://github.com/bookbrainz/bookbrainz-data-js/blob/fc02949c3426a7eb213ad68df6c544045709fbf9/src/models/userCollection.js#L25)

Also `UserCollection().collaborators.attach()` is giving error : `attach is not a function` because of `hasMany` relation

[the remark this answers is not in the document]
MonkeyDo: Well, in my experience it should work for both sides of the relationship, but maybe I'm getting it wrong. 
It should be `UserCollection().collaborators().attach()` with parenthesis after collaborators; Wanna try it? It'll make your job a bit easier.

src/client/components/forms/userCollection.js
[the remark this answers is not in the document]
MonkeyDo: I don't think it's new new but the ability to do it for multiple lines is https://github.blog/changelog/2020-02-26-multi-line-code-suggestions-beta/

src/server/helpers/collectionRouteUtils.js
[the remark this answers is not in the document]
prabalsingh24: Oh wait! oops. I was trying it wrong it think. I was doing `UserCollection.collaborators.....`

src/server/helpers/middleware.js
MonkeyDo: I would suggest `UserCollectionCollaborator.collaborator()` and `UserCollectionCollaborator.collection()`

Looking at `Editor.collaboratesOn.collection`, you could skip a step: with the ORM, you can return the array of `collection` models an editor collaborates on directly, with `[.through()`](https://bookshelfjs.org/api.html#Collection-instance-through).

So you would have instead `editor.collaboratesOn()` which would return a collection of UserCollection items.
Same thing for `UserCollection.collaborators()` which would return a collection of Editor items.
Here's an example where this is used in the ORM: https://github.com/bookbrainz/bookbrainz-data-js/blob/fc02949c3426a7eb213ad68df6c544045709fbf9/src/models/editor.js#L25

prabalsingh24:
Screenshot_2020-05-27 BookBrainz – The Open Book Database

src/server/helpers/collectionRouteUtils.js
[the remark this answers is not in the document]
prabalsingh24: Umm no `attach` works only with `belongsToMany` relation. :(

src/server/routes/collection.js
```diff
@@ -0,0 +1,86 @@
...
+import target from '../templates/target';
+
+
+const router = express.Router();
+
+/* If the route specifies a CollectionId, load the Collection for it. */
+router.param(
+	'collectionId',
+	middleware.makeCollectionLoader()
+);
```
MonkeyDo, line 38: You can move this after the /create route; that way you're sure it only looks for collectionID parameter in the routes that need it (I recently changes this in the entity routes in PR #429 )

src/server/routes/collection.js
```diff
@@ -0,0 +1,86 @@
...
+		</Layout>
+	);
+	res.send(target({
+		markup,
+		props: escapeProps(props),
+		script
+	}));
+});
+
+router.post('/create/handler', auth.isAuthenticatedForHandler, makeCollectionCreateOrEditHandler);
```
MonkeyDo, line 78: Regarding the comment above, you'll want to move this handler before the collectionId param routes

src/server/routes/search.js
```diff
@@ -134,11 +134,6 @@ router.get('/reindex', auth.isAuthenticated, (req, res) => {
 		// TODO: This is hacky, and we should replace it once we switch to SOLR.
 		const trustedUsers = ['Leftmost Cat', 'LordSputnik', 'Monkey', 'iliekcomputers'];
 
-		const NO_MATCH = -1;
-		if (trustedUsers.indexOf(req.user.name) === NO_MATCH) {
-			throw new error.PermissionDeniedError(null, req);
-		}
-
```
MonkeyDo, line 141: Did you mean to commit this?

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+		this.handleSubmit = this.handleSubmit.bind(this);
+		this.isValid = this.isValid.bind(this);
+	}
+
+	handleSubmit(evt) {
+		evt.preventDefault();
+
+		if (!this.isValid()) {
+			this.setState(prevState => ({
+				collection: prevState.collection,
```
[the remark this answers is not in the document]
prabalsingh24: It is updating the errorText. it shows the error in the bottom.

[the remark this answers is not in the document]
prabalsingh24: Oh do you mean only update the error text?

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+	handleAddCollaborator() {
+		const newCollaborators = this.state.collection.collaborators;
+		newCollaborators.push({
+			id: null,
+			name: ''
+		});
+		this.setState(prevState => ({
+			collection: {
+				...prevState.collection,
+				collaborators: newCollaborators
```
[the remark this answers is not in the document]
prabalsingh24: Hmm yess. I'll make this change

src/client/components/forms/userCollection.js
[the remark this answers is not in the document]
MonkeyDo: Yes, as far as I know this line `collection: prevState.collection,` isn't going to change anything and should be removed.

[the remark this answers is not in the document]
prabalsingh24: Oh okay. I'll test it can make this change everywhere

src/server/routes/search.js
prabalsingh24: Oops. I'll undo this. I did this because I was changing users and refreshing index while testing. and it was taking time to add user in trustedUsers each time :P.

src/client/components/forms/userCollection.js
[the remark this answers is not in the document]
MonkeyDo: I wonder if it wouldn't be even better to have collaborators separate on the state instead of a property of collection, so that React can optimize rendering better: if collaborators change but not the collection items, no need to redraw whatever uses the collection items. Currently I think the whole collection object in the state gets invalidated in this call.

It will require a bit of destructuring in the constructor.

src/server/routes/search.js
MonkeyDo: Yep, totally understand. It's a far from ideal situation at the moment.
I just assumed you committed it by mistake :)

src/server/helpers/collectionRouteUtils.js
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+import * as handler from '../helpers/handler';
+import * as search from './search';
+import _ from 'lodash';
+
+
+export async function makeCollectionCreateOrEditHandler(req, res) {
+	try {
+		const {UserCollection, UserCollectionCollaborator} = req.app.local.orm;
```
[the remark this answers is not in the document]
prabalsingh24: Yeah sorry I was testing that `try` `catch` block :P

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+
+		/* eslint-disable react/jsx-no-bind */
+		return (
+			<Grid>
+				<h1>Create a collection</h1>
+				<Row>
+					<Col md={12}>
+						<p className="lead">Create Your Collection</p>
+					</Col>
+				</Row>
```
MonkeyDo, line 185: The title is duplicated ("Create a collection" followed by "Create your collection") needlessly

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+		const initialPrivacy = this.state.collection.public ? 'Public' : 'Private';
+		const initialType = this.state.collection.entityType;
+		const {errorText} = this.state;
+		const errorAlertClass =
+			classNames('text-center', 'margin-top-1', {hidden: !errorText});
+		const submitLabel = this.props.collection.name ? 'UPDATE' : 'CREATE';
+
+		/* eslint-disable react/jsx-no-bind */
+		return (
+			<Grid>
```
MonkeyDo, line 179: The page layout is a bit off. Instead of a Grid -> Row -> Col hierarchy, try div -> Col only
The Grid adds a 'container' class which throws off the layout in some screen sizes.

I find that there is no need for Row + Col and just having your Col elements a level higher works better (I'm attaching a screenshot where you can see the flattened hierarchy
Capture d’écran 2020-06-03 à 15 46 59

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+								base={ReactSelect}
+								defaultValue={initialPrivacy}
+								idAttribute="name"
+								label="Privacy"
+								labelAttribute="name"
+								options={privacyOptions}
+								placeholder="Select Privacy"
+								ref={(ref) => this.privacy = ref}
+							/>
+							{
```
MonkeyDo, line 229: You could add a small separation between the collection details and the collaborators section, like an `<hr>` element, and add a small title for that section.

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+									<div key={collaborator.id}>
+										<EntitySearchFieldOption
+											instanceId="collaboratorSearchField"
+											label="Select Collaborator"
+											name="editor"
+											type="editor"
+											value={collaborator}
+											onChange={(newCollaborator) => this.handleChangeCollaborator(newCollaborator, index)}
+										/>
+										<Button
```
MonkeyDo, line 240: Putting the "remove" buttons on the same line as the collaborator select makes it clear that they are related. You can pass a `buttonAfter` property, which with a small modification to the EntitySearchFieldOption component will be [passed down to CustomInput](https://github.com/bookbrainz/bookbrainz-site/blob/77475b87fe748aecbb5202bc6774188b251051d8/src/client/input.js#L124)

The modification is passing down the props to CustomInput [here](https://github.com/bookbrainz/bookbrainz-site/blob/77475b87fe748aecbb5202bc6774188b251051d8/src/client/entity-editor/common/entity-search-field-option.js#L108): 
```
<CustomInput label={labelElement} tooltipText={this.props.tooltipText}>
<CustomInput label={labelElement} tooltipText={this.props.tooltipText} {...this.props}>
```

With that set up, you can do the following:
```
this.state.collection.collaborators.map((collaborator, index) => {
								const buttonAfter = (
									<Button
										bsSize="small"
										bsStyle="danger"
										type="button"
										onClick={() => this.handleRemoveCollaborator(index)}
									>
										Remove
									</Button>
								);
								return (
									<div key={collaborator.id}>
										<EntitySearchFieldOption
											buttonAfter={buttonAfter}
											instanceId="collaboratorSearchField"
											label="Select Collaborator"
											name="editor"
											type="editor"
											value={collaborator}
											onChange={(newCollaborator) => this.handleChangeCollaborator(newCollaborator, index)}
										/>
									</div>
								);
							})
```

Capture d’écran 2020-06-03 à 16 17 09

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+											bsSize="small"
+											bsStyle="primary"
```
MonkeyDo, line 242:
```suggestion
											bsStyle="danger"
```

You could also add a trash bin icon like in some other pages

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+											onClick={() => this.handleRemoveCollaborator(index)}
+										>
+											Remove
+										</Button>
+									</div>
+								))
+							}
+							<div className="text-center">
+								<Button
+									type="button"
```
MonkeyDo, line 253: How about `bsStyle="primary"`  to avoid the grey buttons?

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+									onClick={this.handleAddCollaborator}
+								>
+									Add Collaborator
```
MonkeyDo, line 256: Currently, selecting a user from the dropdown already selects them without having to press this button correct?
Meaning this button is to add an extra empty collaborator?

If so, then the button should read "Add another collaborator", maybe with a little "plus" sign icon

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+								</Button>
+							</div>
+							<div className="form-group text-center margin-top-1">
+								<Button
+									bsSize="large"
+									bsStyle="primary"
```
MonkeyDo, line 262:
```suggestion
									bsStyle="success"
```

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+			name: entity
+		}));
+		const initialName = this.state.collection.name;
+		const initialDescription = this.state.collection.description;
+		const initialPrivacy = this.state.collection.public ? 'Public' : 'Private';
+		const initialType = this.state.collection.entityType;
+		const {errorText} = this.state;
+		const errorAlertClass =
+			classNames('text-center', 'margin-top-1', {hidden: !errorText});
+		const submitLabel = this.props.collection.name ? 'UPDATE' : 'CREATE';
```
MonkeyDo, line 175: Why all caps?
I would say "Update collection" and "Create collection", since we have all the space in the world

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+					</Col>
+				</Row>
+				<Row>
+					<Col
+						id="collectionForm"
+						md={8}
+						mdOffset={2}
+					>
+						<form
+							className="form-horizontal"
```
MonkeyDo, line 193: I would replace this with `padding-sides-0`, which I prevented some differences in padding between form elements and buttons/h elements outside of from-groups.

prabalsingh24:
> Here's a review of the user interface, with a few improvements.

@MonkeyDo Thank you for the review. I will make these changes :)

src/server/helpers/collectionRouteUtils.js
```diff
@@ -0,0 +1,97 @@
...
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as handler from '../helpers/handler';
+import * as search from './search';
+import _ from 'lodash';
+
+
+export async function makeCollectionCreateOrEditHandler(req, res) {
```
MonkeyDo, line 24: Contrarily to the other makeXXXXHandler, this one does isn't a function factory that returns a function when invoked.
So `make` here is misleading.

src/server/helpers/collectionRouteUtils.js
```diff
@@ -0,0 +1,97 @@
...
+			newCollection = await new UserCollection({
+				ownerId: req.user.id
+			});
+			method = 'insert';
+		}
+		else {
+			newCollection = await new UserCollection({id: req.params.collectionId}).fetch({
+				require: true
+			});
+			method = 'update';
```
MonkeyDo, line 40: In this else block I guess you should also set lastModified with the current timestamp, otherwise it is not updated.

src/client/components/forms/userCollection.js
```diff
@@ -0,0 +1,306 @@
...
+
+import * as bootstrap from 'react-bootstrap';
+import CustomInput from '../../input';
+import EntitySearchFieldOption from '../../entity-editor/common/entity-search-field-option';
+import PropTypes from 'prop-types';
+import React from 'react';
+import ReactSelect from 'react-select';
+import SelectWrapper from '../input/select-wrapper';
+import classNames from 'classnames';
+import request from 'superagent-bluebird-promise';
```
MonkeyDo, line 28: I just merged #438 which updates a few packages, including superagent that now doesn't need the extra 'superagent-bluebird-promise' package anymore.

You will also need to remove the `.promise()` call after `request.post…`.
Here's an example: https://github.com/bookbrainz/bookbrainz-site/pull/438/files#diff-e8473140e224bec0384c75907f7aee19

src/client/components/forms/userCollection.js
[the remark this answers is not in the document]
prabalsingh24: or this can be done too 
```
<UserCollectionForm
	canEditType={canEditType}
	collection={props.collection}
        collaborator={props.collaborator}
/>

File: src/client/components/forms/userCollection.js [hunk 23]
```diff
@@ -0,0 +1,291 @@
...
+		else {
+			submissionURL = '/collection/create/handler';
+		}
+		request.post(submissionURL)
+			.send(data)
+			.then((res) => {
+				window.location.href = `/collection/${res.body.id}`;
+			}, (error) => {
+				this.setState({
+					errorText: 'Internal Error'
```
prabalsingh24, line 92: I don't think we should show the internal error to the user. ( it won't make any sense to them anyway)
we can do print the error in the logs. Will be useful while debugging

src/client/components/forms/userCollection.js
prabalsingh24: Thank you for thiss. I forgot about `buttonAfter` I was trying to adjust the button to the right side by `col-md-10`  `col-md-2`, but that way it was looking very bad

src/server/routes/search.js
MonkeyDo: This change hasn't been removed.

src/server/helpers/collectionRouteUtils.js
MonkeyDo: I see you commented out the last_modified update below. What's the reason?

src/client/components/forms/userCollection.js
MonkeyDo: I still think a separation between the main body of the form and the collaborators will be helpful.

prabalsingh24:
Screenshot_2020-06-09 BookBrainz – The Open Book Database(1)
 I didn't add `<hr/>`. Is this alright?

prabalsingh24:
I didn't add `console.error(`, Had to suppress `eslint` for that.

PR #434, opened 2020-05-31
Collections: Editor Profile Tab

prabalsingh24:
### Problem
`collections` tab in user -profile

### Solution
Screenshot_2020-05-31 prabalsinghbhandari's Collections – BookBrainz


src/server/routes/editor.js
```diff
@@ -368,6 +369,93 @@ router.get('/:id/revisions/revisions', async (req, res, next) => {
...
+			qb.whereIn('id', collectionIds)
+				.orWhere('owner_id', req.params.id);
+			qb.orderBy('name');
+		}).fetchPage({
+			limit: size,
+			offset: from,
+			withRelated: [
+				'items'
+			]
+		});
```
prabalsingh24, line 395: I think this can done in one DB query, instead of two here. I will have a look once.

prabalsingh24:
Also what do you think of another column which tells you whether you're the `owner` or the `collaborator` of the said collection

src/server/routes/editor.js
prabalsingh24: Also I am thinking of adding `entity_type` in params. That way collections of a particular type can be fetched. 
<br>
This route then can be used in `Add Entity to Collection` in the search page.
`editor/:id/collections/collections?type=edition` this will return a JSON object with list of editon-collection for the user

src/client/components/pages/parts/editor-collections-table.js
```diff
@@ -0,0 +1,97 @@
...
+				results.length > 0 ?
+					<Table
+						responsive
+						className={tableCssClasses}
+					>
+						<thead>
+							<tr>
+								<th className="col-sm-2">Collection Name</th>
+								<th className="col-sm-5">Description</th>
+								<th className="col-sm-3">EntityType</th>
```
MonkeyDo, line 48:
```suggestion
								<th className="col-sm-3">Entity Type</th>
```

src/client/components/pages/parts/editor-collections-table.js
```diff
@@ -0,0 +1,97 @@
...
+											<a
+												href={`/collection/${collection.id}`}
+											>
+												{collection.name}
+											</a>
+										</td>
+										<td>{collection.description}</td>
+										<td>{collection.entityType}</td>
+										<td>{collection.public ? 'Public' : 'Private'}</td>
+										<td>{collection.lastModified.toString()}</td>
```
MonkeyDo, line 68: Not an ideal display format, IMO.
Look for the `formatDate` utility in src/client/helpers/utils.js

src/client/components/pages/parts/editor-collections-table.js
```diff
@@ -0,0 +1,97 @@
...
+			<hr className="thin"/>
+			{
+				results.length > 0 ?
+					<Table
+						responsive
+						className={tableCssClasses}
+					>
+						<thead>
+							<tr>
+								<th className="col-sm-2">Collection Name</th>
```
MonkeyDo, line 46:
```suggestion
								<th className="col-sm-2">Name</th>
```

src/server/routes/editor.js
MonkeyDo: OK, let me know when I should review, after you've looked into doing a single query.


I think the addition of `entity_type` (or just `type`) could be useful indeed.

src/server/routes/editor.js
```diff
@@ -368,6 +369,93 @@ router.get('/:id/revisions/revisions', async (req, res, next) => {
...
+		res.send(target({
+			markup,
+			page: 'collections',
+			props: escapeProps(props),
+			script: '/js/editor/editor.js',
+			title: `${props.editor.name}'s Collections`
+		}));
+	}
+	catch (err) {
+		return next(err);
```
MonkeyDo, line 443:
```suggestion
		next(err);
```

That way you should be able to remove the  `// eslint-disable-next-line consistent-return` above.

src/server/routes/editor.js
```diff
@@ -368,6 +369,93 @@ router.get('/:id/revisions/revisions', async (req, res, next) => {
...
+// eslint-disable-next-line consistent-return
+router.get('/:id/collections/collections', async (req, res, next) => {
+	try {
+		const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+		const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+		const orderedCollections = await getOrderedCollections(from, size, req);
+		res.send(orderedCollections);
+	}
+	catch (err) {
+		return next(err);
```
MonkeyDo, line 456:
```suggestion
		next(err);
```
Same comment as above about eslint

src/server/routes/editor.js
prabalsingh24: `// eslint-disable-next-line consistent-return` this needs to be added if it is done like that. :(

src/server/routes/editor.js
```diff
@@ -512,4 +512,102 @@ router.post('/:id/achievements/', auth.isAuthenticated, (req, res) => {
...
+		.where((builder) => {
+			if (!isThisCurrentUser) {
+				builder.where('public', true);
+			}
+			if (entityType) {
+				builder.where('entity_type', entityType);
+			}
+		})
+		.where((builder) => {
+			builder.where('editor_id', '=', req.params.id).orWhere('owner_id', '=', req.params.id);
```
prabalsingh24, line 535: I am wondering if `user_collection_collaborator.editor_id` should be changed to `user_collection_collaborator.collaborator_id`

src/server/routes/editor.js
```diff
@@ -512,4 +512,111 @@ router.post('/:id/achievements/', auth.isAuthenticated, (req, res) => {
...
+// eslint-disable-next-line consistent-return
+router.get('/:id/collections', async (req, res, next) => {
+	const {Editor, TitleUnlock} = req.app.locals.orm;
+	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
+	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+	let type = req.query.type ? req.query.type : null;
+	const entityTypes = _.keys(utils.getEntityModels(req.app.locals.orm));
+	if (!entityTypes.includes(type)) {
+		type = null;
+	}
```
prabalsingh24, line 560: this is required because `entity_type` in the table `User_Collection` can only take valid values. Otherwise it throws an error

prabalsingh24:
@MonkeyDo Let me know if this is alright. I will then start writing tests for `getOrderedCollection` function

src/server/helpers/collections.js
```diff
@@ -0,0 +1,75 @@
...
+				builder.where('public', true);
+			}
+			if (entityType) {
+				builder.where('entity_type', entityType);
+			}
+		})
+		.where((builder) => {
+			builder.where('editor_id', '=', req.params.id).orWhere('owner_id', '=', req.params.id);
+		})
+		.fetchPage({
```
prabalsingh24, line 63: I am wondering if in the `user_collection_collaborator` table , `user_collection_collaborator.editor_id` should be changed to `user_collection_collaborator.collaborator_id`

prabalsingh24: because in `user_collection` table, we have `owner_id`

MonkeyDo: Sure, it makes sense.
In other tables we also rename `editor_id` to something else that matches the table (like `revision.author_id` for example)

prabalsingh24:
@MonkeyDo Have a look at this. I added `public collection` page too.
Tests here failed because bookbrainz-data wasn't updated..
Reminder: You'll have to update test database and bookbrainz-data in your local machine.

prabalsingh24:
Also i tried looking for an fontawesome icon for collections. Couldn't find anything nice :((

src/client/containers/layout.js
```diff
@@ -195,6 +195,11 @@ class Layout extends React.Component {
 						{' Revisions '}
 					</NavItem>
 				</Nav>
+				<Nav pullRight>
+					<NavItem href="/collections">
+						{' Collections '}
```
MonkeyDo, line 200: I would like to see this menu item come *after* "Revisions" in the menu bar.
I believe in the code the order is reversed (because of the pullRight attributes), so in code it would need to be *before* Revisions

src/client/components/pages/parts/collections-table.js
```diff
@@ -0,0 +1,173 @@
...
+											{
+												showOwnerCollaborator ?
+													<td>{collection.isOwner ? 'Owner' : 'Collaborator'}</td> : null
+											}
+											{
+												showOwner ?
+													<td>{collection.owner.name}</td> : null
+
+											}
+										</tr>
```
MonkeyDo, line 136: What do you think about showing the last modified datetime here, using the formatDate utility?

src/client/components/pages/collections.js
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	render() {
+		return (
+			<div id="pageWithPagination">
+				<EditorCollectionsTable
+					entityTypes={this.props.entityTypes}
+					results={this.state.results}
+					showOwner={this.props.showOwner}
+					showOwnerCollaborator={this.props.showOwnerCollaborator}
```
MonkeyDo, line 58: This property could be named more explicitly, something like `showIfOwnerOrCollaborator`

test/src/server/helpers/collections.js
```diff
@@ -0,0 +1,337 @@
...
+				'id',
+				'ownerId',
+				'name',
+				'description',
+				'entityType',
+				'public',
+				'createdAt',
+				'lastModified',
+				'owner'
+			);
```
MonkeyDo, line 104: Instead of checking all the keys (which you already do in the previous test), you should test that collection.entityType === "author" here, as that's what's important for this test.

test/src/server/helpers/collections.js
```diff
@@ -0,0 +1,337 @@
...
+			user: {
+				id: editor.get('id')
+			}
+		};
+
+		const orderedCollections = await getOrderedCollectionsForEditorPage(0, 10, null, req);
+
+		expect(orderedCollections.length).to.equal(5);
+		orderedCollections.forEach((collection) => {
+			expect(collection).to.have.keys(
```
MonkeyDo, line 178: I don't think this is necessary in any of the `getOrderedCollectionsForEditorPage` tests you currently have.
The expect(…).to.have.keys part would instead be in a specific separate test like you did for `getOrderedPublicCollections` ("should return collections with expected keys").

Instead, I'd like you to ensure that the collections that `getOrderedCollectionsForEditorPage` returns are the ones you expect.
You can compare by the collection ids. If you map `orderedCollections` to get an array of ids, you can compare it with an array composed of the ids of the collections you're expecting to be returned.
That will make for much more robust tests, as currently you're only comparing the length of the array of returned values.

src/server/routes/collections.js
```diff
@@ -0,0 +1,106 @@
...
+		});
+
+		/*
+		 * Renders react components server side and injects markup into target
+		 * file object spread injects the app.locals variables into React as
+		 * props
+		 */
+		const markup = ReactDOMServer.renderToString(
+			<Layout {...propHelpers.extractLayoutProps(props)}>
+				<CollectionsPage {...propHelpers.extractChildProps(props)}/>
```
MonkeyDo, line 62: It might be clearer to have tableHeading = "Public Collections" in the props.

src/client/components/pages/collections.js
```diff
@@ -0,0 +1,100 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import EditorCollectionsTable from './parts/collections-table';
```
MonkeyDo, line 19: Why 'Editor' in here, rather than just `CollectionsTable` ?

src/server/routes/editor.js
```diff
@@ -512,4 +513,79 @@ router.post('/:id/achievements/', auth.isAuthenticated, (req, res) => {
...
```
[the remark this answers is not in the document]
prabalsingh24: Hm yess I forgot about thiss

PR #435, opened 2020-06-01
UserCollection: Add to collections from search results

prabalsingh24:
### Problem
UserCollection: Add to collections from search results

### Solution
1. Changed the `result-table` to react class
2. Added checkboxes feature

### DIFF
--- a/src/client/components/pages/parts/add-to-collection-modal.js
+++ b/src/client/components/pages/parts/add-to-collection-modal.js
@@ -36,14 +36,14 @@ class AddToCollectionModal extends React.Component {
 	async componentDidMount() {
 		const collections = this.props.entityType ? await this.getCollections() : [];
 		// eslint-disable-next-line react/no-did-mount-set-state
-		this.setState({collectionsAvailable: collections});
+		this.setState({collectionsAvailable: collections || []});
 	}
 
 	async componentDidUpdate(prevProps) {
 		if (prevProps.entityType !== this.props.entityType) {
 			const collections = this.props.entityType ? await this.getCollections() : [];
 			// eslint-disable-next-line react/no-did-update-set-state
-			this.setState({collectionsAvailable: collections});
+			this.setState({collectionsAvailable: collections || []});
 		}
 	}
 
--- a/src/client/components/pages/parts/add-to-collection-modal.js
+++ b/src/client/components/pages/parts/add-to-collection-modal.js
@@ -34,9 +34,17 @@ class AddToCollectionModal extends React.Component {
 	}
 
 	async componentDidMount() {
-		const collections = await this.getCollections();
+		const collections = this.props.entityType ? await this.getCollections() : [];
 		// eslint-disable-next-line react/no-did-mount-set-state
 		this.setState({collectionsAvailable: collections});
+	}
+
+	async componentDidUpdate(prevProps) {
+		if (prevProps.entityType !== this.props.entityType) {
+			const collections = this.props.entityType ? await this.getCollections() : [];
+			// eslint-disable-next-line react/no-did-update-set-state
+			this.setState({collectionsAvailable: collections});
+		}
 	}
 
 	toggleRow(collectionId) {
--- a/src/client/components/pages/parts/search-results.js
+++ b/src/client/components/pages/parts/search-results.js
@@ -21,6 +21,7 @@ import * as bootstrap from 'react-bootstrap';
 
 import {differenceBy as _differenceBy, kebabCase as _kebabCase, startCase as _startCase} from 'lodash';
 
+import AddToCollectionModal from './add-to-collection-modal';
 import CallToAction from './call-to-action';
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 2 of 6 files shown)


prabalsingh24:
Screenshot_2020-06-01 Search Results – BookBrainz

prabalsingh24:
@MonkeyDo Have a look at this :)

src/client/components/pages/parts/search-results.js
```diff
@@ -21,122 +21,257 @@ import * as bootstrap from 'react-bootstrap';
...
+						</thead>
+					}
+					<tbody>
+						{results}
+					</tbody>
+					{
+						this.props.user ?
+							<Button
+								bsSize="small"
+								bsStyle="danger"
```
MonkeyDo, line 251:
```suggestion
								bsStyle="primary"
```

src/client/components/pages/parts/search-results.js
```diff
@@ -21,122 +21,257 @@ import * as bootstrap from 'react-bootstrap';
...
+							</tr>
+						</thead>
+					}
+					<tbody>
+						{results}
+					</tbody>
+					{
+						this.props.user ?
+							<Button
+								bsSize="small"
```
MonkeyDo, line 250: Any particular reason for having a small button?

src/client/components/pages/parts/search-results.js
```diff
@@ -21,122 +21,257 @@ import * as bootstrap from 'react-bootstrap';
...
+								<th className="col-sm-3">Type</th>
+								<th className="col-sm-5">Name</th>
+								<th className="col-sm-4">Aliases</th>
+							</tr>
+						</thead>
+					}
+					<tbody>
+						{results}
+					</tbody>
+					{
```
MonkeyDo, line 247: This whole block should be after the closing `</Table>` tag.

You'll be able to get rid of the whole `className="pull-left margin-top-d5"` on the button too.


I tried with an `<hr className='thin'/>` in-between the table and the button, too. I'll let you decide which one looks best in your opinion:

FireShot Capture 101 - Search Results – BookBrainz - localhost
FireShot Capture 102 - Search Results – BookBrainz - localhost

src/client/components/pages/parts/search-results.js
```diff
@@ -21,122 +21,257 @@ import * as bootstrap from 'react-bootstrap';
...
-					!props.condensed &&
-					<thead>
-						<tr>
-							<th style={{width: '150px'}}>Type</th>
-							<th>Name</th>
-							<th>Aliases</th>
-						</tr>
-					</thead>
+					this.state.message.text ?
+						<Alert bsStyle={this.state.message.type} onDismiss={this.handleAlertDismiss}>{this.state.message.text}</Alert> : null
```
MonkeyDo, line 262: This could probably benefit from a `className='margin-top-1'` to have some space between the button and the alert.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -34,11 +34,19 @@ class AddToCollectionModal extends React.Component {
...
+		const collections = this.props.entityType ? await this.getCollections() : [];
 		// eslint-disable-next-line react/no-did-mount-set-state
 		this.setState({collectionsAvailable: collections});
 	}
 
+	async componentDidUpdate(prevProps) {
+		if (prevProps.entityType !== this.props.entityType) {
+			const collections = this.props.entityType ? await this.getCollections() : [];
+			// eslint-disable-next-line react/no-did-update-set-state
+			this.setState({collectionsAvailable: collections});
```
MonkeyDo, line 46: `collections` could potentially be `undefined` here (admittedly in rare cases where a call to `getCollections` fails).

If so, further down a check on `this.state.collectionsAvailable.length` will throw an error `Cannot read property 'length' of undefined`.

So we either have to ensure `collectionsAvailable` is never set to `undefined`, or check if it is `undefined` wherever we use it.
I suggest this simple method :
```suggestion
			this.setState({collectionsAvailable: collections || []});
```

And the same should apply for `componentDidMount`.

src/client/components/pages/parts/search-results.js
MonkeyDo: Also I think it would be consistent to add the collections icon in this button

src/client/components/pages/parts/search-results.js
```diff
@@ -21,122 +21,279 @@ import * as bootstrap from 'react-bootstrap';
...
+				}
+				{
+					this.props.user ?
+						<ButtonGroup>
+							<Button
+								bsStyle="primary"
+								type="button"
+								onClick={this.handleAddToCollection}
+							>
+								<FontAwesomeIcon icon="grip-vertical"/>
```
MonkeyDo, line 272: Let's use this utility here: `{genEntityIconHTMLElement('Collection')}`

src/client/components/pages/parts/search-results.js
```diff
@@ -21,122 +21,279 @@ import * as bootstrap from 'react-bootstrap';
...
+							{
+								this.state.selected.length ?
+									<Button
+										bsStyle="danger"
+										type="button"
+										onClick={this.handleClearSelected}
+									>
+										Clear <span className="circle-number">{this.state.selected.length}</span> selected
+									</Button> : null
+							}
```
MonkeyDo, line 284: A couple of suggestions for this button:
- using the `<Badge>` element for Bootstrap for the number
- making the button disabled instead of removing it if there is no selection (i'm of the opinion we should do the same for the "add to collection" button)
- changing the color (to keep red for more important destructive actions)

```suggestion
							<Button
								bsStyle="warning"
								disabled={!this.state.selected.length}
								type="button"
								onClick={this.handleClearSelected}
							>
								Clear <Badge>{this.state.selected.length}</Badge> selected
							</Button>
```

src/client/stylesheets/style.less
```diff
@@ -493,3 +493,11 @@ hr.wide {
...
 	vertical-align: text-top;
 }
+
+.circle-number {
+	width: 25px;
+	color: #c64040;
+	background-color: #ffffff;
+	border-radius: 50%;
+	display: inline-block;
+}
```
MonkeyDo, line 503: Using Bootstrap's `<Badge>` element, this should be unused.

PR #438, opened 2020-06-04
Fix audit

leftmostcat:
### Problem

npm lists numerous audit issues.

### Solution

Address audit issues. Additionally, update dependencies and remove unneeded.

### Areas of Impact

General.


package.json
```diff
@@ -41,10 +41,9 @@
     "@fortawesome/free-brands-svg-icons": "^5.11.2",
     "@fortawesome/free-solid-svg-icons": "^5.11.2",
     "@fortawesome/react-fontawesome": "^0.1.7",
-    "babel-plugin-transform-require-ignore": "^0.1.1",
```
leftmostcat, line 44: Moved to dev dependencies.

src/client/entity-editor/submission-section/actions.js
```diff
@@ -81,8 +80,7 @@ function postSubmission(url: string, data: Map<string, mixed>): Promise {
 	 */
 
 	const [, submissionEntity] = url.split('/');
-	return request.post(url).send(data)
-		.promise()
+	return request.post(url).send(Object.fromEntries(data))
```
leftmostcat, line 83: Either `superagent` or `superagent-bluebird-promise` seems to have coerced `Map`s to plain JS objects previously, but this isn't the case with the update, so we do it ourselves.

src/server/helpers/entityRouteUtils.js
```diff
@@ -247,7 +247,7 @@ export function makeEntityCreateOrEditHandler(
 	entityType: string,
 	transformNewForm: Function,
 	propertiesToPick: string | string[],
-	isMergeHandler?: boolean = false
+	isMergeHandler: boolean = false
```
leftmostcat, line 250: Gives off an error in the linter when both `?` and a default value are provided.

PR #461, opened 2020-06-19
fix BB-509 docs(api): Transform API docs from Swagger2 to OpenAPI3

MonkeyDo:
See https://tickets.metabrainz.org/browse/BB-509


MonkeyDo:
Hm, it looks like mutually exclusive parameters are not available in OpenAPI 3.
https://github.com/OAI/OpenAPI-Specification/issues/256

This would be for the browse requests: allow for example only one of `/author?author=$UUID` OR `/author?edition=$UUID`, etc.

PR #463, opened 2020-06-22
Add author credit display and edit for Editions

leftmostcat:
### Problem

Author credits are supported for Editions in the database, but there is no way to create or display them.

### Solution

Add display and edit capabilities for Edition author credits.

### Areas of Impact

Entity editors, particularly Edition. Edition display.


src/client/components/pages/entities/title.js
```diff
@@ -35,7 +35,6 @@ function EntityTitle({entity}) {
 		<div>
 			<h1>{label}{disambiguation}</h1>
 			{aliases}
-			<hr/>
```
leftmostcat, line 38: To allow displaying author credits immediately below edition title without intervening rule.

src/client/entity-editor/author-credit-editor/author-credit-section.js
```diff
@@ -0,0 +1,128 @@
...
+
+	const editButton = (
+		<Button bsStyle="success" onClick={onEditAuthorCredit}>
+			<FontAwesomeIcon icon="pencil-alt"/>
+			&nbsp;Edit
+		</Button>);
+
+	const label = (
+		<ValidationLabel error={!isValid}>
+			Author Credit
```
MonkeyDo, line 81: I wonder if this is going to be clear enough for new users.
Should we have something like "Author(s):" or "Who is the author?" / "Who are the authors?" instead?

We could keep the more formal "Author Credit" in the AC editor I suppose, and tie it back to the question “who's the author” that way.

src/client/components/author-credit-display.js
```diff
@@ -0,0 +1,49 @@
...
+			<a href={`/author/${name.authorBBID}`}>
+				{name.name}
+			</a>
+			{name.joinPhrase}
+		</span>
+	));
+
+	return (
+		<span>
+			{nameElements}
```
MonkeyDo, line 36: What do you think of `by {nameElements}`?

src/client/entity-editor/author-credit-editor/author-credit-section.js
leftmostcat: I'll work on something to clarify this.

src/client/components/author-credit-display.js
leftmostcat: This actually reminds me of something I'd like to have longer-term; I'd like to be able to have lead-in text for an author credit. There are a lot of cases out there where the credit is something like "Edited by Foo", and it'd be nice to be able to represent that in data. With the potential for lead-in text, the "by" would be out-of-place, but the lead-in text itself bears further discussion.

MonkeyDo: Looking at the `author_credit` table in the database, there is a `begin_phrase` text column that seems to be there for that.
It's not part of the sql schema though; I'm not sure when it was added but it was removed here: https://github.com/bookbrainz/bookbrainz-site/commit/7cc6f6fdad569e5595ba622321b4a7e5535b1f19
Not sure why, it seems like it would be useful.

MonkeyDo:
@leftmostcat I've added a very basic diff formatting for ACs, if you want to have a look.
It's not ideal, but with the current way we do diffs I went with the easy implementation.

Ultimately, we'll probably want to display AC changes differently, but for now this gets the PR out the door.

New AC
Capture d’écran 2020-06-25 à 21 21 53

New AC with multiple authors
Capture d’écran 2020-06-25 à 21 26 51


Modified author name
Capture d’écran 2020-06-25 à 21 22 02

Modified 2 author names and added another author
Capture d’écran 2020-06-25 à 21 22 51

Modified author BBID
Capture d’écran 2020-06-25 à 21 25 24

MonkeyDo:
I forgot a modification I made locally to bookbrainz-data-js, so the diffs won't be working at the moment.

MonkeyDo:
Hi @leftmostcat !
I'd like to move this feature along, and I was wondering if I could do anything to help (including finish what's left to finish, if you don't currently have time for it).
Please let me know, I'm eager to have author credits and have a thorough community discussion about how to use them and on which entities.

MonkeyDo:
A day before the two-year anniversary of this PR (🎉  🍰 ) it's ready to be merged and deployed on beta !
Good group effort ! Thank you so much @leftmostcat  and @tr1ten  !

PR #464, opened 2020-06-22
Delete collections

prabalsingh24:
### Problem
Added the feature to delete a collection


prabalsingh24:
I noticed I forgot to add `ON DELETE CASCADE` in the schema. Added that in this PR. You'll have to create table in your pc again (Hopefully last time :P)

prabalsingh24:
If we add `Delete` button in the `Edit`. Then for a user, the flow would be  `collection` --> `Edit` --> `Delete`
I was planning the flow to be `collection` --> `Delete Collection` or `collection` --> `Remove Collaborator` (depending whether user is owner or collaborator)

Also I think we'll have to make new page for `Remove Yourself as collaborator` anyway.
I was thinking of reusing this page for `Remove yourself as a collaborator` as well.

MonkeyDo:
> I was planning the flow to be collection --> Delete Collection
Well I think we could have the same deletion confirmation modal (both for collection deletion and being removed from collaborators) in the collection page rather than the collection edit page then, no?
Meaning having a "delete"/"remove from collaborators" button in the collection page that opens a reusable modal?

prabalsingh24:
@MonkeyDo What do you think of this?


Screenshot_2020-06-25 BookBrainz – The Open Book Database(1)

src/client/components/forms/userCollection.js
```diff
@@ -160,6 +172,24 @@ class UserCollectionForm extends React.Component {
...
 			<div>
 				<h1>Create your collection</h1>
+				<DeleteCollectionModal
+					closeDeleteModalCallback={this.closeDeleteModalCallback}
+					collection={this.props.collection}
+					show={this.state.showModal}
+				/>
+				{
+					this.props.collection.id ?
+						<Button
```
MonkeyDo, line 182: The placement of this button looks a bit random at the moment, and in some screen sizes something comes over the button and prevents me from clicking it.
I think it would make more sense to group it with the submit button at the bottom of the page, in a `<ButtonGroup>` element.

src/client/components/forms/userCollection.js
```diff
@@ -160,6 +172,24 @@ class UserCollectionForm extends React.Component {
...
+							bsSize="small"
+							bsStyle="danger"
+							className="pull-right"
+							type="button"
+							onClick={this.handleShowDeleteModal}
+						>
+							<FontAwesomeIcon icon="times"/>&nbsp;Delete this collection
```
MonkeyDo, line 189: I'd say the trash icon would be better here for consistency, and I'd remove `this` to match the grammar of the other button

src/client/components/pages/parts/delete-collection-modal.js
```diff
@@ -0,0 +1,117 @@
...
+						Make sure you actually want to delete this Collection <br/>
+						There is no way to undo this.
+					</p>
+					{errorComponent}
+				</Modal.Body>
+				<Modal.Footer>
+					<Button onClick={this.handleClose}>
+						<FontAwesomeIcon icon="times-circle"/> Cancel
+					</Button>
+					<Button onClick={this.handleDelete}>
```
MonkeyDo, line 81:
```suggestion
					<Button bsStyle="danger" onClick={this.handleDelete}>
```

src/client/components/pages/parts/delete-collection-modal.js
```diff
@@ -0,0 +1,117 @@
...
+						</h4>
+					</Alert>
+					<p>
+						Make sure you actually want to delete this Collection <br/>
+						There is no way to undo this.
+					</p>
+					{errorComponent}
+				</Modal.Body>
+				<Modal.Footer>
+					<Button onClick={this.handleClose}>
```
MonkeyDo, line 78:
```suggestion
					<Button bsStyle="info" onClick={this.handleClose}>
```

I would also suggest removing the `times-circle` icon entirely, just so that it stays absolutely clear which button actually deletes the collection.  What do you think?

Capture d’écran 2020-06-25 à 12 15 00

src/client/components/pages/parts/delete-collection-modal.js
```diff
@@ -0,0 +1,117 @@
...
+				</Modal.Footer>
+			</Modal>
+		);
+	}
+}
+
+
+DeleteCollectionModal.displayName = 'DeleteCollectionModal';
+DeleteCollectionModal.propTypes = {
+	closeDeleteModalCallback: PropTypes.func.isRequired,
```
MonkeyDo, line 93: I would call this property `onClose` in this component, where the `DeleteModal` part is redundant, and the `on…` format for callbacks follows what we do in other components

src/client/components/forms/userCollection.js
MonkeyDo: While you're in there, you could also add a "save" icon to the submit button :)

src/client/components/forms/userCollection.js
MonkeyDo: I've also seen before the use of elipsis (`…`) on delete buttons (and often in menu items) to suggest there is a further action, or at least that it's not an instant deletion. What do you think about that?

src/client/components/pages/parts/delete-collection-modal.js
```diff
@@ -0,0 +1,117 @@
...
+					<Alert bsStyle="warning">
+						<h4>
+							<FontAwesomeIcon icon="exclamation-triangle"/>&nbsp;
+							You’re about to delete the Collection - {collection.name}.
+						</h4>
+					</Alert>
+					<p>
+						Make sure you actually want to delete this Collection <br/>
+						There is no way to undo this.
+					</p>
```
MonkeyDo, line 74: I would just put this paragraph in the alert, under the h4 element.

src/client/components/pages/parts/delete-collection-modal.js
```diff
@@ -0,0 +1,117 @@
...
+				<Alert bsStyle="danger">{this.state.error}</Alert>;
+		}
+
+		return (
+			<Modal
+				show={this.state.show}
+				onHide={this.handleClose}
+			>
+				<Modal.Header closeButton>
+					<Modal.Title>Confirm Deletion</Modal.Title>
```
MonkeyDo, line 62: And a tiny nitpick :) : 
```suggestion
					<Modal.Title>Confirm deletion</Modal.Title>
```

prabalsingh24:
What do you think of this?

Screenshot_2020-07-01 BookBrainz – The Open Book Database

prabalsingh24:
Screenshot_2020-07-01 BookBrainz – The Open Book Database(1)

PR #465, opened 2020-06-22
Pay down some tech debt

leftmostcat:
### Problem

There's a fair bit of tech debt lying around.

### Solution

Take an initial crack at paying it down. Delete unused or duplicated code. Clean up lints. Update some doc comments. Do some style work. Introduce subtle yet horrifying bugs.

### Areas of Impact

General.


src/client/components/pages/search.js
```diff
@@ -67,7 +65,7 @@ class SearchPage extends React.Component {
 		if (this.state.query === fullQuery) {
 			return;
 		}
-		this.setState({from: 0, query: fullQuery});
+		this.setState({query: fullQuery});
```
prabalsingh24, line 68: This should not be removed. I can see why you did this but actually there is a bug which needs to be fixed first.

Umm. If you go to this [https://test.bookbrainz.org/search?q=harry+potter&from=10](https://test.bookbrainz.org/search?q=harry+potter&from=10) , you'll notice the pagination is showing `from` = 1. It should be showing `from` = 11

I checked and saw `from` is missing in the props [here](https://github.com/bookbrainz/bookbrainz-site/blob/cb479394f762cb3c22c84653d10b05f49e8ebe2b/src/server/routes/search.js#L62)
```
				entityTypes,
				from,
				hideSearch: true,
				nextEnabled,
				resultsPerPage: size,
				...searchResults
			});
```
<br>
Then this change won't be required.

MonkeyDo: As discussed on IRC, `from` is not needed in the state.
There is another issue that will be solved in a separate PR.

PR #467, opened 2020-06-26
Add entity to collection entity page

prabalsingh24:
### Problem
Add Entity To Collection From Entity Page

### Solution

1. Changed all entityPage to React Class
2. Added `Add To Collection` Modal
3. Created POST `collection/:collectionId/add` route

### DIFF
--- a/src/client/components/pages/entities/footer.js
+++ b/src/client/components/pages/entities/footer.js
@@ -26,13 +26,14 @@ import React from 'react';
 
 const {formatDate} = utilsHelper;
 const {
-	Button, ButtonGroup, Col, Row
+	Alert, Button, ButtonGroup, Col, Row
 } = bootstrap;
 
 class EntityFooter extends React.Component {
 	constructor(props) {
 		super(props);
 		this.state = {
+			error: null,
 			showModal: false
 		};
 
@@ -49,22 +50,30 @@ class EntityFooter extends React.Component {
 			this.setState({showModal: true});
 		}
 		else {
-			window.location.href = '/auth';
+			this.setState({error: 'You need to be logged in for this'});
 		}
 	}
 
 	render() {
 		return (
 			<div>
-				<div>
-					<AddToCollectionModal
-						bbids={[this.props.bbid]}
-						closeCallback={this.onCloseModal}
-						entityType={this.props.entityType}
-						show={this.state.showModal}
-						userId={this.props.user.id}
-					/>
-				</div>
+				{
+					this.props.user ?
+						<div>
+							<AddToCollectionModal
+								bbids={[this.props.bbid]}
+								closeCallback={this.onCloseModal}
+								entityType={this.props.entityType}
+								show={this.state.showModal}
+								userId={this.props.user.id}
+							/>
+						</div> : null
+				}
+				{
+					this.state.error ?
+						<Alert bsStyle="danger">{this.state.error}</Alert> : null
+
+				}
 				<Row>
 					<Col md={8} mdOffset={2}>
 						<ButtonGroup justified>
--- a/src/client/components/pages/entities/footer.js
+++ b/src/client/components/pages/entities/footer.js
@@ -18,7 +18,7 @@
 
 import * as bootstrap from 'react-bootstrap';
 import * as utilsHelper from '../../../helpers/utils';
-
+import AddToCollectionModal from '../parts/add-to-collection-modal';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
 import React from 'react';
@@ -29,70 +29,106 @@ const {
 	Button, ButtonGroup, Col, Row
 } = bootstrap;
 
-function EntityFooter({bbid, deleted, entityUrl, lastModified, handleAddToCollection}) {
-	return (
-		<div>
-			<Row>
-				<Col md={6} mdOffset={3}>
-					<ButtonGroup justified>
-						<Button
-							bsStyle="warning"
-							disabled={deleted}
-							href={`${entityUrl}/edit`}
-							title="Edit Entity"
-						>
-							<FontAwesomeIcon icon="pencil-alt"/>&nbsp;Edit
-						</Button>
-						<Button
-							bsStyle="primary"
-							href={`${entityUrl}/revisions`}
-							title="Revision History"
-						>
-							<FontAwesomeIcon icon="history"/>&nbsp;History
-						</Button>
-						<Button
-							bsStyle="danger"
-							disabled={deleted}
-							href={`${entityUrl}/delete`}
-							title="Delete Entity"
-						>
-							<FontAwesomeIcon icon="times"/>&nbsp;Delete
-						</Button>
-						<Button
-							bsStyle="default"
-							href={`/merge/add/${bbid}`}
-							title="Select entity for merging"
-						>
-							<FontAwesomeIcon flip="vertical" icon="code-branch"/>
-							&nbsp;Merge
-						</Button>
-						<Button
-							bsStyle="default"
-							title="Add To Collection"
-							onClick={handleAddToCollection}
-						>
-							<FontAwesomeIcon flip="plus" icon="code-branch"/>
-							&nbsp;Collection
-						</Button>
-					</ButtonGroup>
-				</Col>
-			</Row>
-			<div className="text-center margin-top-d10">
-				<dl>
-					<dt>Last Modified</dt>
-					<dd>{formatDate(new Date(lastModified))}</dd>
-				</dl>
+class EntityFooter extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+			showModal: false
+		};
+
+		this.onCloseModal = this.onCloseModal.bind(this);
+		this.handleShowModal = this.handleShowModal.bind(this);
+	}
+
+	onCloseModal() {
+		this.setState({showModal: false});
+	}
+
+	handleShowModal() {
+		if (this.props.user) {
+			this.setState({showModal: true});
+		}
+		else {
+			window.location.href = '/auth';
+		}
+	}
+
+	render() {
+		return (
+			<div>
+				<div>
+					<AddToCollectionModal
+						closeCallback={this.onCloseModal}
+						entities={[this.props.entity]}
+						entityType={this.props.entity.type}
+						show={this.state.showModal}
+						user={this.props.user}
+					/>
+				</div>
+				<Row>
+					<Col md={8} mdOffset={2}>
+						<ButtonGroup justified>
+							<Button
+								bsStyle="warning"
+								disabled={this.props.deleted}
+								href={`${this.props.entityUrl}/edit`}
+								title="Edit Entity"
+							>
+								<FontAwesomeIcon icon="pencil-alt"/>&nbsp;Edit
+							</Button>
+							<Button
+								bsStyle="primary"
+								href={`${this.props.entityUrl}/revisions`}
+								title="Revision History"
+							>
+								<FontAwesomeIcon icon="history"/>&nbsp;History
+							</Button>
+							<Button
+								bsStyle="danger"
+								disabled={this.props.deleted}
+								href={`${this.props.entityUrl}/delete`}
+								title="Delete Entity"
+							>
+								<FontAwesomeIcon icon="times"/>&nbsp;Delete
+							</Button>
+							<Button
+								bsStyle="default"
+								href={`/merge/add/${this.props.bbid}`}
+								title="Select entity for merging"
+							>
+								<FontAwesomeIcon flip="vertical" icon="code-branch"/>
+								&nbsp;Merge
+							</Button>
+							<Button
+								bsStyle="primary"
+								href="#"
+								title="Add To Collection"
+								onClick={this.handleShowModal}
+							>
+								<FontAwesomeIcon icon="plus"/>
+								&nbsp;Collection
+							</Button>
+						</ButtonGroup>
+					</Col>
+				</Row>
+				<div className="text-center margin-top-d10">
+					<dl>
+						<dt>Last Modified</dt>
+						<dd>{formatDate(new Date(this.props.lastModified))}</dd>
+					</dl>
+				</div>
 			</div>
-		</div>
-	);
+		);
+	}
 }
 EntityFooter.displayName = 'EntityFooter';
 EntityFooter.propTypes = {
 	bbid: PropTypes.string.isRequired,
 	deleted: PropTypes.bool,
+	entity: PropTypes.object.isRequired,
 	entityUrl: PropTypes.string.isRequired,
-	handleAddToCollection: PropTypes.func.isRequired,
-	lastModified: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)]).isRequired
+	lastModified: PropTypes.oneOfType([PropTypes.string, PropTypes.instanceOf(Date)]).isRequired,
+	user: PropTypes.object.isRequired
 };
 EntityFooter.defaultProps = {
 	deleted: false
--- a/src/client/components/pages/entities/footer.js
+++ b/src/client/components/pages/entities/footer.js
@@ -33,17 +33,20 @@ class EntityFooter extends React.Component {
 	constructor(props) {
 		super(props);
 		this.state = {
-			error: null,
+			message: {
+				text: null,
+				type: null
+			},
 			showModal: false
 		};
 
 		this.onCloseModal = this.onCloseModal.bind(this);
 		this.handleShowModal = this.handleShowModal.bind(this);
+		this.handleAlertDismiss = this.handleAlertDismiss.bind(this);
+		this.closeModalAndShowMessage = this.closeModalAndShowMessage.bind(this);
 	}
 
 	onCloseModal() {
-		// eslint-disable-next-line no-console
-		console.log('i am here');
 		this.setState({showModal: false});
 	}
 
@@ -52,11 +55,26 @@ class EntityFooter extends React.Component {
 			this.setState({showModal: true});
 		}
 		else {
-			this.setState({error: 'You need to be logged in'});
+			this.setState({
+				message: {
+					text: 'You need to be logged in',
+					type: 'danger'
+				}
+			});
 		}
 	}
 
-	/* eslint-disable react/jsx-handler-names */
+	closeModalAndShowMessage(message) {
+		this.setState({
+			message,
+			showModal: false
+		});
+	}
+
+	handleAlertDismiss() {
+		this.setState({message: {}});
+	}
+
 	render() {
 		return (
 			<div>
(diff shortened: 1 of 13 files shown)


prabalsingh24:
@MonkeyDo  Have a look at thiss.
Idk why the buttons are collapsing in the footer. They're still working though.

Screenshot_2020-06-26 Harry Potter (Work) – BookBrainz(1)

prabalsingh24:
what do you think of `checkboxes` instead of numbers and `Add to selected collections` button next to cancel?

prabalsingh24:
Screenshot_2020-06-27 I’m Just Here for More Food (Edition) – BookBrainz(1)

prabalsingh24:
Instead of disabling the button, I did this.. We can do this for `merge` and `edit` too.. 

Screenshot_2020-06-27 Bitters (EditionGroup) – BookBrainz

prabalsingh24:
Screenshot_2020-06-27 A A Long (Author) – BookBrainz(1)

Screenshot_2020-06-27 A A Long (Author) – BookBrainz(2)

prabalsingh24:
Screenshot_2020-06-27 The New Moosewood Cookbook (Edition) – BookBrainz

prabalsingh24:
what do you think of `large` size modal

src/client/components/pages/entities/footer.js
```diff
@@ -18,72 +18,126 @@
...
+	onCloseModal() {
+		this.setState({showModal: false});
+	}
+
+	handleShowModal() {
+		if (this.props.user) {
+			this.setState({showModal: true});
+		}
+		else {
+			this.setState({error: 'You need to be logged in for this'});
```
MonkeyDo, line 53:
```suggestion
			this.setState({error: 'You need to be logged in'});
```

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+	async componentDidUpdate(prevProps) {
+		const collections = await this.getCollections();
+		if (prevProps.show !== this.props.show) {
+			// eslint-disable-next-line react/no-did-update-set-state
+			this.setState({collectionsAvailable: collections, show: this.props.show});
+		}
+	}
+
+	toggleRow(collectionId) {
+		// eslint-disable-next-line react/no-access-state-in-setstate
```
MonkeyDo, line 46: That does not look good. Why is this needed?

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+		this.handleShowCollectionForm = this.handleShowCollectionForm.bind(this);
+		this.handleShowAllCollections = this.handleShowAllCollections.bind(this);
+		this.handleAddToNewCollection = this.handleAddToNewCollection.bind(this);
+		this.isValid = this.isValid.bind(this);
+	}
+
+	async componentDidUpdate(prevProps) {
+		const collections = await this.getCollections();
+		if (prevProps.show !== this.props.show) {
+			// eslint-disable-next-line react/no-did-update-set-state
```
MonkeyDo, line 40: Not 100% sure what's going on here.
Why do you need to store `show` on the state, instead of using `this.props.show` directly?

As for setting collectionsAvailable, I think it would be better to do it in `componentDidMount` instead. I don't really think we need to re-fetch if a user closes and reopens the modal.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+			newSelected = oldSelected.push(collectionId) && oldSelected;
+		}
+		this.setState({
+			selectedCollections: newSelected
+		});
+	}
+
+	async getCollections() {
+		// Get all collections of the user (unlikely that a user will have more than 10000 collections
+		const req = await request.get(`/editor/${this.props.userId}/collections/collections?type=${this.props.entityType}&size=10000`);
```
MonkeyDo, line 62: Why the `size` here? Is it because we have a small default size?
If so, we might want to think about implementing `size=0` to fetch all the collections?

Your comment made me think about what would happen if there are a lot of collections.
Maybe at the end of the project we can think about using a virtualized scroll.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+		const collections = req.body;
+		return collections;
+	}
+
+	handleClose() {
+		this.setState({show: false}, this.props.closeCallback);
+	}
```
MonkeyDo, line 69: To me it would make more sense to just call `onCloseCallback` and letting the parent component set this.props.show and using that directly.

src/client/components/pages/parts/add-to-collection-modal.js
prabalsingh24: Yeah we can use `componentDidMount`. I was using `componentDidUpdate` because earlier I didn't have the feature of creating the collection in the modal. and had to open a new tab to create a new collection. In that case `componentDidUpdate` was required.
But now that we can create collection in the modal itself we don't need to fetch collections everytime. One time is enough

src/client/stylesheets/style.less
```diff
@@ -478,3 +478,13 @@ hr.wide {
 .badge.new {
 	background-color: @green;
 }
+
+.label-checkbox {
+	padding: 0 .5em;
+	vertical-align: .25em;
```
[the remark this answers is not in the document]
prabalsingh24: This is how it is looking for `vertical-align: text-top;`

Screenshot_2020-06-30 W Pustyni i w Puszczy (Work) – BookBrainz

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+						this.state.showCollectionForm ?
+							<Button bsStyle="primary" onClick={this.handleAddToNewCollection}>
+								<FontAwesomeIcon icon="plus"/> Add to this new collection
+							</Button> :
+							<Button bsStyle="primary" onClick={this.handleAddToCollection}>
+								<FontAwesomeIcon icon="plus"/> Add to selected collections
+							</Button>
+					}
+					<Button bsStyle="danger" onClick={this.handleClose}>
+						<FontAwesomeIcon icon="times-circle"/> Cancel
```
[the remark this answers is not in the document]
prabalsingh24: left or right?

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+							)))
+						}
+					</div>
+				</div>
+			);
+		}
+		else {
+			existingCollections = (
+				<div>
+					Oops, Looks like you do not have any collection of type {this.props.entityType} . <br/>
```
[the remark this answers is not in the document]
prabalsingh24: Umm we don't really need to do this `entityType` is already `EditionGroup`

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+					message: {
+						text: 'Internal Error',
+						type: 'danger'
+					}
+				});
+			});
+	}
+
+	isValid() {
+		return this.name.getValue() && this.privacy.getValue();
```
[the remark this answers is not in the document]
prabalsingh24: Hmmm I'll make this change in `userCollectionForm` as well

src/server/routes/collection.js
```diff
@@ -85,4 +84,28 @@ router.get('/:collectionId', (req, res) => {
...
+		const {UserCollectionItem} = req.app.locals.orm;
+		for (const bbid of bbids) {
+			try {
+				await new UserCollectionItem({
+					bbid,
+					collectionId: collection.id
+				}).save(null, {method: 'insert'});
+			}
+			catch (err) {
+				// entity already in the collection
```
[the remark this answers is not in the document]
prabalsingh24: Umm I don't really think we need to log this. Because user can accidentally add same entity multiple times and it's not really a bug. 
It would've been nice if bookshelf had `upsert` option.
The other thing we can do is first check if exists, if not then add it. But that would require 2 calls. And since this is gonna be used alot, I don't think we want this.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,309 @@
...
+			entityType,
+			name,
+			privacy
+		};
+		const {bbids} = this.props;
+		request.post('/collection/create/handler')
+			.send(data)
+			.then((res) => {
+				request.post(`/collection/${res.body.id}/add`)
+					.send({bbids}).then(() => {
```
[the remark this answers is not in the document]
prabalsingh24: Yeah we can refactor it later..

src/client/components/pages/parts/add-to-collection-modal.js
prabalsingh24: Yes it has default 20 size and also that endpoint uses `fetchPage` and we need to pass a size in it, otherwise it takes up default `10` size.

The modal is scrollable, so if there are alot of collectoins, the modal will become very long...

src/client/components/pages/parts/add-to-collection-modal.js
prabalsingh24: Yeah this is not needed. I forgot to remove it

src/client/components/pages/parts/add-to-collection-modal.js
MonkeyDo: OK, let's cross that bridge when we come to it, then.
There is a virtualized scrolling library already installed that will help us with that (we use it for the long list of languages to select from)

src/server/routes/collection.js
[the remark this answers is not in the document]
MonkeyDo: Yes, upsert would be useful here, definitely.
Let's not do the double call, as therse are no dire consequences if adding an entity twice.
The logging I meant more in case there is another type of error.

This can be achieved by checking the type of error, displaying an "entity already added" message if it's a [NoRowsUpdatedError](https://bookshelfjs.org/api.html#Model-static-NoRowsUpdatedError) and logging the error otherwise.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+						this.setState({
+							message: {
+								text: `Successfully added to your new collection: ${name}`,
+								type: 'success'
+							}
+						});
```
MonkeyDo, line 136: This looks nice currently, but a user is allowed to continue clicking on "add to new collection" and it will create as many duplicate collections.
We can avoid that easily by either:
1. closing the modal (and maybe showing a success alert in the entity footer (but what about adding from the search page?) or 
2. disabling the button after successful creation. I think that is more elegant and self-contained.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+					});
+			}, (error) => {
+				this.setState({
+					message: {
+						text: 'Internal Error: Try again later',
```
MonkeyDo, line 141:
```suggestion
						text: 'Something went wrong! Please try again later',
```

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+		this.setState({message: {}, showCollectionForm: false});
+	}
+
+	handleAddToNewCollection(evt) {
+		evt.preventDefault();
+
+		if (!this.isValid()) {
+			this.setState({
+				message: {
+					text: 'Internal Error: Try again later',
```
MonkeyDo, line 107: I liked the "Incomplete form"  message you had previously.
How about "The form is incomplete. Please fill in a name and privacy option before continuing."

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+						this.state.showCollectionForm ?
+							<Button bsStyle="success" onClick={this.handleAddToNewCollection}>
+								<FontAwesomeIcon icon="plus"/> Add to new collection
+							</Button> :
+							<Button bsStyle="success" onClick={this.handleAddToCollection}>
+								<FontAwesomeIcon icon="plus"/>Add to selected collection{this.state.selectedCollections.length > 1 ? 's' : null}
+							</Button>
+					}
+					<Button bsStyle="danger" onClick={this.props.onCloseCallback}>
+						<FontAwesomeIcon icon="times"/> Cancel
```
MonkeyDo, line 278: After some thought and using the modal, I feel like "Close" would be better than "Cancel" since if I added to a or created a collection, the button will not revert that action.
```suggestion
						<FontAwesomeIcon icon="times"/> Close
```


If I close the modal, the collection selection is not cleared (which is good IMO) but the new collection form is cleared. It's because when we toggle the collection rows it's stored in the state, while the name/description/privacy isn't.
I don't think that's an issue in this case, but something to keep in mind I guess.

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+			newSelected = [...oldSelected, collectionId];
+		}
+		this.setState({
+			selectedCollections: newSelected
+		});
+	}
+
+	async getCollections() {
+		// Get all collections of the user (unlikely that a user will have more than 10000 collections
+		const req = await request.get(`/editor/${this.props.userId}/collections/collections?type=${this.props.entityType}&size=10000`);
```
MonkeyDo, line 58: It might be a good idea to wrap this in a try/catch block for safety, and display a generic user-friendly error message if there is a failure, something like "Sorry, we could not fetch your collections ({error})"

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+		const collections = req.body;
+		return collections;
```
MonkeyDo, line 60:
```suggestion
		return req.body;
```

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+				message: {
+					text: `Successfully added to selected collection${selectedCollections.length > 1 ? 's' : ''}`,
+					type: 'success'
+				}
+			});
+		}
+		else {
+			this.setState({
+				message: {
+					text: 'No Collection Selected',
```
MonkeyDo, line 86:
```suggestion
					text: 'No collection selected',
```

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,296 @@
...
+						<FontAwesomeIcon icon="times"/> Cancel
+					</Button>
+				</Modal.Footer>
+			</Modal>
+		);
+	}
+}
+
+
+AddToCollectionModal.displayName = 'DeleteCollectionModal';
```
MonkeyDo, line 287:
```suggestion
AddToCollectionModal.displayName = 'AddToCollectionModal';
```

src/client/components/pages/parts/add-to-collection-modal.js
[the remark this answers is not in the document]
MonkeyDo: Right, it should be `Edition Groups` instead with a space

src/client/components/pages/parts/add-to-collection-modal.js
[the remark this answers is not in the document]
MonkeyDo: Hm, I think we may have inconsistent placement and color across the website. I was looking at the relationship editor where the close button is on the left.

prabalsingh24:
> Maybe `Add to collection`?
> What do you think?

With `add to collection` the button is becoming bigger than others...

prabalsingh24:
> This looks nice currently, but a user is allowed to continue clicking on "add to new collection" and it will create as many duplicate collections.
> We can avoid that easily by either:
> 
>     1. closing the modal (and maybe showing a success alert in the entity footer (but what about adding from the search page?) or
> 
>     2. disabling the button after successful creation. I think that is more elegant and self-contained.


<br><br>
I like the first option better. User might want to add to different collection too.

Screenshot_2020-06-30 Flatbreads Flavors (Edition) – BookBrainz(1)

<br>
As for search result page, I am planning to show error message like this, we can show `success` message too

Screenshot_2020-06-30 Search Results – BookBrainz

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,307 @@
...
+								<FontAwesomeIcon icon="plus"/>
+								&nbsp;New collection
+							</Button>
+					}
+					{
+						this.state.showCollectionForm ?
+							<Button bsStyle="success" onClick={this.handleAddToNewCollection}>
+								<FontAwesomeIcon icon="plus"/> Add to new collection
+							</Button> :
+							<Button bsStyle="success" onClick={this.handleAddToCollection}>
```
MonkeyDo, line 283: I'd suggest making this `disabled` if `!this.state.collectionsAvailable.length`

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,307 @@
...
+
+		return (
+			<Modal
+				scrollable
+				show={this.props.show}
+				onHide={this.props.handleCloseModal}
+			>
+				<Modal.Header closeButton>
+					<Modal.Title>
+						Select Collection
```
MonkeyDo, line 252:
```suggestion
						{this.state. showCollectionForm ? 'Create' : 'Select'} Collection
```

src/client/components/pages/parts/add-to-collection-modal.js
```diff
@@ -0,0 +1,307 @@
...
+		if (selectedCollections.length) {
+			const promiseArray = [];
+			selectedCollections.forEach((collectionId) => {
+				const submissionURL = `/collection/${collectionId}/add`;
+				promiseArray.push(
+					request.post(submissionURL)
+						.send({bbids})
+				);
+			});
+			await Promise.all(promiseArray);
```
MonkeyDo, line 85: We probably want to wrap this in a try/catch block and display a generic error message if something goes south

src/server/routes/collection.js
```diff
@@ -85,4 +85,31 @@ router.get('/:collectionId', (req, res) => {
...
+		const {UserCollectionItem} = req.app.locals.orm;
+		for (const bbid of bbids) {
+			try {
+				await new UserCollectionItem({
+					bbid,
+					collectionId: collection.id
+				}).save(null, {method: 'insert'});
+			}
+			catch (err) {
+				if (err.constraint !== 'user_collection_item_pkey') {
```
MonkeyDo, line 102: I wish there was a better way to check if it's an "entry already exists" error, but nothing I tried is better than this. So be it.
A comment would be useful above this line to explain why we're doing this, as it's not clear unless you're very familiar with the db errors (which I'll admit I'm not).

However in this block you would `throw err` instead of logging it so that it can be dealt with by the parent catch block. Otherwise you're in a situation where the addition failed but you're sending a success response.

PR #468, opened 2020-06-29
CollectionView page

prabalsingh24:
### Problem
CollectionView Page

### DIFF
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -17,15 +17,18 @@
  */
 
 import * as bootstrap from 'react-bootstrap';
+import {ENTITY_TYPE_ICONS} from '../../helpers/entity';
 import EditionTable from './entities/edition-table';
+import EntityImage from './entities/image';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
 import React from 'react';
 import WorkTable from './entities/work-table';
+import {formatDate} from '../../helpers/utils';
 import request from 'superagent';
 
 
-const {Alert, Button, ButtonGroup, Row} = bootstrap;
+const {Alert, Button, Col, Row} = bootstrap;
 
 function getEntityTable(entityType) {
 	const tables = {
@@ -42,6 +45,45 @@ function getEntityKey(entityType) {
 	};
 	return keys[entityType];
 }
+
+function CollectionAttributes({collection}) {
+	return (
+		<div>
+			<Row>
+				<Col md={3}>
+					<dl>
+						<dt>Description</dt>
+						<dd>{collection.description}</dd>
+					</dl>
+				</Col>
+				<Col md={3}>
+					<dl>
+						<dt>Owner</dt>
+						<dd>{collection.owner.name}</dd>
+						<dt>Collaborators</dt>
+						<dd>{collection.collaborators.map((collaborator) => `${collaborator.text} `)}</dd>
+					</dl>
+				</Col>
+				<Col md={3}>
+					<dl>
+						<dt>Last Modified</dt>
+						<dd>{formatDate(new Date(collection.lastModified), true)}</dd>
+						<dt>Created At</dt>
+						<dd>{formatDate(new Date(collection.createdAt), true)}</dd>
+					</dl>
+				</Col>
+				<Col md={3}>
+					<dt>Privacy</dt>
+					<dd>{collection.public ? 'Public' : 'Private'}</dd>
+				</Col>
+			</Row>
+		</div>
+	);
+}
+CollectionAttributes.displayName = 'CollectionAttributes';
+CollectionAttributes.propTypes = {
+	collection: PropTypes.object.isRequired
+};
 
 class CollectionPage extends React.Component {
 	constructor(props) {
@@ -103,45 +145,49 @@ class CollectionPage extends React.Component {
 		return (
 			<div>
 				<Row className="entity-display-background">
-					<div>
+					<Col className="entity-display-image-box text-center" md={2}>
+						<EntityImage
+							backupIcon={ENTITY_TYPE_ICONS[this.props.collection.entityType]}
+						/>
+					</Col>
+					<Col md={10}>
 						<h1><strong>{this.props.collection.name}</strong></h1>
-						<h4>{this.props.collection.description}</h4>
-						<hr/>
-						{
-							this.props.isOwner ?
-								<ButtonGroup className="pull-right margin-bottom-1">
-									<Button
-										bsStyle="warning"
-										className="margin-right-d2"
-										href={`/collection/${this.props.collection.id}/edit`}
-										title="Edit Collection"
-									>
-										<FontAwesomeIcon icon="pencil-alt"/>&nbsp;Edit
-									</Button>
-									<Button
-										bsStyle="danger"
-										className="margin-right-1"
-										title="Delete this Collection"
-									>
-										<FontAwesomeIcon icon="trash-alt"/> Delete
-									</Button>
-								</ButtonGroup> : null
-						}
-					</div>
+						<CollectionAttributes collection={this.props.collection}/>
+					</Col>
 				</Row>
 				<EntityTable{...propsForTable}/>
 				{errorComponent}
-				{
-					this.props.showCheckboxes ?
-						<Button
-							bsStyle="danger"
-							className="pull-left margin-top-1"
-							onClick={this.handleRemoveEntities}
-						>
-							<FontAwesomeIcon icon="times"/>
-							&nbsp;Remove selected {this.entityKey}
-						</Button> : null
-				}
+				<div className="margin-top-1 text-center">
+					{
+						this.props.showCheckboxes ?
+							<Button
+								bsStyle="danger"
+								onClick={this.handleRemoveEntities}
+							>
+								<FontAwesomeIcon icon="times"/>
+								&nbsp;Remove selected {this.entityKey}
+							</Button> : null
+					}
+					{
+						this.props.isOwner ?
+							<Button
+								bsStyle="warning"
+								href={`/collection/${this.props.collection.id}/edit`}
+								title="Edit Collection"
+							>
+								<FontAwesomeIcon icon="pencil-alt"/>&nbsp;Edit collection
+							</Button> : null
+					}
+					{
+						this.props.isOwner ?
+							<Button
+								bsStyle="danger"
+								title="Delete Collection"
+							>
+								<FontAwesomeIcon icon="trash-alt"/>&nbsp;Delete collection
+							</Button> : null
+					}
+				</div>
 			</div>
 		);
 	}
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -49,32 +49,47 @@ function getEntityKey(entityType) {
 function CollectionAttributes({collection}) {
 	return (
 		<div>
+			{
+				collection.description.length ?
+					<Row>
+						<Col md={12}>
+							<dt>Description</dt>
+							<dd>{collection.description}</dd>
+						</Col>
+					</Row> : null
+			}
 			<Row>
 				<Col md={3}>
-					<dl>
-						<dt>Description</dt>
-						<dd>{collection.description}</dd>
-					</dl>
+					<dt>Owner</dt>
+					<dd><a href={`/editor/${collection.ownerId}`}>{collection.owner.name}</a></dd>
 				</Col>
-				<Col md={3}>
-					<dl>
-						<dt>Owner</dt>
-						<dd>{collection.owner.name}</dd>
-						<dt>Collaborators</dt>
-						<dd>{collection.collaborators.map((collaborator) => `${collaborator.text} `)}</dd>
-					</dl>
-				</Col>
-				<Col md={3}>
-					<dl>
-						<dt>Last Modified</dt>
-						<dd>{formatDate(new Date(collection.lastModified), true)}</dd>
-						<dt>Created At</dt>
-						<dd>{formatDate(new Date(collection.createdAt), true)}</dd>
-					</dl>
-				</Col>
+				{
+					collection.collaborators.length ?
+						<Col md={3}>
+							<dt>Collaborators</dt>
+							<dd>
+								{
+									collection.collaborators.map((collaborator) =>
+										<a href={`/editor/${collaborator.id}`} key={collaborator.id}>{collaborator.text}, </a>)
+								}
+							</dd>
+						</Col> : null
+				}
 				<Col md={3}>
 					<dt>Privacy</dt>
 					<dd>{collection.public ? 'Public' : 'Private'}</dd>
+				</Col>
+				<Col md={3}>
+					<dt>Collection type</dt>
+					<dd>{collection.entityType}</dd>
+				</Col>
+				<Col md={3}>
+					<dt>Created At</dt>
+					<dd>{formatDate(new Date(collection.createdAt), true)}</dd>
+				</Col>
+				<Col md={3}>
+					<dt>Last Modified</dt>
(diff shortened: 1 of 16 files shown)


prabalsingh24:
@MonkeyDo What do you think of thiss? I've created only `EditionCollection` page right now. If you think this is alright, I will create other `EntityTables` similar to this as well

Screenshot_2020-06-29 BookBrainz – The Open Book Database

prabalsingh24:
I made collectionView page for Work also. Reused the existing `Work-Table`

Screenshot_2020-06-29 BookBrainz – The Open Book Database(3)

src/client/components/pages/collection.js
```diff
@@ -0,0 +1,164 @@
...
+	handleRemoveEntities() {
+		if (this.state.selectedEntities.length) {
+			const bbids = this.state.selectedEntities;
+			const submissionUrl = `/collection/${this.props.collection.id}/remove`;
+			request.post(submissionUrl)
+				.send({bbids})
+				.then((res) => {
+					window.location.href = `/collection/${this.props.collection.id}`;
+				}, (error) => {
+					this.setState({error: 'Internal Error'});
```
MonkeyDo, line 84: Same as in the other PR, it would be great if this was a more user-friendly error message

src/client/components/pages/collection.js
```diff
@@ -0,0 +1,164 @@
...
+			this.setState({error: `No ${this.entityKey} selected`});
+		}
+	}
+
+	render() {
+		const errorComponent = this.state.error ?
+			<Alert bsStyle="danger">{this.state.error}</Alert> : null;
+		const EntityTable = getEntityTable(this.props.collection.entityType);
+		const propsForTable = {
+			[this.entityKey]: this.props.entities,
```
MonkeyDo, line 97: That's a smart way to solve the issue of each entity table requiring a specific named key. Nice!

src/client/components/pages/collection.js
```diff
@@ -0,0 +1,164 @@
...
+			showAdd: false,
+			showCheckboxes: this.props.showCheckboxes,
+			toggleRow: this.toggleRow
+		};
+		return (
+			<div>
+				<Row className="entity-display-background">
+					<div>
+						<h1><strong>{this.props.collection.name}</strong></h1>
+						<h4>{this.props.collection.description}</h4>
```
MonkeyDo, line 108: I think you can get the formatting closer to the existing entity pages, and also use a big entity icon. (I was originally thinking the icon of the entity type)

src/client/components/pages/collection.js
```diff
@@ -0,0 +1,164 @@
...
+						<hr/>
+						{
+							this.props.isOwner ?
+								<ButtonGroup className="pull-right margin-bottom-1">
```
MonkeyDo, line 112: I wonder if it would make more sense to show these buttons below the entity table, along with the batch remove button.

If that doesn't work out well, I'd suggest having the buttons left aligned rather than right-aligned

src/client/components/pages/collection.js
```diff
@@ -0,0 +1,164 @@
...
+				<EntityTable{...propsForTable}/>
+				{errorComponent}
+				{
+					this.props.showCheckboxes ?
+						<Button
+							bsStyle="danger"
+							className="pull-left margin-top-1"
+							onClick={this.handleRemoveEntities}
+						>
+							<FontAwesomeIcon icon="times"/>
```
MonkeyDo, line 141:
```suggestion
							<FontAwesomeIcon icon="times-circle"/>
```

I think here the times-circle icon will be ideal

src/client/components/pages/collection.js
```diff
@@ -0,0 +1,164 @@
...
+							&nbsp;Remove selected {this.entityKey}
+						</Button> : null
+				}
+			</div>
+		);
+	}
+}
+
+
+CollectionPage.displayName = 'CollectionsPage';
```
MonkeyDo, line 151:
```suggestion
CollectionPage.displayName = 'CollectionPage';
```

src/server/routes/collection.js
```diff
@@ -55,6 +79,42 @@ router.param(
...
+	}
+	const {orm} = req.app.locals;
+	const relations = getEntityRelations(collection.entityType);
+
+	const entitiesPromise = collection.items.map(item => orm.func.entity.getEntity(orm, collection.entityType, item.bbid, relations));
+	const entities = await Promise.all(entitiesPromise);
+	const isOwner = req.user && parseInt(collection.ownerId, 10) === parseInt(req.user.id, 10);
+	let showCheckboxes = false;
+	if (req.user && (req.user.id === collection.ownerId ||
+		collection.collaborators.filter(collaborator => collaborator.id === req.user.id).length)) {
```
MonkeyDo, line 95:
```suggestion
	let showCheckboxes = isOwner;
	if (collection.collaborators.filter(collaborator => collaborator.id === req.user?.id).length)) {
```

src/client/components/pages/entities/edition-table.js
```diff
@@ -145,11 +167,18 @@ function EditionTable({editions, entity, showAdd}) {
 EditionTable.displayName = 'EditionTable';
 EditionTable.propTypes = {
 	editions: PropTypes.array.isRequired,
-	entity: PropTypes.object.isRequired,
-	showAdd: PropTypes.bool
+	entity: PropTypes.object,
+	selectedEntities: PropTypes.array,
+	showAdd: PropTypes.bool,
+	showCheckboxes: PropTypes.bool,
+	toggleRow: PropTypes.func
```
MonkeyDo, line 174:
```suggestion
	onToggleRow: PropTypes.func
```

This should be changed wherever that function is passed to a child component, to clearly identify that it's a event  function

src/client/components/pages/entities/edition-table.js
```diff
@@ -31,16 +31,27 @@ const {
...
 
-function EditionTableRow({edition}) {
+function EditionTableRow({edition, showCheckboxes, selectedEntities, toggleRow}) {
 	const name = getEntityLabel(edition);
 	const disambiguation = getEntityDisambiguation(edition);
 	const releaseDate = getEditionReleaseDate(edition);
 	const isbn = getISBNOfEdition(edition);
 	const editionFormat = getEditionFormat(edition);
 
+	/* eslint-disable react/jsx-no-bind */
```
MonkeyDo, line 41: I wonder if you could do without this.
Would changing `onClick={() => toggleRow(edition.bbid)}` 
to: `onClick={toggleRow.bind(this, edition.bbid)}` do the trick?

src/client/stylesheets/style.less
```diff
@@ -478,3 +478,8 @@ hr.wide {
 .badge.new {
 	background-color: @green;
 }
+
+.checkboxes {
+	margin-right: 0.5em !important;;
+	vertical-align: -.25em;
```
MonkeyDo, line 484: We'll need to modify this according to the resolution in the other PR

src/server/routes/collection.js
```diff
@@ -55,6 +79,42 @@ router.param(
...
 
+router.get('/:collectionId', auth.isAuthenticatedForCollectionView, async (req, res) => {
+	const {collection} = res.locals;
+	if (collection.entityType !== 'Edition' && collection.entityType !== 'Work') {
+		return res.status(200).send(collection);
+	}
+	const {orm} = req.app.locals;
+	const relations = getEntityRelations(collection.entityType);
+
+	const entitiesPromise = collection.items.map(item => orm.func.entity.getEntity(orm, collection.entityType, item.bbid, relations));
```
MonkeyDo, line 90: In the future we might want to think about optimizing this to fetch all the entities in a single batch, but that's for another day.

src/server/routes/collection.js
```diff
@@ -80,9 +140,23 @@ router.get('/:collectionId/edit', auth.isAuthenticated, auth.isCollectionOwner,
 
 router.post('/:collectionId/edit/handler', auth.isAuthenticatedForHandler, auth.isCollectionOwner, collectionCreateOrEditHandler);
 
-router.get('/:collectionId', (req, res) => {
+router.post('/:collectionId/remove', auth.isAuthenticated, auth.isCollectionOwnerOrCollaborator, async (req, res) => {
+	const {bbids} = req.body;
```
MonkeyDo, line 144: We probably want to send a detailed error if there are no bbids (missing in body or empty array)

src/server/routes/collection.js
```diff
@@ -80,9 +140,23 @@ router.get('/:collectionId/edit', auth.isAuthenticated, auth.isCollectionOwner,
...
-	res.status(200).send(collection);
+	try {
+		const {UserCollectionItem} = req.app.locals.orm;
+		const promiseArray = bbids.map((bbid) =>
+			new UserCollectionItem()
+				.where('bbid', bbid)
+				.where('collection_id', collection.id)
+				.destroy()
+		);
+		await Promise.all(promiseArray);
```
MonkeyDo, line 154: I wonder if you could do this in one batch instead of multiple calls?
Perhaps using Bookshelf's collections ?

src/server/routes/collection.js
```diff
@@ -80,9 +140,23 @@ router.get('/:collectionId/edit', auth.isAuthenticated, auth.isCollectionOwner,
...
+		res.status(200).send();
+	}
+	catch (err) {
+		res.status(500).send();
```
MonkeyDo, line 158: Let's `log.debug` this.

src/client/components/pages/entities/edition-table.js
prabalsingh24: that fixes that error but shows a different eslint error. :(

src/server/routes/collection.js
prabalsingh24: I'll add this to that doc

prabalsingh24:
@MonkeyDo Have a look at this. Also which identifier should be shown for Publisher, Work, Author?
There is ISBN identifier row in EditionTable..

MonkeyDo:
> Also should there be an identifier in Publisher, Work, Author Table? if yes then which?
> There is ISBN identifier row in EditionTable..

ISBNs for Editions is a distinguishing feature, I don't think we want to show any particular identifier for the other entities.
However, for Publishers, let's show Date Founded and Date Disolved
Not much more to show for Edition Groups.

We'll also want pagination for the collection items, looking at [reosarevok's collection of 7189 artists…](https://musicbrainz.org/collection/b2dcf238-c640-4f05-921c-22d39f8a1fec)
If you want that can be done in a separate PR, in case it requires to refactor a bunch of other components, to limit the scope.

MonkeyDo:
Also I think disabling the "Remove selected work" button rather than hiding it would be better, I find changes in layout are not appreciated by users if they can be avoided.

prabalsingh24:
Screenshot_2020-07-04 BookBrainz – The Open Book Database(1)

src/server/helpers/collections.js
```diff
@@ -106,3 +106,32 @@ export async function getOrderedPublicCollections(from, size, entityType, orm) {
...
+	return result.rows.map(row => row.bbid);
+	// const {UserCollectionItem} = orm;
+	// const items = await new UserCollectionItem()
+	// 	.where('collection_id', collectionId)
+	// 	.orderBy('added_at')
+	// 	.fetchPage({
+	// 		debug: true
+	// 	});
+	// const itemsJSON = items.toJSON();
+	// return itemsJSON.map(item => item.bbid);
```
prabalsingh24, line 136: `fetchPage` is not working here. `debug: true` shows two sql queries for `fetchPage` (idk why..). The problem is with second query which counts the number of items.. I think because of composite primary key

prabalsingh24: `fetchAll` is working fine though

prabalsingh24:
@MonkeyDo Have a look at this

prabalsingh24:
I also noticed couple of bugs. One in `UserCollectionForm` and other in `Editor Collection Page`. I will solve them in different PR.

src/server/routes/collection.js
```diff
@@ -112,6 +253,11 @@ router.post('/:collectionId/add', auth.isAuthenticated, auth.isCollectionOwnerOr
 					bbid,
 					collectionId: collection.id
 				}).save(null, {method: 'insert'});
+				const collectionModel = await new UserCollection({id: collection.id}).fetch({
+					require: true
+				});
+				collectionModel.set('last_modified', new Date());
+				await collectionModel.save(null, {method: 'update'});
```
MonkeyDo, line 260: Here we can do that in one query. [From the docs](https://bookshelfjs.org/api.html#Model-instance-save):
```
// UPDATE authors SET "bio" = 'Short user bio' WHERE "id" = 1
new Author({id: 1, first_name: 'User'})
  .save({bio: 'Short user bio'}, {patch: true})
  .then((model) => {
    // ...
  })
```

src/server/helpers/collections.js
MonkeyDo: I'll have a look into this. Have you checked if it behaves differently for other paginated pages?

MonkeyDo:
Thinking a bout it again, perhaps there should be a `count` or `entitiesCount` method on the ORM model to count the UserCollectionItems?

src/server/helpers/collections.js
MonkeyDo: At first I hadn't understood that it wasn't working *at all*, as in throwing an error :D
I thought the not working part was that there were two queries instead of one.

I see the issue. It's undoubtedly related to Bookshelf not supporting composite primary keys ([#1664](https://github.com/bookshelf/bookshelf/issues/1664)).

You solved it the right way :+1:. You can remove the commented out fetchpage code I think, we'll just do it this way.

src/server/routes/collection.js
```diff
@@ -56,16 +107,72 @@ router.param(
...
+		const relations = getEntityRelations(collection.entityType);
+
+		// fetch 1 more bbid to check next enabled for pagination
+		const bbids = await getCollectionItemBBIDs(collection.id, from, size + 1, orm);
+		// get next enabled for pagination
+		const {newResultsArray, nextEnabled} = utils.getNextEnabledAndResultsArray(bbids, size);
+		// load entities from bbids
+		const entitiesPromise = newResultsArray.map(bbid => orm.func.entity.getEntity(orm, collection.entityType, bbid, relations));
+		const entities = await Promise.all(entitiesPromise);
+		const isOwner = req.user && parseInt(collection.ownerId, 10) === parseInt(req.user.id, 10);
```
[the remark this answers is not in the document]
prabalsingh24: I have made this change, but I don't think this was needed. We're checking `req.user` first and because of `&&` if the first query is false the second is not calculated.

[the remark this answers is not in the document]
MonkeyDo: True.

src/client/components/pages/entities/publisher-table.js
```diff
@@ -0,0 +1,130 @@
...
+		tableContent = (
+			<React.Fragment>
+				<Table striped>
+					<thead>
+						<tr>
+							<th className="col-sm-6">Name</th>
+							<th className="col-sm-1">Area</th>
+							<th className="col-sm-1">Type</th>
+							<th className="col-sm-2">Date founded</th>
+							<th className="col-sm-2">Date dissolved</th>
```
MonkeyDo, line 87: I just noticed these column classes.
It's making  some rows wrap around for me, which we probably don't want.
Was there a specific reason for the col-… classes?
If it's to ensure you have enough space for the name, you can add a simple `style={{width: '40%'}}` to the Name <th element.

prabalsingh24:
is this alright? or should this be done in css class in `style.less` ?

src/client/components/pages/entities/publisher-table.js
prabalsingh24: yes because here there are 5 columns, there was little less space for name and there was little extra space for date columns

PR #469, opened 2020-07-08
fix: disable edit EntityType field for non empty collection

prabalsingh24:
### Problem
Collection Type field was editable for non empty collections

### DIFF
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -18,6 +18,7 @@
 
 
 import * as bootstrap from 'react-bootstrap';
+import {trim, uniqBy} from 'lodash';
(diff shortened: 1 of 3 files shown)


prabalsingh24:
how did you debug this btw? I still don't know why it's working this way.

MonkeyDo:
> how did you debug this btw? I still don't know why it's working this way.

I used the React Developer Tools extension to go have a look at the properties of the component. That's usually my first step when a component misbehaves.
I realized `canEditType` was `true` even though it shouldn't be. I didn't dig any further since there was a handy solution that meant we could simply remove it.
The next step would be to put a breakpoint in the server code where `canEditType` was generated to see why it was getting assigned wrong.

PR #477, opened 2020-07-27
Test create/edit collection

prabalsingh24:
Tests for `POST` `create/edit handler`

### DIFF
--- a/src/server/helpers/collectionRouteUtils.js
+++ b/src/server/helpers/collectionRouteUtils.js
@@ -19,9 +19,23 @@
 import * as handler from '../helpers/handler';
 import * as search from './search';
 import {camelCase, differenceWith, isEqual, toLower, upperFirst} from 'lodash';
+import log from 'log';
 
+<<<<<<< HEAD
 
 export async function collectionCreateOrEditHandler(req, res, next) {
+=======
+/**
+ * A middleware handler for create or edit actions on collections.
+ * @param {object} req - request object
+ * @param {object} res - response object
+ * @returns {promise} res.send promise
+ * @description
+ * Creates a new collection or updates the existing collection
+ * If it's a new collection or it's is changed, ElasticSearch index is updated
+ */
+export async function collectionCreateOrEditHandler(req, res) {
+>>>>>>> chore: add jsoc doc for createOrEditHandler
 	try {
 		const {UserCollection, UserCollectionCollaborator} = req.app.locals.orm;
 		const isNew = !res.locals.collection;
@@ -98,4 +112,4 @@ export async function collectionCreateOrEditHandler(req, res, next) {
 	catch (err) {
 		return next(err);
 	}
-}
+};
--- a/src/server/helpers/collectionRouteUtils.js
+++ b/src/server/helpers/collectionRouteUtils.js
@@ -22,14 +22,14 @@ import {camelCase, differenceWith, isEqual, toLower, upperFirst} from 'lodash';
 
 
 /**
- * A middleware handler for create or edit actions on collections.
+ * A handler for create or edit actions on collections.
  * @param {object} req - request object
  * @param {object} res - response object
  * @param {object} next - next object
  * @returns {promise} res.send promise
  * @description
  * Creates a new collection or updates the existing collection
- * If it's a new collection or it's is changed, ElasticSearch index is updated
+ * If it's a new collection or it's name is changed, ElasticSearch index is also updated
  */
 export async function collectionCreateOrEditHandler(req, res, next) {
 	try {
@@ -44,7 +44,7 @@ export async function collectionCreateOrEditHandler(req, res, next) {
 			method = 'insert';
 		}
 		else {
-			if (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType) {
+			if (res.locals.collection.items.length && (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType)) {
 				throw new Error('Trying to change entityType of a non empty collection');
 			}
 			newCollection = await new UserCollection({id: req.params.collectionId}).fetch({
--- a/src/server/helpers/collectionRouteUtils.js
+++ b/src/server/helpers/collectionRouteUtils.js
@@ -44,7 +44,7 @@ export async function collectionCreateOrEditHandler(req, res, next) {
 			method = 'insert';
 		}
 		else {
-			if (res.locals.collection.items.length && (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType)) {
+			if (res.locals.collection?.items?.length && (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType)) {
 				throw new Error('Trying to change entityType of a non empty collection');
 			}
 			newCollection = await new UserCollection({id: req.params.collectionId}).fetch({
--- a/src/server/helpers/collectionRouteUtils.js
+++ b/src/server/helpers/collectionRouteUtils.js
@@ -19,23 +19,19 @@
 import * as handler from '../helpers/handler';
 import * as search from './search';
 import {camelCase, differenceWith, isEqual, toLower, upperFirst} from 'lodash';
-import log from 'log';
 
-<<<<<<< HEAD
 
-export async function collectionCreateOrEditHandler(req, res, next) {
-=======
 /**
  * A middleware handler for create or edit actions on collections.
  * @param {object} req - request object
  * @param {object} res - response object
+ * @param {object} next - next object
  * @returns {promise} res.send promise
  * @description
  * Creates a new collection or updates the existing collection
  * If it's a new collection or it's is changed, ElasticSearch index is updated
  */
-export async function collectionCreateOrEditHandler(req, res) {
->>>>>>> chore: add jsoc doc for createOrEditHandler
+export async function collectionCreateOrEditHandler(req, res, next) {
 	try {
 		const {UserCollection, UserCollectionCollaborator} = req.app.locals.orm;
 		const isNew = !res.locals.collection;
@@ -112,4 +108,4 @@ export async function collectionCreateOrEditHandler(req, res) {
 	catch (err) {
 		return next(err);
 	}
-};
+}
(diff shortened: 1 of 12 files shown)


src/server/helpers/collectionRouteUtils.js
```diff
@@ -34,7 +44,7 @@ export async function collectionCreateOrEditHandler(req, res, next) {
 			method = 'insert';
 		}
 		else {
-			if (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType) {
+			if (res.locals.collection.items.length && (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType)) {
```
MonkeyDo, line 47: I wonder if it would be a good idea to play it a bit safer:
```suggestion
			if (res.locals.collection?.items?.length && (upperFirst(camelCase(req.body.entityType)) !== res.locals.collection.entityType)) {
```

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+		const data = {
+			description: 'some description',
+			entityType: 'Author',
+			name: 'collectionName',
+			privacy: 'public'
+		};
+		const res = await agent.post('/collection/create/handler').send(data);
+		const collection = await new UserCollection({id: res.body.id}).fetch();
+		expect(collection.get('id')).to.equal(res.body.id);
+		expect(collection.get('ownerId')).to.equal(collectionOwner.get('id'));
```
MonkeyDo, line 48: Let's take the time to check the other attributes here (description, name, privacy, entity type).
We won't have to do it in other tests, but it doesn't cost us anything to be sure everything works as expected.

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+			entityType: 'Author',
+			name: 'collectionName',
+			privacy: 'public'
+		};
+		const res = await agent.post('/collection/create/handler').send(data);
+		const collection = await new UserCollection({id: res.body.id}).fetch();
+
+		const res2 = await agent.get(`/search/search?q=${data.name}&type=collection`);
+		// TODO
+		expect(1).to.equal(1);
```
MonkeyDo, line 64: Rather than adding an expectation that passes silently (thus easy to forget about), it's better to change `it('should…` to `xit('should…`.
That way the test will show as skipped.

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+		collectionJSON = collection.toJSON();
+	});
+	afterEach((done) => {
+		// Clear DB tables then close superagent server
+		truncateEntities().then(() => {
+			agent.close(done);
+		});
+	});
+
+	it('should update the collection and return 200 status code', async () => {
```
MonkeyDo, line 140: I'd like to see this test split into three separate tests:

- one where you test changing the name, description, privacy, and check that each field has been changed properly like you did
- one where you test adding collaborators
- one where you test removing collaborators

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+
+		const data = {
+			collaborators: [collaborator.toJSON()],
+			description: 'some description',
+			entityType: 'Edition',
+			name: 'collectionName',
+			privacy: 'public'
+		};
+		const res = await agent.post(`/collection/${collectionJSON.id}/edit/handler`).send(data);
+		expect(res.status).to.equal(500);
```
MonkeyDo, line 201: It would be ideal to check the error message to ensure it's this specific error that is being thrown, otherwise if something else breaks we could have a false negative.
You would `expect(res.body.message).to.equal('Trying to change entityType of a non empty collection')`

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+		await newAgent.get('/cb');
+		const newData = {
+			description: 'new description',
+			entityType: 'Edition',
+			name: 'new collection name',
+			privacy: 'private'
+		};
+
+		const res = await newAgent.post(`/collection/${collectionJSON.id}/edit/handler`).send(newData);
+		// TODO expect(res.status).to.equal(500);
```
MonkeyDo, line 218: Same comments as above, better to skip this test with `xit(…`, and when you do get to implementing the check, also check the error message.

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+			description: 'some description',
+			entityType: 'Edition',
+			name: 'collectionName',
+			privacy: 'public'
+		};
+		const res = await agent.post(`/collection/${collectionJSON.id}/edit/handler`).send(data);
+		expect(res.status).to.equal(500);
+	});
+
+	it('should throw error when unauthorized user tries to edit the collection', async () => {
```
MonkeyDo, line 204: I think it would make sense to also test that a collection's collaborators can't change the details fo the collection (name, description, privacy and entity type)

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+	it('should return status code 500 for incorrect entityType', async () => {
+		const data = {
+			collaborators: [collaborator.toJSON()],
+			description: 'some description',
+			entityType: 'incorrect',
+			name: 'collectionName',
+			privacy: 'public'
+		};
+		const res = await agent.post(`/collection/${collectionJSON.id}/edit/handler`).send(data);
+		expect(res.status).to.equal(500);
```
MonkeyDo, line 171: Let's check the error message here (see comment below)

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,221 @@
...
+	it('should return status code 500 for empty collection name', async () => {
+		const data = {
+			collaborators: [collaborator.toJSON()],
+			description: 'some description',
+			entityType: 'Author',
+			name: '',
+			privacy: 'public'
+		};
+		const res = await agent.post(`/collection/${collectionJSON.id}/edit/handler`).send(data);
+		expect(res.status).to.equal(500);
```
MonkeyDo, line 183: What error message are we expecting here?

src/server/helpers/middleware.js
```diff
@@ -211,3 +211,34 @@ export function makeCollectionLoader() {
...
+		return next(new error.BadRequestError('Invalid collection name: Empty string not allowed', req));
+	}
+	const entityTypes = _.keys(utils.getEntityModels(orm));
+	if (!entityTypes.includes(entityType)) {
+		return next(new error.BadRequestError(`Invalid entity type: ${entityType} does not exist`, req));
+	}
+	const collaboratorIds = collaborators.map(collaborator => collaborator.id);
+	for (let i = 0; i < collaboratorIds.length; i++) {
+		const collaboratorId = collaboratorIds[i];
+		if (!Number.isInteger(collaboratorId) || collaboratorId <= 0) {
```
MonkeyDo, line 230: I'm worried that this might misfire for strings, considering `Number.isInteger("5") => false`.
Number manipulation is always a PITA in javascript…

For this specific case where we want to ascertain it's a positive integer, the easiest way I know out of this connundrum is a simple regexp: `/^\d+$/.test("1234")` which will work for strings and numbers alike and allow nothing other than positive integers, with nothing before or after.

test/src/server/routes/collection.js
```diff
@@ -14,6 +13,388 @@ const {expect} = chai;
...
+		expect(collection.get('id')).to.equal(res.body.id);
+		expect(collection.get('ownerId')).to.equal(collectionOwner.get('id'));
+		expect(collection.get('name')).to.equal(data.name);
+		expect(collection.get('entityType')).to.equal(data.entityType);
+		expect(collection.get('description')).to.equal(data.description);
+		expect(collection.get('public')).to.equal(true);
+		expect(res.status).to.equal(200);
+	});
+
+	xit('should add the collection in the ES index', async () => {
```
MonkeyDo, line 58: Is the `xit` here expected?

test/src/server/routes/collection.js
```diff
@@ -14,6 +13,388 @@ const {expect} = chai;
...
+
+		const res = await agent.post(`/collection/${collectionJSON.id}/edit/handler`).send(newData);
+		const updatedCollection = await new UserCollection({id: collectionJSON.id}).fetch({withRelated: ['collaborators']});
+		const updatedCollectionJSON = updatedCollection.toJSON();
+
+		expect(res.status).to.equal(200);
+		expect(updatedCollectionJSON.collaborators.length).to.equal(0);
+	});
+
+	xit('should update the collection in the ES index', async () => {
```
MonkeyDo, line 250: Same as before, is `xit` expected?

prabalsingh24:
I guess you'll have to fix this. It's showing no live es connection

src/server/helpers/handler.js
```diff
@@ -23,12 +23,11 @@ import log from 'log';
...
 	return promise
-		.then((result) => {
-			res.send(result);
-
+		.then(async (result) => {
 			if (typeof processingCallback === 'function') {
-				return processingCallback(result);
+				await processingCallback(result);
 			}
+			res.send(result);
```
prabalsingh24, line 31: this was required. Otherwise it was doing `res.send(result)` before updating the ES index

prabalsingh24:
I modified `handler.sendPromiseResult` because it was doing `res.send` before updating updating ES which was making tests inconsistent.

MonkeyDo:
> I modified `handler.sendPromiseResult` because it was doing `res.send` before updating updating ES which was making tests inconsistent.

Unfortunately it links the success of the ElasticSearch callback with the success of the whole transaction.
The reason why it's implemented that way is so that if ES is down for some reason, transactions (create new entity, etc.) still go through, even if the `processingCallback` fails.

If you want to try it out yourself, you can run the current PR, but don't start ES. Try creating  a new entity, you'll see it takes a long time and then throws an error.
What we want to happen is for the main transaction to go through in all cases, and the optional callback not to stop tat from happening. I'm not sure you'll be able to implement that differently from what it currently is.

Is there anything that can be done on the test instead, maybe adding a timeout to give time for the ES callback to finish?
In an async test, I sometimes do `await new Promise(resolve => setTimeout(resolve, 500));` for an easy 500ms delay in the test.

prabalsingh24:
@MonkeyDo Yess i used `setTimeout` and the tests passed. :)

PR #483, opened 2020-08-05
tests for `collection/add` and `collection/remove`

prabalsingh24:
### Problem
tests for POST `collection/collectionId/add`


test/src/server/routes/collection.js
```diff
@@ -0,0 +1,429 @@
...
+		const data = {
+			bbids: [author.get('bbid')]
+		};
+		const res = await agent.post(`/collection/${collection.get('id')}/add`).send(data);
+		const item = await new UserCollectionItem({
+			collectionId: collection.get('id')
+		}).fetchAll({});
+		const itemJSON = item.toJSON();
+
+		expect(res.status).to.equal(403);
```
MonkeyDo, line 197: Worth checking the `res.statusMessage` now that we've discussed how to do it.

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,429 @@
...
```
MonkeyDo, line 221: Same as above, let's check the statusMessage

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,429 @@
...
+		expect(itemJSON.length).to.equal(0);
+	});
+});
+
+
+describe('POST /collection/:collectionID/add', () => {
```
MonkeyDo, line 227: This should read /remove instead of /add

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,429 @@
...
+		}).save(null, {method: 'insert'});
+		const data = {
+			bbids: [author.get('bbid')]
+		};
+		const res = await agent.post(`/collection/${collection.get('id')}/remove`).send(data);
+		const item = await new UserCollectionItem({
+			collectionId: collection.get('id')
+		}).fetchAll({});
+		const itemJSON = item.toJSON();
+		expect(res.status).to.equal(403);
```
MonkeyDo, line 398: Same here, let's check statusMessage

test/src/server/routes/collection.js
```diff
@@ -0,0 +1,429 @@
...
```
MonkeyDo, line 426: Same here, let's check statusMessage

MonkeyDo:
I approved too fast…sorry!
After some thinking here are a couple more edge cases we want to test for; what happens when I:

- add or remove an invalid BBID
- add or remove an entity that doesn't exist
- add an entity of the wrong type for that collection

These 5 tests are currently missing, and raise some questions about the routes' implementation (that's why we love tests! :p).

- From what I can tell, I get a generic 500 server error when I try to add a non-existent or invalid BBID, where I would probably want a more specific "BBID XXXXX does not exist"
- Same for removing, I would probably expect an "BBID XXXX is not in the collection" type of error
- For invalid BBIDs I'd expect an "Invalid BBID" error message (see `makeEntityLoader` for [the check](https://github.com/bookbrainz/bookbrainz-site/blob/894168717536255800bba4b28159b93800fc610d/src/server/helpers/middleware.js#L157) and [the error](https://github.com/bookbrainz/bookbrainz-site/blob/894168717536255800bba4b28159b93800fc610d/src/server/helpers/middleware.js#L174)
- I can currently successfully add an author to an edition collection by POSTing to the /add route (not from the website), we want to add an entity type check. Adding works fine, but then an error will be thrown if I tried to visit that collection.

prabalsingh24:
@MonkeyDo Have a look at this :)

test/src/server/routes/collection.js
```diff
@@ -160,3 +161,511 @@ describe('POST /collection/collectionID/delete', () => {
...
+		};
+		const response = await agent.post(`/collection/${collection.get('id')}/add`).send(data);
+		const item = await new UserCollectionItem().where('collection_id', collection.get('id')).fetchAll({});
+		const itemJSON = item.toJSON();
+
+		expect(response).to.have.status(400);
+		expect(response.res.statusMessage).to.equal('Trying to add an entity of type Author to a collection of type Edition');
+		expect(itemJSON.length).to.equal(0);
+	});
+});
```
MonkeyDo, line 423: Another good test I hadn't thought about, but came up while looking at the middleware: what happens if I send an empty array? Same for the /remove endpoint.

src/server/helpers/middleware.js
```diff
@@ -211,3 +211,51 @@ export function makeCollectionLoader() {
...
+	}
+	const entities = await new Entity().where('bbid', 'in', bbids).fetchAll();
+	const entitiesJSON = entities.toJSON();
+	for (let i = 0; i < bbids.length; i++) {
+		const bbid = bbids[i];
+		const entity = entitiesJSON.find(currEntity => currEntity.bbid === bbid);
+		if (!entity) {
+			return next(new error.BadRequestError('Trying to add an entity which is not present', req));
+		}
+		if (entity.type !== collectionType) {
```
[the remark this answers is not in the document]
prabalsingh24: This wasn't required, but I converted it to lowerCase before comparing, just to be on the safe side,

src/server/helpers/middleware.js
```diff
@@ -211,3 +211,51 @@ export function makeCollectionLoader() {
...
+	const {collection} = res.locals;
+	for (let i = 0; i < bbids.length; i++) {
+		const bbid = bbids[i];
+		if (!commonUtils.isValidBBID(bbid)) {
+			return next(new error.BadRequestError('Trying to remove an entity having invalid bbid', req));
+		}
+	}
+	for (let i = 0; i < bbids.length; i++) {
+		const bbid = bbids[i];
+		const isBbidInCollection = collection.items.find(item => item.bbid === bbid);
```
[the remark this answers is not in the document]
prabalsingh24: I don't think we can use `array.includes(bbid)` here, because array here is array of array of objects.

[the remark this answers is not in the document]
MonkeyDo: You're right, we'd have to `.map` it beforehand, but what's the point? :)

src/server/helpers/middleware.js
```diff
@@ -211,3 +211,57 @@ export function makeCollectionLoader() {
...
+			return next(new error.BadRequestError(`Invalid BBID ${bbid}`, req));
+		}
+	}
+	const entities = await new Entity().where('bbid', 'in', bbids).fetchAll({require: false});
+	const entitiesJSON = entities.toJSON();
+	for (let i = 0; i < bbids.length; i++) {
+		const bbid = bbids[i];
+		const entity = entitiesJSON.find(currEntity => currEntity.bbid === bbid);
+		if (!entity) {
+			return next(new error.BadRequestError(`${collectionType} ${bbid} does not exist`, req));
```
MonkeyDo, line 235: I hadn't thought about this before, but I guess we should send an `error.NotFoundError`  (404) here instead.

MonkeyDo:
:rocket: !

PR #484, opened 2020-08-05
Test: Delete Collection

prabalsingh24:
### Problem

Tests for POST `collection/:collectionID/delete/handler`


prabalsingh24:
I just realized I missed a test `it('should remove remove the collection from the ES index')`
I have to implement this feature as well :)

prabalsingh24:
In your words, "that's why we love tests!" haha

PR #487, opened 2020-08-10
Feat: Adding Entity from Collection Page

prabalsingh24:
### Problem

Adding Entity from Collection Page

Screenshot_2020-08-10 BookBrainz – The Open Book Database(2)

### DIFF
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -277,7 +277,7 @@ class CollectionPage extends React.Component {
 						this.props.isCollaborator || this.props.isOwner ?
 							<Button
 								bsSize="small"
-								bsStyle="primary"
+								bsStyle="success"
 								title={`Add ${this.props.collection.entityType}`}
 								onClick={this.handleShowAddEntityModal}
 							>
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -123,7 +123,11 @@ class CollectionPage extends React.Component {
 		super(props);
 		this.state = {
 			entities: this.props.entities,
-			error: null,
+			message: {
+				text: null,
+				type: null
+			},
+			refreshTable: false,
 			selectedEntities: [],
 			showAddEntityModal: false,
 			showDeleteModal: false
@@ -139,6 +143,8 @@ class CollectionPage extends React.Component {
 		this.handleCloseAddEntityModal = this.handleCloseAddEntityModal.bind(this);
 		this.handleAlertDismiss = this.handleAlertDismiss.bind(this);
 		this.searchResultsCallback = this.searchResultsCallback.bind(this);
+		this.closeAddEntityModalShowMessageAndRefreshTable = this.closeAddEntityModalShowMessageAndRefreshTable.bind(this);
+		this.refreshDone = this.refreshDone.bind(this);
 	}
 
 	searchResultsCallback(newResults) {
@@ -167,13 +173,31 @@ class CollectionPage extends React.Component {
 			request.post(submissionUrl)
 				.send({bbids})
 				.then((res) => {
-					window.location.href = `/collection/${this.props.collection.id}`;
+					this.setState({
+						message: {
+							text: `Removed ${bbids.length} ${_.kebabCase(this.props.collection.entityType)}
+							${bbids.length > 1 ? 's' : ''}`,
+							type: 'success'
+						},
+						refreshTable: true,
+						selectedEntities: []
+					});
 				}, (error) => {
-					this.setState({error: 'Something went wrong! Please try again later'});
+					this.setState({
+						message: {
+							text: 'Something went wrong! Please try again later',
+							type: 'danger'
+						}
+					});
 				});
 		}
 		else {
-			this.setState({error: `No ${_.kebabCase(this.props.collection.entityType)} selected`});
+			this.setState({
+				message: {
+					text: `No ${_.kebabCase(this.props.collection.entityType)} selected`,
+					type: 'danger'
+				}
+			});
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


prabalsingh24:
I am not able to solve these two bugs. Can you look at this please?
1.  there is a horizontal line below `select edition`. Idk why it's showing up
2. the error message is showing up like this. It's covering up the `selection field`. 
Screenshot_2020-08-10 BookBrainz – The Open Book Database(3)

src/client/components/pages/parts/add-entity-to-collection-modal.js
```diff
@@ -0,0 +1,217 @@
...
+		if (this.state.entities.length === 0) {
+			this.handleAddEntity();
+		}
+
+		let errorComponent = null;
+		if (this.state.error) {
+			errorComponent =
+				(
+					<div className="text-center margin-top-1">
+						<Alert bsStyle="danger" onDismiss={this.handleAlertDismiss}>{this.state.error}</Alert>;
```
MonkeyDo, line 120: The trailing semicolon ';' at the end appears as text in the page, right after the alert:

Capture d’écran 2020-08-11 à 13 00 17

src/client/components/pages/parts/add-entity-to-collection-modal.js
```diff
@@ -0,0 +1,217 @@
...
+										onClick={() => this.handleRemoveEntity(index)}
+									>
+										<FontAwesomeIcon icon="times"/>&nbsp;Remove
+									</Button>
+								);
+								return (
+									<div key={entity.id}>
+										<EntitySearchFieldOption
+											buttonAfter={buttonAfter}
+											instanceId="collaboratorSearchField"
```
MonkeyDo, line 152: This looks like leftover from a copy paste

src/client/components/pages/parts/add-entity-to-collection-modal.js
```diff
@@ -0,0 +1,217 @@
...
+				<Modal.Footer>
+					{errorComponent}
+					<ButtonGroup>
+						<Button
+							bsStyle="primary"
+							type="button"
+							onClick={this.handleAddEntity}
+						>
+							<FontAwesomeIcon icon="plus"/>
+							&nbsp;Add another {lowerCase(this.props.collectionType)} field
```
MonkeyDo, line 191:
```suggestion
							&nbsp;Add another {lowerCase(this.props.collectionType)}
```

src/client/components/pages/parts/add-entity-to-collection-modal.js
```diff
@@ -0,0 +1,217 @@
...
+						</Button>
+						<Button
+							bsStyle="success"
+							disabled={!cleanedEntities.length}
+							onClick={this.handleSubmit}
+						>
+							{genEntityIconHTMLElement('collection')}
+							&nbsp;Add <Badge>{cleanedEntities.length}</Badge>&nbsp;
+							{lowerCase(this.props.collectionType)}{cleanedEntities.length > 1 ? 's' : ''} in the collection
```
MonkeyDo, line 200:
```suggestion
							{lowerCase(this.props.collectionType)}{cleanedEntities.length > 1 ? 's' : ''} to the collection
```

src/client/components/pages/parts/add-entity-to-collection-modal.js
```diff
@@ -0,0 +1,217 @@
...
+
+		return (
+			<Modal
+				bsSize="large"
+				show={this.props.show}
+				onHide={this.props.onCloseModal}
+			>
+				<Modal.Header closeButton>
+					<Modal.Title>
+						Add {lowerCase(this.props.collectionType)}s in the collection
```
MonkeyDo, line 176:
```suggestion
						Add {lowerCase(this.props.collectionType)}s to the collection
```

src/server/helpers/collections.js
```diff
@@ -121,7 +121,7 @@ export async function getCollectionItemBBIDs(collectionId, from, size, orm) {
 						SELECT bookbrainz.user_collection_item.bbid
 						FROM bookbrainz.user_collection_item
 						WHERE collection_id='${collectionId}'
-						ORDER BY user_collection_item.added_at ASC
+						ORDER BY user_collection_item.added_at DESC
```
MonkeyDo, line 124: This threw me off at first.
Did you change this so that it's clearer that an element was added to the collection, since it will appear at the top?

I'm not sure how we should display this, but ascending order made more sense to me.
Eventually, we'll want to give the user the freedom to sort however they want with some fancy table library, but for now…

prabalsingh24:
What do you think of `refreshDone` function? I used it in `remove entities` feature as well. Earlier I was refreshing the whole page.

src/client/components/pages/parts/pager.js
```diff
@@ -46,6 +46,10 @@ class PagerElement extends React.Component {
 			// eslint-disable-next-line react/no-did-update-set-state
 			this.setState({from: 0, query: this.props.query}, this.triggerSearch);
 		}
+		if (this.props.refreshTable) {
```
MonkeyDo, line 49: While it looks to me like this pattern would work, I don't think it's the most straightforward solution, and might generate extra rendering in the process.

To be honest there's no ideal way to communicate punctual events from parent to child, but what I usually end up doing is creating a reference to the child component and calling `myChildRef.doSomething()` from the parent component.
That way, the logic of deciding what should happen and when stays with the parent where it belongs, while the details of what's supposed to happen stay with the child.

Here's an example where we use that pattern:
https://github.com/bookbrainz/bookbrainz-site/blob/7ef08d9d5ca84f7e8befa6230117c5232d606c70/src/client/components/input/entity-search.js#L80
https://github.com/bookbrainz/bookbrainz-site/blob/7ef08d9d5ca84f7e8befa6230117c5232d606c70/src/client/components/input/entity-search.js#L38

MonkeyDo: In this case, in the parent (CollectionPage) it would look something like this:
```
closeAddEntityModalShowMessageAndRefreshTable(message) {
		this.setState({
			message,
			showAddEntityModal: false
		}, this.pagerElementRef.triggerSearch();
	}
```

prabalsingh24:
I forgot about `ref`. That's why i was using `refreshTable` `refreshDone` :P

src/client/components/pages/collection.js
```diff
@@ -163,45 +171,83 @@ class CollectionPage extends React.Component {
 			request.post(submissionUrl)
 				.send({bbids})
 				.then((res) => {
-					window.location.href = `/collection/${this.props.collection.id}`;
+					this.setState({
+						message: {
+							text: `Removed ${bbids.length} ${_.kebabCase(this.props.collection.entityType)}
+							${bbids.length > 1 ? 's' : ''}`,
```
MonkeyDo, line 177: SomehowBecause this is on a new line, it ends up with a space between entity type and 's':

Capture d’écran 2020-08-13 à 13 00 13

This is the only place it does that though, the other messages are fine.

MonkeyDo:
Great stuff!
This is going to make it much easier for users to create collections!

PR #489, opened 2020-08-11
fix: throw error when trying to add collection, area, editor to a collection

prabalsingh24:
### Problem
If you select `Editor` or `Area` or `Collection` in the search results, you can press `Add to collection` button and it doesn't throw any error.

### Solution

Screenshot_2020-08-11 Search Results – BookBrainz


prabalsingh24:
Test for this is added in the PR [477](https://github.com/bookbrainz/bookbrainz-site/pull/477/files)

PR #490, opened 2020-08-13
Delete collection from Elasticsearch index

prabalsingh24:
### Problem
After deleting the collection from Database, it has to be deleted from ES index as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ cache:
 
 services:
   - postgresql
+  - elasticsearch
 
 addons:
   postgresql: "12"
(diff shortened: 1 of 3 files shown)


prabalsingh24:
I've seen that sometimes `UserCollection({id: 1}).fetchAll()` gives incorrect result. whereas with `fetch` it works fine.  <br>
When i have to use `fetch` , I use `UserCollection({id: 1}).fetch()` whereas when i have to do `fetchAll` I use `UserCollection().where('id', 1).fetchAll`

MonkeyDo:
Huh, interesting. I wonder why that is.
Thanks for the tip!

prabalsingh24:
I could be wrong too. But i experienced it at many places.

prabalsingh24:
Next time i see a similar issue, I'll check it properly. Will use debug true and see what happens

PR #495, opened 2020-08-26
Create Collection Button and Added At Timestamp

prabalsingh24:
### Problem

Create collection from /collections page and editor collection tab 
Show added_at timestamp in collection table


MonkeyDo (changes requested):
I would rename `isCollectionTable` to `showAddedAtColumn` since we're only using it for that specific purpose.

Other than that everything else looks good to me, nicely done!

MonkeyDo (approved):
Good to go ! :rocket:

PR #496, opened 2020-08-27
fix(css): Adjust button margins

MonkeyDo:
With the User Collections we're doing new things with buttons that we haven't done before, which is unearthing a couple of small CSS issues.
For some reason the Lobes theme has slightly overlapping buttons, which means an opacity change (disabled) looks really bad.
We're also using badges in buttons which requires a small padding adjustment.

### Problem
Buttons are overlapping, which looks terrible with opacity (disabled buttons).
Also buttons with badges are a tiny bit higher that without.
Capture d’écran 2020-08-27 à 14 28 38

### Solution
Adjust button css here and there:
Capture d’écran 2020-08-27 à 14 37 14

### Areas of Impact
Just the LESS file.

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -73,6 +73,19 @@ body {
     }
 }
 
+// We use bootstrap badges in buttons, which makes them a smidge taller than buttons with just text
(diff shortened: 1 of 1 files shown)


prabalsingh24 (approved):
LGTM :)

PR #505, opened 2020-09-01
Merge UserCollection into Master

prabalsingh24:
### Problem
Merge UserCollection into Master


package.json
```diff
@@ -43,8 +43,10 @@
     "@fortawesome/free-brands-svg-icons": "^5.14.0",
     "@fortawesome/free-solid-svg-icons": "^5.14.0",
     "@fortawesome/react-fontawesome": "^0.1.11",
+    "bluebird": "^3.5.1",
```
MonkeyDo, line 46: The need for the bluebird library has been removed in a previous PR, and we shouldn't need the package anymore.

src/server/helpers/middleware.js
```diff
@@ -23,6 +23,8 @@ import * as commonUtils from '../../common/helpers/utils';
 import * as error from '../../common/helpers/error';
 import * as utils from '../helpers/utils';
 import type {$Request, $Response, NextFunction} from 'express';
+import Promise from 'bluebird';
```
MonkeyDo, line 26: Same comment as on package.json, bluebird has been removed and this line shouldn't be needed.

src/client/helpers/setupIconLibrary.js
```diff
@@ -38,6 +38,9 @@ config.autoAddCss = true;
 // Add Icons to FontAwesome library
 library.add(
 	fab, faAngleDoubleLeft, faAngleDoubleUp, faBook, faCalendarAlt, faChartLine, faCheck, faCircle,
+	faCircleNotch, faCodeBranch, faComment, faCommentDots, faComments, faEnvelope, faExclamationTriangle,
+	faExternalLinkAlt, faGlobe, faGripVertical, faHistory, faInfo, faListUl, faPenNib, faPencilAlt, faPencilRuler,
+	faPlus, faQuestionCircle, faRemoveFormat, faSave, faSearch, faSignInAlt,
```
MonkeyDo, line 43: It looks like these lines were duplicated during the merge from master.
You'll need to make sure the result does have the 3 new icons in there (faComment, faGripVertical and faSave)

MonkeyDo (changes requested):
I see a couple of merge issues, but nothing dramatic.
After that, it's ready to merge ! 🎉 🚀

prabalsingh24:
Coverage decreased (-1.7%), that's sad. I was hoping it to be increased. But I think because we don't have tests for front end stuff, and alot of front end code was added, it decreased.

MonkeyDo (approved):
Let's go !

PR #507, opened 2020-09-01
test: Prevent conflict in test fixtures ids

MonkeyDo:
### Problem
Tests were failing erratically because of two reasons:
1. We were assigning an id to pretty much all the test fixtures we created using the ORM, when it's not necessary
2. The random number generator we were using to generate the ids was creating conflicting ids

### Solution
Let the database create ids, which required a few changes in the test helpers.

### Areas of Impact
Tests only.


MonkeyDo:
@prabalsingh24 Requesting your eyes to double-check in case you spot something stupid :)

MonkeyDo:
I'll merge this now, but you can still review and comment if you spot anything !

PR #509, opened 2020-09-08
fix(test): add 500ms timeout after deleting a collection

prabalsingh24:
### Problem
One test was inconsistent and showing false negative. Adding timeout after deleting a collection will make sure that ES index is updated

### Solution
wait 500ms after deleting a collection to make sure ES index is updated

### Areas of Impact
test/src/server/routes/collection.js


MonkeyDo (approved):
👌 Thank you !

PR #511, opened 2020-09-08
Collection: Change default privacy to public

prabalsingh24:
### Problem
changing default privacy of a collection to public

### DIFF
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -161,8 +161,14 @@ class UserCollectionForm extends React.Component {
 		}));
 		const initialName = this.state.collection.name;
 		const initialDescription = this.state.collection.description;
-		const initialPrivacy = this.state.collection.public ? 'Public' : 'Private';
 		const initialType = this.state.collection.entityType;
+		let initialPrivacy;
+		if (this.props.collection.name) {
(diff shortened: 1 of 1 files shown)


src/client/components/forms/userCollection.js
```diff
@@ -161,8 +161,14 @@ class UserCollectionForm extends React.Component {
 		}));
 		const initialName = this.state.collection.name;
 		const initialDescription = this.state.collection.description;
-		const initialPrivacy = this.state.collection.public ? 'Public' : 'Private';
 		const initialType = this.state.collection.entityType;
+		let initialPrivacy;
+		if (this.props.collection.name) {
+			initialPrivacy = this.state.collection.public ? 'Public' : 'Private';
+		}
```
MonkeyDo, line 168: What does the new condition do?
Is it to check if the collection already exists (vs a new collection)?

prabalsingh24: Yes. If we're editing the collection, then it should show the proper privacy settings and if we're creating a new collection, it should show 'Public'. This is what we want to achieve right?

MonkeyDo (approved):

MonkeyDo:
Thanks for this !

PR #512, opened 2020-09-08
fix pagination url for collection table

prabalsingh24:
### Problem
Pagination is collection table was not working properly

### Solution
fixed the pagination url. Removed an extra `?`


prabalsingh24:
This incorrect pagination url was causing some error in the pagination. The url was becoming `collection?q=?size=20`.
Removing that extra `?` will fix it

MonkeyDo (approved):
Ah, yes, good catch, I made those changes to the pagination mechanism on master, and forgot it would break the collections page!
Sorry and thanks for fixing it quick!

PR #513, opened 2020-09-09
Collections public dumps

MonkeyDo:
### Problem
With the introduction of user collections, private collections are currently going to be exported in the dumps.

### Solution
This PR aims to create a database dump without collections, and another dump with *only* the public collections.
For that purpose, we create temporary tables (ie `user_collection` -> `public_user_collection`) with the appropriate select statements to ignore private collections its items and collaborators, and dump those three tables to a file before removing them.

We also want to rename the table names in the collections dump file once it has been created (for example to rename `bookbrainz.public_user_collection` -> `bookbrainz.user_collection`

### DIFF
--- a/scripts/create-public-collection-dumps.sql
+++ b/scripts/create-public-collection-dumps.sql
@@ -1,0 +1,8 @@
+CREATE table if not exists public_user_collection AS
+select * from bookbrainz.user_collection uc where uc.public is true;
+
+CREATE table if not exists public_user_collection_item AS
+select uci.* from user_collection_item uci right join public_user_collection on uci.collection_id = public_user_collection.id;
+
+CREATE table if not exists public_user_collection_collaborator AS
+select ucc.* from user_collection_collaborator ucc inner join public_user_collection on ucc.collection_id = public_user_collection.id;
--- a/DEPENDENCIES_MANUAL_INSTALL.md
+++ b/DEPENDENCIES_MANUAL_INSTALL.md
@@ -44,22 +44,26 @@ No setup is required for Redis or Elasticsearch. However, it is necessary to
 perform some initialization for PostgreSQL and import the latest BookBrainz
 database dump.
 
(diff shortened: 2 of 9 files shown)


MonkeyDo:
Pinging @paramsingh instead of a review request which doesn't work…

scripts/create-public-collection-dumps.sql
```diff
@@ -0,0 +1,8 @@
+CREATE table if not exists public_user_collection AS
```
paramsingh, line 1: Should this be inside a transaction?

paramsingh (commented):
The changes look reasonable to me, however, I'm wondering if it's a better idea to keep the collections in the same dump. different data dump files make it difficult for users to get into the project, or consume the data, is there a reason why we did the two dumps?

MonkeyDo:
> Different data dump files make it difficult for users to get into the project, or consume the data, is there a reason why we did the two dumps?

I didn't find any other way to not dump the private collections, other than create a dump without the collections and a dump with the public collections only (consisting of selected rows of three tables).

Do you think I could just concatenate the two sql dump files? I must admit I haven't tried that.

MonkeyDo:
I did end up concatenating the two dump files, and it works like a charm, thanks @paramsingh !
Back to a single dump file.

I also realized my duplicated tables didn't have their foreign keys, and had to add them to the collection dump sql script.

scripts/create-public-collection-dumps.sql
```diff
@@ -0,0 +1,36 @@
+BEGIN TRANSACTION;
+
+-- duplicate user_collection table with public collections only
+
+CREATE table if not exists public_user_collection (LIKE bookbrainz.user_collection INCLUDING ALL);
```
mayhem, line 5: I would prefer to have these tables named starting with tmp_  to make it clear that they are dump tables. The clean-public-collection-dump-tables.sql script is a bit terrifying to read without the knowledge that these are not temp tables.

mayhem (changes requested):
Please rename tables to make it clear they are temp. Next, what happens with private collections? Will there be a follow up PR to handle those?

PR #517, opened 2020-09-29
fix(BB-452): Index auto-created Edition Groups

MonkeyDo:
Edition Groups created automatically by the ORM ([when no EG is set on a  new Edition](https://github.com/bookbrainz/bookbrainz-data-js/blob/master/src/util.js#L249-L262)) were not being indexed for search.


MonkeyDo:
Followed up in b3b6215992e2108dea7ebf9e214cbbc479d7c9d5

PR #521, opened 2020-10-07
refactor: No need of body parser

MehtaManan07:
### Problem

There is no need to use body-parser

### Solution

Small refactor and changed the code where body-parser was used in src/api/app.js

### Areas of Impact

I have changed only one file, src/api/app.js


MonkeyDo:
Hi @MehtaManan07 !
Thanks for this PR! I wasn't aware that body-parser is now part of core Express.
There are a couple of things I'd like you to do:
1) The tests are currently [failing because of style issues](https://travis-ci.org/github/bookbrainz/bookbrainz-site/jobs/733675324#L698-L701), which you should be able to solve by running the command `npm run lint -- --fix` (the double '--' is on purpose).
2) There is another file using body-parser, which would benefit from the same treatment: https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/app.js#L63-L66
3) Once it has been removed in the second file, we don't need the `body-parser` package anymore, so let's remove it !
You'll need to run `npm rm --save body-parser` and commit the resulting changes in the `package.json` and `package-lock.json` files


For posterity and anyone wondering, the functionality of body-parser has been re-added to the ExpressJS core from v4.16.0 onwards, and is thus not necessary anymore as a separate module (at time of writing we use ExpressJS v4.17.1).

MehtaManan07:
> Hi @MehtaManan07 !
> Thanks for this PR! I wasn't aware that body-parser is now part of core Express.
> There are a couple of things I'd like you to do:
> 
> 1. The tests are currently [failing because of style issues](https://travis-ci.org/github/bookbrainz/bookbrainz-site/jobs/733675324#L698-L701), which you should be able to solve by running the command `npm run lint -- --fix` (the double '--' is on purpose).
> 2. There is another file using body-parser, which would benefit from the same treatment: https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/app.js#L63-L66
> 3. Once it has been removed in the second file, we don't need the `body-parser` package anymore, so let's remove it !
>    You'll need to run `npm rm --save body-parser` and commit the resulting changes in the `package.json` and `package-lock.json` files
> 
> For posterity and anyone wondering, the functionality of body-parser has been re-added to the ExpressJS core from v4.16.0 onwards, and is thus not necessary anymore as a separate module (at time of writing we use ExpressJS v4.17.1).

Thank you so much for considering it. I have made the changes as specified. Should I make a new PR or how should I proceed further?

MonkeyDo:
> Thank you so much for considering it. I have made the changes as specified. Should I make a new PR or how should I proceed further?

No need to open another PR, this looks just fine :)

I will review and test it, and once that is done I will merge the PR.
Thanks a lot !

MonkeyDo (changes requested):
Thanks, that looks a lot better!
I think we're only missing one thing: you didn't commit the changes to the `package.json` file (which should show only the removal of the `body-parser` package).
You did commit the changes to `package-lock.json`, which is good.

MehtaManan07:
> Thanks, that looks a lot better!
> I think we're only missing one thing: you didn't commit the changes to the `package.json` file (which should show only the removal of the `body-parser` package).
> You did commit the changes to `package-lock.json`, which is good.

Actually, I am using windows, and for the installation of some packages, the path was looking incorrect for the system. That is why I had to make a small change in that file which is windows specific. 
That is why I thought it would be better not to commit that file.
So should I commit it or what changes should I do?

MonkeyDo:
We definitely need it to be commited, but maybe without that windows-specific change.
Out of curiosity, what was the change? Perhaps it con't be any harm and will help other windows users!

Otherwise you can  commit only the part needed with `git add --patch package.json`, which will start an interactive staging seccion that allows you to choose which parts of the file to stage, before commiting.

MonkeyDo:
Hi @MehtaManan07 !
Can I do anything to help you move forward, and to finish and merge this PR together?

MehtaManan07:
> Hi @MehtaManan07 !
> Can I do anything to help you move forward, and to finish and merge this PR together?

Actually, I was a bit confused. But I will commit the changes and push them as soon as possible. Thanks by the way @MonkeyDo

MehtaManan07:
I have done the required changes and pushed the code to the local branch. @MonkeyDo please let me know if I am doing anything wrong anywhere.

MonkeyDo (approved):
Thanks, the PR looks good!
One less dependency ! :rocket: :heart:

PR #526, opened 2020-10-21
Changed the styling of buttons to enhance the semanticity of website

frappelatte28:
### Problem
https://tickets.metabrainz.org/browse/BB-421 

### Solution

Enhanced the UI of entity editor by changing the style of buttons  , bookbrainz uses react-bootstrap version ( 0.33.1) for most for of UI components , i changed the ``bsStytle`` props of   ``<button>`` component from "link"  to "success" to make the button look green .
Which not only enhances the buttons semanticity to provide the meaning that it is adding  something( here an entity)  but  also it matches with styling of other button components there.

### Areas of Impact
src/client/entity-editor/button-bar/alias-button.js
src/client/entity-editor/button-bar/identifier-button.js
src/client/entity-editor/edition-section/edition-section.js

### DIFF
--- a/src/client/entity-editor/button-bar/alias-button.js
+++ b/src/client/entity-editor/button-bar/alias-button.js
@@ -50,7 +50,7 @@ function AliasButton({
 		<FontAwesomeIcon className="margin-right-0-5 text-danger" icon="times"/>;
 
 	return (
-		<Button bsStyle="link" {...props}>
+		<Button bsStyle="success" {...props}>
 			{iconElement}
 			{text}
 		</Button>
--- a/src/client/entity-editor/button-bar/identifier-button.js
+++ b/src/client/entity-editor/button-bar/identifier-button.js
@@ -53,7 +53,7 @@ function IdentifierButton({
 		<FontAwesomeIcon className="margin-right-0-5 text-danger" icon="times"/>;
 
 	return (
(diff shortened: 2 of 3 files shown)


frappelatte28:
@MonkeyDo please kindly review this PR!

MonkeyDo:
Hi @frappelatte28!
Thanks for opening this PR! I do appreciate some help with improved the UI  :)

As for those buttons… well, the end goal would be to redesign the edit pages to get rid of the modals, and instead have those sections on the page directly.
For an example, have a look at the "external links" section on the MusicBrainz edit pages: https://musicbrainz.org/artist/create
Try pasting the following link in there to see what happens: https://www.wikidata.org/wiki/Q5383

frappelatte28:
@MonkeyDo 
Okay! I think the same, we should redesign the edit page as its confusing for users. But as u said it will take a lot time and  i have just started my journey here, I think I should be focusing more on smaller issues for now but in near time as i gain my momentum, i will surely look into it :)

MonkeyDo:
Here's a list of more accessible tickets that we tag with a "good-first-bug" label, if you want to have a look and see if anything strikes your fancy: https://tickets.metabrainz.org/issues/?filter=11910

Of course, simply using the website, discovering issues and fixing them like you did in #527 is always great :)

PR #527, opened 2020-10-27
Fixed responsiveness of buttons in create-collection page

frappelatte28:
### Problem
https://tickets.metabrainz.org/browse/BB-558
WhatsApp Image 2020-10-27 at 2 27 13 PM

### Solution
Screenshot from 2020-10-27 18-12-44

Applied top margin for mobile screens using media query.

### Areas of Impact

src/client/components/forms/userCollection.js

### DIFF
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -262,7 +262,8 @@ class UserCollectionForm extends React.Component {
 							<div className={errorAlertClass}>
 								<Alert bsStyle="danger">Error: {errorText}</Alert>
 							</div>
-							<div className="text-center buttons-container">
+							<div className="row">
+								<div className = "col-sm-6">
 								<Button
 									bsStyle="primary"
 									type="button"
@@ -271,12 +272,15 @@ class UserCollectionForm extends React.Component {
 									<FontAwesomeIcon icon="plus"/>
 									&nbsp;Add another collaborator
 								</Button>
+								</div>
+								<div className = "col-sm-6">
 								<Button
 									bsStyle="success"
 									type="submit"
 								>
 									<FontAwesomeIcon icon="save"/>&nbsp;{submitLabel}
 								</Button>
+								</div>
 								{
 									this.props.collection.id ?
 										<Button
@@ -324,3 +328,4 @@ UserCollectionForm.defaultProps = {
 };
 
 export default UserCollectionForm;
+
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -262,7 +262,7 @@ class UserCollectionForm extends React.Component {
 							<div className={errorAlertClass}>
 								<Alert bsStyle="danger">Error: {errorText}</Alert>
 							</div>
-							<div className="text-center">
+							<div className="text-center buttons-container">
 								<Button
 									bsStyle="primary"
 									type="button"
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -262,9 +262,10 @@ class UserCollectionForm extends React.Component {
 							<div className={errorAlertClass}>
 								<Alert bsStyle="danger">Error: {errorText}</Alert>
 							</div>
-							<div className="row">
+							<div className="row margin-bottom-2">
 								<div className="col-sm-6 margin-top-d5">
 									<Button
+										block
 										bsStyle="primary"
 										type="button"
 										onClick={this.handleAddCollaborator}
@@ -275,6 +276,7 @@ class UserCollectionForm extends React.Component {
 								</div>
 								<div className="col-sm-6 margin-top-d5">
 									<Button
+										block
 										bsStyle="success"
 										type="submit"
 									>
@@ -283,13 +285,16 @@ class UserCollectionForm extends React.Component {
 								</div>
 								{
 									this.props.collection.id ?
-										<Button
-											bsStyle="danger"
-											type="button"
-											onClick={this.handleShowModal}
-										>
-											<FontAwesomeIcon icon="trash-alt"/>&nbsp;Delete collection
-										</Button> : null
+										<div className="col-sm-6 margin-top-d5">
+											<Button
+												block
+												bsStyle="danger"
+												type="button"
+												onClick={this.handleShowModal}
+											>
+												<FontAwesomeIcon icon="trash-alt"/>&nbsp;Delete collection
+											</Button>
+										</div> : null
 								}
 							</div>
 						</form>
@@ -328,4 +333,3 @@ UserCollectionForm.defaultProps = {
 };
 
 export default UserCollectionForm;
-
--- a/src/client/components/forms/userCollection.js
+++ b/src/client/components/forms/userCollection.js
@@ -263,23 +263,23 @@ class UserCollectionForm extends React.Component {
 								<Alert bsStyle="danger">Error: {errorText}</Alert>
 							</div>
 							<div className="row">
-								<div className = "col-sm-6">
-								<Button
-									bsStyle="primary"
-									type="button"
-									onClick={this.handleAddCollaborator}
-								>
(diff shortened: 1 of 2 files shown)


frappelatte28:
@MonkeyDo Please  kindly review this PR :)

MonkeyDo (commented):
Thanks, good find!

I have a suggestion for a different approach:
First of all, let's fix the issue with the overlapping buttons in a way that will fix it across the website. I've identified the issue is coming from the bootstrap theme, but can (and has) been mitigated; see this bit of CSS used previously to sort the issue of horizontal overlap:
https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/stylesheets/style.less#L83-L87
I propose to change that simply to `margin: 0;` to fix the overlapping issue vertically as well.

With that out of the way, I think there is a simple way to display our buttons properly using existing css.
We can use bootstrap's responsive grid system by modifying the structure just a bit:
```html
<div class="row">
  <div class="col-sm-6">
    <button type="button" class="btn btn-block btn-primary">Add another collaborator</button>
  </div>
  <div class="col-sm-6">
    <button type="submit" class="btn btn-block btn-success">Create collection</button>
  </div>
</div>
```

1. We change the `text-center` class of the parent element to `row`
2. We create two `<div class="col-sm-6">` elements around the existing buttons
3. Profit.

I think it looks quite good like that, with full width buttons in mobile sizes, and better looking as well in desktop sizes I find:

Capture d’écran 2020-10-28 à 14 23 52
Capture d’écran 2020-10-28 à 14 23 35



How does that sound?

frappelatte28:
@MonkeyDo 
i think we can achieve the same with little less mess ,
we can use predefined " no-gutters " class on .row to remove negative margin of row and extra gutters of columns .

https://getbootstrap.com/docs/4.0/layout/grid/#no-gutters


also , i think giving space between two buttons will look more nicer so i think we can use above "no-gutter" method and then add the space manually in this section .

what do you think?

MonkeyDo:
> on .row to remove negative margin of row and extra gutters of columns

I'm not sure I understand your suggestion. Which .row element do you want to modify?
I don't follow how it will solve the issue.

We're using an older version of Bootstrap that doesn't have the `no-gutter` class, but that's not a huge issue as we can always define some css class for it.

frappelatte28:
> We're using an older version of Bootstrap that doesn't have the `no-gutter` class, but that's not a huge issue as we can always define some css class for it.

yes i agree on this .

also 
i have tested this code 
```
<div class="row">
  <div class="col-sm-6">
    <button type="button" class="btn btn-block btn-primary">Add another collaborator</button>
  </div>
  <div class="col-sm-6">
    <button type="submit" class="btn btn-block btn-success">Create collection</button>
  </div>
</div>
```

looks great to me , 
i should make a commit using this !

frappelatte28:
@MonkeyDo  please review the changes !

src/client/components/forms/userCollection.js
```diff
@@ -262,7 +262,8 @@ class UserCollectionForm extends React.Component {
 							<div className={errorAlertClass}>
 								<Alert bsStyle="danger">Error: {errorText}</Alert>
 							</div>
-							<div className="text-center">
+							<div className="row">
+								<div className = "col-sm-6">
```
MonkeyDo, line 266: You can add a simple Bootstrap class to add a space on top of both of these `col-sm-6` elements, which will give a bit of space between the buttons when they are on top of each other:

```suggestion
								<div className = "col-sm-6 margin-top-d5">
```

With that simply change on both elements, you can remove the changes to the `style.less` file.

src/client/stylesheets/style.less
```diff
@@ -549,3 +550,10 @@ hr.wide {
 	margin-right: 0.5em !important;;
 	vertical-align: text-top;
 }
+
+@media (max-width: 498px){
+	.buttons-container > button:nth-child(2) {
```
MonkeyDo, line 555: This `buttons-container` class is not used anywhere, so this code block is unused, and can be removed.

As mentioned in the comment above, we can add a bit of space between the buttons using existing bootstrap classes instead.

MonkeyDo (changes requested):
It's looking quite good !
A suggestion below to simplify the fix.

Also, currently our continuous integration (CI) [tests are failing](https://travis-ci.org/github/bookbrainz/bookbrainz-site/jobs/742261082#L701) on simple code style issues, which you can easily fix by running this command: `npm run lint -- --fix` before committing the changes to the `userCollection.js` file.

frappelatte28:
@MonkeyDo Please review the changes!

src/client/components/forms/userCollection.js
```diff
@@ -262,21 +262,25 @@ class UserCollectionForm extends React.Component {
...
-								</Button>
-								<Button
-									bsStyle="success"
-									type="submit"
-								>
-									<FontAwesomeIcon icon="save"/>&nbsp;{submitLabel}
-								</Button>
+							<div className="row">
+								<div className="col-sm-6 margin-top-d5">
+									<Button
```
MonkeyDo, line 267:
```suggestion
									<Button
										block
```

This makes the button take up all the available space, which I think looks quite nice:

Before:
Capture d’écran 2020-11-10 à 15 51 10


After:
Capture d’écran 2020-11-10 à 15 43 48

src/client/components/forms/userCollection.js
```diff
@@ -262,21 +262,25 @@ class UserCollectionForm extends React.Component {
...
+										bsStyle="primary"
+										type="button"
+										onClick={this.handleAddCollaborator}
+									>
+										<FontAwesomeIcon icon="plus"/>
+										&nbsp;Add another collaborator
+									</Button>
+								</div>
+								<div className="col-sm-6 margin-top-d5">
+									<Button
```
MonkeyDo, line 277:
```suggestion
									<Button
										block
```

Same as above.

src/client/components/forms/userCollection.js
```diff
@@ -262,21 +262,25 @@ class UserCollectionForm extends React.Component {
...
-									<FontAwesomeIcon icon="plus"/>
-									&nbsp;Add another collaborator
-								</Button>
-								<Button
-									bsStyle="success"
-									type="submit"
-								>
-									<FontAwesomeIcon icon="save"/>&nbsp;{submitLabel}
-								</Button>
+							<div className="row">
```
MonkeyDo, line 265:
```suggestion
							<div className="row margin-bottom-2">
```

This makes sure there's a bit of space between the buttons and the footer.

MonkeyDo (commented):
Just two more details and I think this will be perfect :)

src/client/components/forms/userCollection.js
```diff
@@ -262,21 +262,25 @@ class UserCollectionForm extends React.Component {
...
+									<Button
+										bsStyle="success"
+										type="submit"
+									>
+										<FontAwesomeIcon icon="save"/>&nbsp;{submitLabel}
+									</Button>
+								</div>
 								{
 									this.props.collection.id ?
 										<Button
```
MonkeyDo, line 286: I just noticed the third button ("Delete"), which is only visible when you edit one of your own collections.

It should receive the same changes as the other buttons, meaning it should also be wrapped in a `<div>` in the same way and the button should have the `block` attribute.

Here's the full block of code:

```jsx
{
	this.props.collection.id ?
		<div className="col-sm-6 margin-top-d5">
			<Button
				block
				bsStyle="danger"
				type="button"
				onClick={this.handleShowModal}
			>
				<FontAwesomeIcon icon="trash-alt"/>&nbsp;Delete collection
			</Button>
		</div> : null
}
```

With three buttons in mind, it's worth you having another pass to see if it looks good with the three buttons.

frappelatte28:
@MonkeyDo  Sorry for the late commit, I was occupied with some work, I have made the requested changes, feel free to let me know if you need any further changes.

MonkeyDo (approved):
Thank you for your PR, this now looks  a whole lot better !
:rocket:

PR #537, opened 2020-11-18
Eslint update

LordSputnik:
### Problem
We're using an outdated version of ESLint, which is unsupported by the current ESLint plugin ecosystem (including https://github.com/typescript-eslint/typescript-eslint).

### Solution
This PR updates ESLint from version 4 to version 7, upgrading plugins along the way where necessary, and replacing the deprecated eslint-loader webpack loader with its successor package, https://github.com/webpack-contrib/eslint-webpack-plugin.

### Areas of Impact
The only changes to the code are minor style tweaks.

The changes do also bump the minimum node version of bookbrainz-site from v8.0 to v10.13.0, which is required by eslint-webpack-plugin. But v8 is past its official end-of-life, so encouraging people to upgrade by making it unsupported is probably a good thing.


test/src/server/helpers/collections.js
```diff
@@ -29,8 +29,10 @@ import orm from '../../../bookbrainz-data';
 
 
 const {expect} = chai;
+/* eslint-disable import/no-commonjs */
 chai.use(require('chai-sorted'));
 chai.use(require('chai-arrays'));
```
MonkeyDo, line 34: These can be imported with ES6 module syntax instead (see [this example](https://github.com/bookbrainz/bookbrainz-site/blob/939b54fa9233ffc9c3da45f5e56ba8029361e1cb/test/src/server/routes/collection.js#L5))  to resolve the eslint issue.
Same goes for `test/src/server/helpers/revisions.js`

src/server/app.js
```diff
@@ -42,7 +42,7 @@ import session from 'express-session';
 
 
 // Initialize log-to-stdout  writer
-require('log-node')();
+require('log-node')(); // eslint-disable-line import/no-commonjs
```
MonkeyDo, line 45: This too can be changed to use ES6 import syntax instead
```javascript

import logNode from 'log-node';
[...]
// Initialize log-to-stdout  writer
logNode();
```

MonkeyDo (commented):
I didn't realize we were so far behind on eslint versions, thanks for the PR !
A couple of suggestions below to fix some warnings instead of disabling them.

I think deprecating Node 8 support is fine.

src/client/input.js
```diff
@@ -51,12 +30,34 @@ export default class Input extends Component {
...
+		wrapperClassName: null
+	};
+
 	constructor(props, context) {
 		super(props, context);
 
 		this.refFormControl = null;
 	}
 
+	// eslint-disable-next-line accessor-pairs
```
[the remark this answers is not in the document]
LordSputnik: Agreed. In fact, this Input component was only ever meant to be temporary. react-bootstrap removed their own Input component and replaced it with a building-block approach, so when we updated, we needed to re-create it as a wrapper around those building blocks. The idea was eventually that occurences might be replaced by bespoke combinations of building blocks, cutting away unnecessary logic/compatability code inside the Input class.

It might even be worth evaluating whether react-bootstrap is still the best option, or whether reactstrap or just using bootstrap directly would be better in some cases...

test/src/server/helpers/collections.js
LordSputnik: Done!

src/server/app.js
LordSputnik: Done!

MonkeyDo (approved):
Good to go ! 🚀 
Thanks !

PR #538, opened 2020-11-24
Typescript - Stage 2

LordSputnik:
### Problem
Switching from Flow to TypeScript. Stage 2 on the TypeScript roadmap in https://tickets.metabrainz.org/browse/BB-557

### Solution
Convert approximately 70 Flow-annotated JavaScript files to Typescript, over 7 commits. Each commit focuses on one area of the codebase. Some common transformations are applied:

Flow | TypeScript | Comment
-----|------------|--------
`mixed` | `unknown` |
`?<type>` | `<type> \| null \| undefined` |
`import {type X}` | `import {X} or import type {X}` |
`Array` | `any[]` |
`Object` | `Record<string, unknown>` or `any` | `Object` in Flow pretty meaningless, so the replacement types are mainly temporary until more descriptive annotations are written
`(number) => string` | `(argName: number) => string` |
`import X from 'x'` | `import * as X from 'x'` |
`ReactComponent` | `ReactComponent<any, any>` |
`X.displayName`<br/>`X.propTypes`<br/>`X.defaultProps` | `static displayName`<br/>`static propTypes`<br/>`static defaultProps` |
`React.Node`<br/>`React.Element` | `React.ReactNode`<br/>`React.ReactElement` |
`Action.metadata` | `Action.meta` | These were just typed incorrectly before - the code has always used `meta`, but the annotation had `metadata`. See https://github.com/redux-utilities/flux-standard-action
`?any` | `any` | Using the transformation in row 2, `?any` becomes `any \| null \| undefined`. But that's meaningless, because `any` already covers those other types.
`param: boolean = false` | `param = false` | Type is explicitly set by default value, so no need for annotation.

### Areas of Impact
Fixed bugs/typing issues in the following files:

* src/client/entity-editor/author-section/author-section.tsx
* src/client/helpers/utils.ts
* src/server/routes/entity/entity.tsx
* src/server/routes/merge.ts

All other files have no functional changes.


src/client/entity-editor/common/validation-label.tsx
```diff
@@ -16,13 +16,14 @@
...
-// @flow
 
 import * as React from 'react';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 
 
-function icon(empty: ?boolean, error: ?boolean, warn: ?boolean): string | null {
+type OptionalBool = boolean | null | undefined;
+type Icon = 'times' | 'exclamation-triangle' | 'check';
+function icon(empty: OptionalBool, error: OptionalBool, warn: OptionalBool): Icon | null {
```
MonkeyDo, line 26:
```suggestion
function icon(empty?: boolean, error?: boolean, warn?: boolean): Icon | null {
```

Here again I wonder if creating an OptionalBool type is necessary when `xyz?: boolean` is an optional bool.

src/client/entity-editor/edition-section/actions.ts
```diff
@@ -95,7 +93,7 @@ export function updateFormat(newFormatId: ?number): Action {
  *                 date.
  * @returns {Action} The resulting UPDATE_RELEASE_DATE action.
  */
-export function debouncedUpdateReleaseDate(newReleaseDate: ?object): Action {
+export function debouncedUpdateReleaseDate(newReleaseDate: number | null | undefined): Action {
```
MonkeyDo, line 96: this was `?object`, which doesn't help much determine the structure of the argument, but I think transposing it to `number` is a typo.

After looking at the code again, it should be  `string | null` (the DateField component that uses this action uses the `dateObjectToISOString` method under the hood)

src/client/entity-editor/edition-section/edition-section-merge.tsx
```diff
@@ -62,34 +62,35 @@ type EditionGroup = {
 };
 
 type OwnProps = {
-	mergingEntities: Array<object>
+	mergingEntities: any[]
 };
 
+type OptionalNumber = number | null | undefined;
```
MonkeyDo, line 68: Same remarks as above for OptionalBool

src/client/entity-editor/entity-editor.tsx
```diff
@@ -73,7 +72,7 @@ const EntityEditor = (props: Props) => {
 				<AliasEditor show={aliasEditorVisible} {...props}/>
 				<NameSection {...props}/>
 				<ButtonBar {...props}/>
-				<RelationshipSection {...props}/>
+				<RelationshipSection {...props as any}/>
```
MonkeyDo, line 75: Is this necessary? I don't see TS complain when I remove it.

Also, the next file in the git diff (entity-merge.tsx) uses `<RelationshipSection {...(props as any)}/>` with the extra parentheses, not sure if that makes a difference.

src/client/entity-editor/relationship-editor/actions.ts
```diff
@@ -30,10 +28,7 @@ export const UNDO_LAST_SAVE = 'UNDO_LAST_SAVE';
 
 export type Action = {
 	type: string,
-	payload?: mixed,
-	metadata?: {
-		debounce?: string
-	}
+	payload?: unknown
```
MonkeyDo, line 31: No `meta?` here compared to the others. Is this on purpose?

src/client/entity-editor/work-section/work-section-merge.tsx
```diff
@@ -48,7 +48,7 @@ type StateProps = {
 };
 
 type DispatchProps = {
-	onTypeChange: (?{value: number}) => mixed
+	onTypeChange: (value: number | null) => unknown
```
MonkeyDo, line 51: In other file, I believe you turned `?{value: number}` into `obj: {value: number} | null` instead, which I would even turn into `obj?: {value: number}`

Same for line 118 I suppose

src/server/helpers/props.ts
```diff
@@ -16,8 +16,12 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+import {Request, Response} from 'express';
```
MonkeyDo, line 19: Should this be `import type {Request, Response}…` ? 
Does it make any difference after transpilation?

src/server/routes/merge.ts
```diff
@@ -329,7 +333,7 @@ router.get('/submit/:targetBBID?', auth.isAuthenticated,
 			targetBBID = bbids[0];
 		}
 
-		if (!_.uniqBy(mergingFetchedEntities, 'type').length === 1) {
+		if (_.uniqBy(mergingFetchedEntities, 'type').length !== 1) {
```
MonkeyDo, line 336: Pffft. I don't know when I wrote that, but thanks for correcting it :D

MonkeyDo (commented):
Everything is building fine, which is very promising. I haven't seen any major issue.
Again, well done on  tackling this tedious task!

My main remark is about the use of `| null | undefined` everywhere to denote optional arguments.
Generally, I think the `function(argument?:<type>)` syntax is more readable than `function(argument:<type> | null | undefined)`, and unless I am mistaken they are equivalent.

src/client/entity-editor/alias-editor/actions.ts
```diff
@@ -87,7 +85,7 @@ export function debouncedUpdateAliasSortName(
  * @param {number} value - The new value to be used for the alias language ID.
  * @returns {Action} The resulting UPDATE_ALIAS_LANGUAGE action.
  */
-export function updateAliasLanguage(rowId: number, value: ?number): Action {
+export function updateAliasLanguage(rowId: number, value: number | null | undefined): Action {
```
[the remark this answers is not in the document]
LordSputnik: Flow and TypeScript both have the `param?: type` syntax for optional parameters. But most/all of these occurences of `<type> | null | undefined` are replacing `?<type>` rather than `param?`, which doesn't necessarily signify an optional parameter - just a [maybe type](https://flow.org/en/docs/types/maybe/). So I'm not sure we can say that all of them should be optional by default.

Taking `updateAliasLanguage` as an example, the only call to this function is in `src/client/entity-editor/alias-editor/alias-row.js`:

    function mapDispatchToProps(dispatch, {index}) {
            return {
                    onLanguageChange: (value) =>
                            dispatch(updateAliasLanguage(index, value && value.value)),
                    onNameChange: (event) =>
                            dispatch(debouncedUpdateAliasName(index, event.target.value)),
                    onPrimaryClick: (event) =>
                            dispatch(updateAliasPrimary(index, event.target.checked)),
                    onRemoveButtonClick: () =>
                            dispatch(removeAliasRow(index)),
                    onSortNameChange: (event) =>
                            dispatch(debouncedUpdateAliasSortName(index, event.target.value))
            };
    }

Since this is the only call, the 'value' parameter is never not supplied, so we can't really say that this should be optional. What's happening here is that value can be null (if the alias language is unselected). So we could maybe reduce the parameter type down to `number | null`, but I wouldn't be confident in doing that until we had better typing for the Select and LanguageField components to prove that this function can never receive `undefined`.

In summary, what I would say is this: this whole PR is a very minimal conversion from Flow to TypeScript, and there are a whole load of areas where we can improve typing and make things clearer. I think there'll be plenty of opportunities to do that once the basics are in. Until we have 100% type coverage and no `any` types anywhere, it's always a work-in-progress.

src/client/entity-editor/author-section/author-section.tsx
```diff
@@ -322,7 +322,7 @@ function mapDispatchToProps(dispatch: Dispatch<Action>): DispatchProps {
 		onEndDateChange: (endDate) =>
 			dispatch(debouncedUpdateEndDate(endDate)),
 		onEndedChange: (event) =>
-			dispatch(updateEnded(event.target.checked)),
+			dispatch(updateEnded((event.target as any).checked)),
```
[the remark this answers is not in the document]
LordSputnik: The reason I have `any` here is to mark it as something to come back to. Ideally we wouldn't need to do a type assertion here, because the `event` would have the correct type from the `onEndedChange` definition in `DispatchProps`. We could change it to HTMLInputElement, but I'm worried about it getting forgotten about (eslint will highlight any use of `any`).

src/client/entity-editor/common/validation-label.tsx
LordSputnik: Did a grep for icon, and it looks like it's only used inside `ValidationLabel`, where all three parameters are always provided in each call. I think the `?boolean` used to be there to indicate that the values could be null or undefined, because they come from the ValidationLabel props, where they are optional. Optional props will get destructured to `undefined`. I'd be happy to change this to `boolean | undefined` if you think that's better than what we have.

src/client/entity-editor/edition-section/actions.ts
LordSputnik: Agreed. I think this slipped in because we have no type information on `common/new-date-field.js`, so TypeScript didn't have any type information for `DateField.onChangeDate`. Will change.

src/client/entity-editor/relationship-editor/actions.ts
LordSputnik: Yeah, meta isn't used in any of these actions. The underlying problem here is that we're not doing action typing in the greatest way, but to keep the diff small I haven't improved that here. I really like the approach [here](https://redux.js.org/recipes/usage-with-typescript), where there's a separate interface for each action type, and then a union type for all actions for a single dispatcher/redux component.

src/client/entity-editor/entity-editor.tsx
LordSputnik: Looks fine to me too without it, I think it may have been left over from an earlier draft of the commit! Will remove.

src/client/entity-editor/work-section/work-section-merge.tsx
LordSputnik: A lot of these event handlers had incorrect typing with Flow. Most of the event handlers typed to receive an object parameter actually took a single value - you can tell from how it was being used in the dispatch function:

	return {
		onTypeChange: (value: ?{value: number}) =>
			dispatch(updateType(value))
	};

And then in the actions:

    export function updateType(newTypeId: ?number): Action {

So there was a conflict there, in that the updateType action creator is expecting a single numeric ID rather than an object. 

To back this up, the event is coming from `MergeField`, which internally uses a `Select` component from `react-select`. The Select component returns an object with `value` and `label` keys, which `MergeField` internally filters to extract the `value` and pass this directly to the `MergeField.onChange` handler. So it's right that the `onChange` handler for MergeField takes a single value, while the onChange handler in the original `WorkSection` component takes an object.

src/server/helpers/props.ts
LordSputnik: Should make no difference - Babel strips out all types and if there are no uses of imported symbols, it should remove the dependency too in the compiled code. I did a quick check of `lib/server/helpers/props.js` and looks like this holds true there.

babel-typescript has a [`onlyRemoveTypeImports` option](https://babeljs.io/docs/en/babel-preset-typescript#onlyremovetypeimports), which is false by default, implying that the default behavior is to remove all unused imports after transpilation.

src/client/entity-editor/relationship-editor/actions.ts
MonkeyDo: Yes, that looks really clean.

src/client/entity-editor/common/validation-label.tsx
MonkeyDo: No need, after all it works fine, is readable and declarative.

MonkeyDo (approved):
Agreeing with you on all counts :)
Let the correction of previous typing loosened be for another PR.

Mergeable once rebased! :rocket:

PR #546, opened 2021-01-11
Migrate to Webpack 5 and improve bundle size

LordSputnik:
### Problem
bookbrainz-site was dependent on webpack 4, and the bundle was 809 kB in size (compressed). Thought that I might be able to achieve some speed and build size improvements by switching to the new webpack 5. Additionally, there was a problem with the way assets were being resolved in the less pipeline (https://tickets.metabrainz.org/projects/BB/issues/BB-573).

### Solution
Switches to webpack 5, and update all of the webpack loaders. Updating less-loader and switching to webpack module resolution fixes BB-573. Also cut out some unneeded dependencies (e.g. style-loader), and followed the webpack migration steps for v4->v5 (https://webpack.js.org/migrate/5/), which cleaned up some deprecated features.

Once I was on webpack 5, I saw quite a bit of slow-down in the build process, and not much improvement in build size. So I looked into why tree-shaking wasn't working, and fixed that, then I updated the way we reference FontAwesome icons to allow them to be tree-shaken. I also removed moment.js, which was only used by Chart.js, and replaced it with date-fns, which was already in the bundle. As a result, bundle size has been reduced by about 60% (now ~350 kB compressed). Further size gains could be achieved by taking Chart.js out of the bundle (it's only used on one page, and 50 kB of the bundle), but that needs a bit of work to use dynamic importing. Also, lodash and immutable are taking up a bit more space than I would expect.

In terms of speed improvements, speed initially doubled going from v4 to v5, from 22 s to about 50 s. I analyzed the build speed using https://github.com/stephencookdev/speed-measure-webpack-plugin and found that ESLintPlugin was taking a long time to run, so I added a rule so that it only gets called for modules in "src/". This reduced build time back down to about 22 s.

### Areas of Impact
Mostly impacts the webpack config. FontAwesome elements in React pages have also been updated to use icon objects directly rather than string references to icons. The babel config has changed slightly. The main style.less file has had some imports updated. I've tested production compilation, the editor profile page (where the Chart.js chart is), server-side rendering and hot module replacement - all seem to work OK.


MonkeyDo (changes requested):
Nicely done, another good critical update. Thank you !
And what can I say about a 60% bundle size reduction other than "woohoo"‽

The image files you moved out of the `static` folder are being automatically copied back into `static/images/` by webpack, leaving my git directory dirty.
That doesn't look intentional; I guarantee some contributors will commit those files by mistake. I'm not exactly sure what is the best option here: have it committed twice?

There are still some oddities with the build system we'll eventually want to improve, but none that were introduced in this PR.

MonkeyDo:
There's also an ESLint issue:
```
/home/travis/build/bookbrainz/bookbrainz-site/src/client/entity-editor/common/validation-label.tsx
  22:1   error  Imports should be sorted alphabetically                                             sort-imports
  22:9   error  IconDefinition not found in '@fortawesome/react-fontawesome'                        import/named
  22:25  error  Member 'FontAwesomeIcon' of the import declaration should be sorted alphabetically  sort-imports
```

MonkeyDo (approved):
Thanks a lot!

I'll follow up by updating #544 after merging this.

PR #550, opened 2021-01-24
fix(BB-562): discrepancy in urls on entity pages

the-good-boy:
…pitalised W+ork)

### Problem

discrepancy in urls to other entities on entity pages

The urls to other entities (relationship) have a capitalised word (Work as opposed to work)

### Solution

I fixed the url discrepancy

### DIFF
--- a/src/client/components/pages/entities/work-table.js
+++ b/src/client/components/pages/entities/work-table.js
@@ -54,7 +54,7 @@ function WorkTableRow({showAddedAtColumn, work, showCheckboxes, selectedEntities
 							onClick={() => onToggleRow(work.bbid)}
 						/> : null
 				}
-				<a href={`/Work/${work.bbid}`}>{name}</a>
+				<a href={`/work/${work.bbid}`}>{name}</a>
 				{disambiguation}
 			</td>
 			<td>{languages}</td>


MonkeyDo (commented):
Hi @the-good-boy !
Thanks for opening this PR.
That's a great start, one capitalization mistake fixed and issue 562 fixed ! :)

I found another —much less obvious— possible capitalization issue on this line which we can fix in this PR too: https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/name-section/name-section.js#L190

Here, entityType can be `author`,`work`, etc. which works fine, but it can also be `editionGroup` and the resulting URL will be invalid (see screenshot below).
We need to use a lodash utility to transform the entityType into [kebab case](https://lodash.com/docs/4.17.4#kebabCase), like we do on this line:
https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/entity-link.js#L32

Capture d’écran 2021-01-25 à 12 40 04


And thank you for coming to IRC to report the issue with the automatic github action.
I have tried to push a fix the issue, and if you are willing to help me test it you can update (or rebase) this PR from master.
We'll see if the action works with the changes I made.

the-good-boy:
Hello @MonkeyDo 
Yeah sure! I'll update that one too.

PR #551, opened 2021-01-25
fix(display): url discrepancy in entities and entity-editor name-section

the-good-boy:
### Problem

Fixed the url discrepancy in entities (earlier it used to appear as 'W+ork' instead of 'work' in url).
Also there was a possible discrepancy in name-section url where kebab-casing was missing 

### Solution

A pretty straight-forward fix for the first case and a lodash kebabCasing in the second case

### DIFF
--- a/src/client/entity-editor/name-section/name-section.js
+++ b/src/client/entity-editor/name-section/name-section.js
@@ -187,7 +187,7 @@ class NameSection extends React.Component {
 											(
 												<ListGroupItem
 													bsStyle="warning"
-													href={`/${entityType}/${match.bbid}`}
+													href={`/${_kebabCase(entityType)}/${match.bbid}`}
 													key={`${match.bbid}`}
 													rel="noopener noreferrer" target="_blank"
 												>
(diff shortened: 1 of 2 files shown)


src/client/entity-editor/name-section/name-section.js
```diff
@@ -187,7 +187,7 @@ class NameSection extends React.Component {
 											(
 												<ListGroupItem
 													bsStyle="warning"
-													href={`/${entityType}/${match.bbid}`}
+													href={`/${_kebabCase(entityType)}/${match.bbid}`}
```
MonkeyDo, line 190: As you can see below there is an error reported: `'_kebabCase' is not defined`.

In the other file we imported it with the name `_kebabCase`, but in this one we import the whole lodash library as `_` (https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/name-section/name-section.js#L44).

So you can call the method like this: `_.kebabCase` with the extra dot.

MonkeyDo (changes requested):
One small correction and we're good to go :)

For your reference, once a pull request is open, you can still push commits to it.
So you don't have to close the PR and open another one if you've got modifications.

src/client/entity-editor/name-section/name-section.js
the-good-boy: Thanks @MonkeyDo for the feedback. I'll try not to make careless mistakes like these in the future.

MonkeyDo (approved):
All good!
Thanks a lot for opening a PR and fixing this issue !

PR #553, opened 2021-02-08
fix(collection-form):Position of "Add another collaborator" button

the-good-boy:
### Problem

Position of "Add another collaborator" button is not intutive.

### Solution

Instead of displaying the button at the bottom, I have moved it next to the Collaborator's heading as suggested.

ss done


MonkeyDo (approved):
Thank you for his improvement !

PR #559, opened 2021-03-11
FIX (BB-456) : CSS style issue in older version of certain browsers

akashgp09:

### Problem

Some Older version of CSS does not support certaion CSS properties

### Solution

By adding vendor pefix we can tackle this

### Areas of Impact

CSS Styles

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -283,6 +283,7 @@ a.contact-text:visited {
 }
 
 .deleted {
+	text-decoration: @brand-danger line-through;
 	text-decoration-line: line-through;
 	text-decoration-color: @brand-danger;
 	-webkit-text-decoration-line: line-through;
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -283,15 +283,8 @@ a.contact-text:visited {
 }
 
 .deleted {
-	text-decoration: @brand-danger line-through;
-	text-decoration-line: line-through;
-	text-decoration-color: @brand-danger;
-	-webkit-text-decoration-line: line-through;
-	-webkit-text-decoration-color: @brand-danger;
-	-moz-text-decoration-line: line-through;
-	-moz-text-decoration-color: @brand-danger;
-	-ms-text-decoration-line: line-through;
-	-ms-text-decoration-color: @brand-danger;
+	text-decoration: line-through;
+	color: @brand-danger;
 }
 
 /* getStarted.jade */
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -31,11 +31,12 @@
 // @import "~react-virtualized-select/styles.css";
 // @import "~react-datepicker/dist/react-datepicker.css";
 
+
 /* Modifications to Lobes */
 /* ====================== */
 @brand-secondary: @bookbrainz;
-@disabled-bg: lighten(desaturate(@bookbrainz, 10%), 42%);
-@disabled-color: lighten(@disabled-bg, 15%);
+@disabled-bg: lighten(desaturate(@bookbrainz,10%),42%);
+@disabled-color: lighten(@disabled-bg,15%);
 
 .navbar.BookBrainz {
 	background: rgba(255, 255, 255, 1);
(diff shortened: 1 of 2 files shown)


MonkeyDo (commented):
Hi @akashgp09 !
Thanks for opening this PR !

I think you have some form of auto-formatting set up for your code editor, that has modified the file a lot and makes it pretty hard to review, especially for a change of a few lines.

Could I ask you to redo it with the auto-formatting disabled in your editor?

You should be able to checkout the version of the less file from the master branch with `git checkout master -- src/client/stylesheets/style.less`, reapply your changes to the `.deleted` class and commit again, without having to open another PR.

Have you been able to test in the older browser versions mentioned in the ticket?

akashgp09:
> Could I ask you to redo it with the auto-formatting disabled in your editor?

sure @MonkeyDo, I have reverted the prettier formatting :)

> Have you been able to test in the older browser versions mentioned in the ticket?

I didn't tested it. But i read few article for the solution and also looked into the MDN docs, it was suggested that using the vendor prefix can solve some css issue for older browser versions

src/client/stylesheets/style.less
```diff
@@ -435,13 +441,6 @@ hr.wide {
 		padding-bottom: 0.3em;
 	}
 	.note-content {
-		quotes: "\201C" "\201D";
```
MonkeyDo, line 438: This removal is part of another PR if I'm not mistaken.

MonkeyDo (changes requested):
This hasn't fixed the issue on the browsers mentioned in the ticket. You can test using BrowserStack for example: https://www.browserstack.com

I suppose it doesn't hurt to keep the prefixes you added if it makes it work in some browsers we don't know about.

However I realized while checking browser support on can I use that the css property `text-decoration: line-through;` will work for all the browsers who don't support `text-decoration-line` 
https://caniuse.com/text-decoration says
> All browsers support the CSS2 version of text-decoration, which matches only the text-decoration-line values (underline, etc.)

So that seems like a good solution to add along with the existing css as a fallback.
However that strikethrough is the same grey color as the text, so maybe there's something there to investigate to see if it can be changed: https://developer.mozilla.org/fr/docs/Web/CSS/text-decoration

src/client/stylesheets/style.less
akashgp09: @MonkeyDo I have reverted this  and added the `text-decoration` css property , let me  know if there's any changes needed :)

src/client/stylesheets/style.less
```diff
@@ -283,8 +283,15 @@ a.contact-text:visited {
...
 .deleted {
+	text-decoration: @brand-danger line-through;
 	text-decoration-line: line-through;
 	text-decoration-color: @brand-danger;
+	-webkit-text-decoration-line: line-through;
+	-webkit-text-decoration-color: @brand-danger;
+	-moz-text-decoration-line: line-through;
+	-moz-text-decoration-color: @brand-danger;
+	-ms-text-decoration-line: line-through;
+	-ms-text-decoration-color: @brand-danger;
```
MonkeyDo, line 294: I am not finding any documentation for this prefix. Does this actually exist for microsoft browsers?

src/client/stylesheets/style.less
```diff
@@ -283,8 +283,15 @@ a.contact-text:visited {
 }
 
 .deleted {
+	text-decoration: @brand-danger line-through;
```
MonkeyDo, line 286: This doesn't work everywhere I tested. On IE 11 and Opera 43 for example `text-decoration: line-through` works, but `text-decoration: $SOME_COLOR line-through;` doesn't.

We can add both one on top of each other; the browsers that don't recognize the color form will fallback automatically to the  form without the color.

MonkeyDo (changes requested):
This is still not working as expected on the browsers I tried.
Please do test that your solution works; I know in this case testing on ancient browsers seems really tricky, but you can use BrowserStack with a free account to visit the website on various browsers (specifically the browsers listed in the ticket).

What I then do to test is visit [a deleted entity](https://bookbrainz.org/author/e8d307d9-d656-4f29-bc16-cd6963634ade) and use whatever developer tools there are on that old  browser to inspect and modify the css, to see what works/doesn't work.

That way you'll be able to find a solution that works everywhere, with progressive fallbacks :)

src/client/stylesheets/style.less
akashgp09: yes **ms** is a valid prefix https://bitsofco.de/css-vendor-prefixes/

src/client/stylesheets/style.less
akashgp09: Tried a different approach this time, tested it in all the listed browser. Works on all of them :)
Also i have removed all the css vendor prefixes because **text-decoration** property is well supported in all the browsers.

src/client/stylesheets/style.less
MonkeyDo: yes, I do know it :)
But looking at the [support table for `text-decoration-line`](https://developer.mozilla.org/fr/docs/Web/CSS/text-decoration-line#compatibilit%C3%A9_des_navigateurs), there is no support for the property on IE at all, so `-ms-text-decoration-line` doesn't do anything.

Now they're gone entirely, we don't have to worry about it :)

src/client/stylesheets/style.less
MonkeyDo: Great, that looks even simpler :) I'll look at it and try to deploy to test.bookbrainz.org so we can test IRL

src/client/stylesheets/style.less
akashgp09: > But looking at the [support table for `text-decoration-line`](https://developer.mozilla.org/fr/docs/Web/CSS/text-decoration-line#compatibilit%C3%A9_des_navigateurs), there is no support for the property on IE at all, so `-ms-text-decoration-line` doesn't do anything.

my bad  ( * _ * )
> Now they're gone entirely, we don't have to worry about it :)

Yes indeed : )

MonkeyDo (approved):
That's a great solution, nice and simple, well done !

PR #561, opened 2021-03-11
[DOCS] : Fixed PULL_REQUEST_TEMPLATE

akashgp09:
### Problem

Incorrect order of number in PR Template

1
1
1

### Solution

It should  be

1
2
3


MonkeyDo (approved):
Thanks !

PR #562, opened 2021-03-11
Previous and next page buttons now have have responsive design in collections

bouhormq:
### Problem
https://tickets.metabrainz.org/browse/BB-579

### Solution
By changing the order in which the buttons are rendered, whenever the screen gets smaller, now the next and previous buttons are aligned and also added some margin in the bottom so that way there is some space between the lower level buttons and the top ones.

Before
BB
 After
BBetter (1)

### Areas of Impact
Only ascetic ones in the collections component.

### DIFF
--- a/src/client/entity-editor/common/sort-name-field.tsx
+++ b/src/client/entity-editor/common/sort-name-field.tsx
@@ -166,7 +166,7 @@ function SortNameField({
 			label={label}
 			ref={(node) => { input = node; }}
 			tooltipText="Alphabetical sorting name. Examples: 'Dickens, Charles', 'Christmas Carol, A'.
-			<br>You can try to fill it automatically with the guess button"
+			You can try to fill it automatically with the guess button"
 			type="text"
 			onChange={onChange}
 			{...rest}
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -414,6 +414,10 @@ hr.wide {
 	}
 }
 
+.next, .previous > a {
+	margin-bottom: 5%;
+}
+
 @media only screen and (max-width: 600px) {
 	#searchpage-button-group > * {
 		width: 100%;
--- a/src/client/components/pages/parts/pager.js
+++ b/src/client/components/pages/parts/pager.js
@@ -143,6 +143,12 @@ class PagerElement extends React.Component {
 								>
 									&larr; Previous Page
 								</Pager.Item>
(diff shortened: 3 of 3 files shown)


src/client/stylesheets/style.less
```diff
@@ -414,6 +414,10 @@ hr.wide {
 	}
 }
 
+.next, .previous > a {
```
MonkeyDo, line 417: This applies differently for next and for previous: you target elements with the .next class as well as <a> elements inside a parent with the class .previous (but not the element with the .previous class).

Using a percentage here can be tricky in some situations; instead let's use `1em` (or 0.7em, or whatever number you feel is right, but as an `em` value).

I would also like to make sure this applies only where we want it to, so altogether I suggest something like this: 
```suggestion
#pageWithPagination {
  .next, .previous {
    > a {
      margin-bottom: 1em;
    }
  }
}
```

src/client/entity-editor/common/sort-name-field.tsx
```diff
@@ -166,7 +166,7 @@ function SortNameField({
 			label={label}
 			ref={(node) => { input = node; }}
 			tooltipText="Alphabetical sorting name. Examples: 'Dickens, Charles', 'Christmas Carol, A'.
-			<br>You can try to fill it automatically with the guess button"
+			You can try to fill it automatically with the guess button"
```
MonkeyDo, line 169: This should not be part of this PR, you'll need to revert this change (and open another PR if you want to fix it ;) )

MonkeyDo (changes requested):
Thanks for opening the PR !

It's a really nice and simple trick to solve that problem !
How did you find that ? It was under our noses the whole time :D

src/client/entity-editor/common/sort-name-field.tsx
bouhormq: Sure thing!

src/client/stylesheets/style.less
bouhormq: Perfect! I'll do some testing to see if it only applies to those exact buttons

MonkeyDo:
@bouhormq Hi! Let me know if you have some time to finish this PR, otherwise do let me know and I'll finish it :)

MonkeyDo:
Closing this as it is getting stale; please feel free to comment if you want to finish it and I'll reopen the PR :)

PR #563, opened 2021-03-12
Chore (BB-586): Added prettier and husky configuration

akashgp09:
### Problem

**This PR Fixes:** [BB-586](https://tickets.metabrainz.org/projects/BB/issues/BB-586?filter=myopenissues)


.eslintrc.js
```diff
@@ -12,14 +11,11 @@ const options = {
 		'plugin:node/recommended',
 		'plugin:react/recommended',
 		'plugin:import/recommended',
-		'plugin:@typescript-eslint/recommended'
+		'plugin:@typescript-eslint/recommended',
+		'plugin:prettier/recommended'
```
INNOVATIVEGAMER, line 15: Hi, @akashgp09 Do we require this line "plugin:prettier/recommended" to work prettier with eslint ?
Try removing this line to get rid of lint error on checks.

akashgp09: I read somewhere that it enables the main config from eslint-config-prettier, which turns off some ESLint core rules that conflict with Prettier.

akashgp09: also the reason the link check is failing is because it cannot find **eslint-plugin-prettier** dependency.

akashgp09: @MonkeyDo should we remove the line from the eslint file as suggested by @NNOVATIVEGAMER or is there any work around to fix it

INNOVATIVEGAMER: > I imagine that it enables the main config from eslint-config-prettier, which turns off some ESLint core rules that conflict with Prettier. seen to be used commonly

Yes , you are right. Have you tried reinstalling the "eslint-plugin-prettier" package.

akashgp09: Yes I did. And the linter isn't throwing any error locally. I think there is some issue with the lint Runner workflow.

.eslintrc.js
```diff
@@ -12,14 +11,11 @@ const options = {
...
 	],
 	parser: '@typescript-eslint/parser',
-	plugins: [
-		'react',
-		'import',
-		'@typescript-eslint'
-	],
+	plugins: ['react', 'import', '@typescript-eslint', 'prettier'],
```
INNOVATIVEGAMER, line 18: As you added `{
  "extends": ["plugin:prettier/recommended"]
}`
You do not need to specify 'prettier' in plugins.

INNOVATIVEGAMER:
You worked a lot on this but if nothing works out.
I have a another solution @akashgp09 .
So if we want to use eslint and prettier for code formatting and want to remove these plugin related errors and downsides, 
Prettier provides us the solution by installing package called `prettier-eslint`.
Please read this section suggested by prettier and you will get the point that I want to say.
https://prettier.io/docs/en/integrating-with-linters.html#notes
and please try going with this solution ;)

akashgp09:
@INNOVATIVEGAMER I really appreaciate your suggestions on this. But before moving forward i would like to have some feedbacks from the project maintainers on this and proceed ahead. Thank You

PR #564, opened 2021-03-12
FIX(BB-587): Better button alignment for smaller viewports

akashgp09:
### Problem

Thir PR Fixes: [BB-587](https://tickets.metabrainz.org/projects/BB/issues/BB-587?filter=allopenissues)

### Solution

**Added a media query for smaller viewports**

## Before:

WhatsApp Image 2021-03-12 at 14 25 08

## After:

WhatsApp Image 2021-03-12 at 14 25 08 (1)

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -553,4 +553,12 @@ hr.wide {
 	vertical-align: text-top;
 }
 
+.text-collection-entity{
+    text-align: right;
+}
 
+@media only screen and (max-width: 720px){
+    .text-collection-entity{
+        text-align: center;
+    }
+}
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -553,11 +553,11 @@ hr.wide {
 	vertical-align: text-top;
 }
 
-.text-collection-entity{
+.collection-page-buttons {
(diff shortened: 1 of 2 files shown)


src/client/stylesheets/style.less
```diff
@@ -553,4 +553,12 @@ hr.wide {
 	vertical-align: text-top;
 }
 
+.text-collection-entity{
+    text-align: right;
+}
 
+@media only screen and (max-width: 720px){
```
MonkeyDo, line 560: Here we can use the bootstrap variable for screen sizes that is used in other places:
```suggestion
@media only screen and (max-width: @screen-sm-max){
```

src/client/stylesheets/style.less
```diff
@@ -553,4 +553,12 @@ hr.wide {
 	vertical-align: text-top;
 }
 
+.text-collection-entity{
```
MonkeyDo, line 556: What do you think about `.collection-page-buttons` as a class name instead?

MonkeyDo (changes requested):
A nice simple improvement !
Thanks !

MonkeyDo (approved):
Thanks for this PR ! :rocket:

PR #567, opened 2021-03-12
fix(toolTipText):<br> tag not working as expected

the-good-boy:
### Problem

The line break tag in tooltip text was not working as expected

### Solution

I tried many possible solutions, but it turns out this is the only one which worked for me. I read somewhere that using 'dangerouslySetInnerHTML' can cause potential security issues, but I dont think we need to worry about it in this particular case. 
sseditiontooltip
ssworktooltip


endurance21:
@the-good-boy some lint tests are failing , please see if you missed the linting part ?

MonkeyDo (commented):
Well you put me on the trail of the issue !

Instead of using dangerouslySetHtml (I try to avoid it if there's any other solution :) ) we can actually pass some html as the tooltipText like this:
```
tooltipText={
	  <>
		  Group together different Editions of the same book.
		  <br/>For example paperback, hardcover and e-book editions.
	  </>
}
```

How do you feel about hunting in the codebase for tooltip texts that use html code and should passed as an html fragment instead of a string?

MonkeyDo (approved):
Thanks  for this change :) :rocket:

PR #568, opened 2021-03-13
Fix (BB-589):  Footer buttons get compressed and the text can't be read

akashgp09:
### Problem
This PR Fixes: [BB-589](https://tickets.metabrainz.org/browse/BB-589)

### Solution

Fixed the design issue by adding some custom `CSS `and removing react bootstrap ` <ButtonGroup />` Component

### Before

https://user-images.githubusercontent.com/55311336/111028258-824d0080-841b-11eb-8da3-de8ece5f2657.mp4

### After

https://user-images.githubusercontent.com/55311336/111028270-97c22a80-841b-11eb-866b-9bf9148bd8ad.mp4


src/client/stylesheets/style.less
```diff
@@ -537,20 +572,24 @@ hr.wide {
...
-	margin-right: 0.5em !important;;
+	margin-right: 0.5em !important;
 	vertical-align: text-top;
 }
 
-
+.custom-width {
+	max-width: 200px;
+	min-width: 175px;
+	margin-top: 8px;
```
endurance21, line 594: should not we have used  bootstrap classes for the same ?
`margin-top-2` or something like that .this will increas code redablity and redundancy.

src/client/stylesheets/style.less
```diff
@@ -537,20 +572,24 @@ hr.wide {
...
 }
 
 .checkboxes {
-	margin-right: 0.5em !important;;
+	margin-right: 0.5em !important;
 	vertical-align: text-top;
 }
 
-
+.custom-width {
```
endurance21, line 591: is it used at many places ? 
or if it is used at some specific location for eg. footer then the className must be more intuitive

src/client/stylesheets/style.less
akashgp09: i don't think so that it is making our code less readable. and writing a single css for margin would be better than duplicating `mt-2` for all the button classes.

src/client/stylesheets/style.less
akashgp09: thanks for the review, I agree with you on this.

MonkeyDo (commented):
At a glance looking at your screen captures (thanks for that!) the solution looks great !

Unfortunately like PR #559 I can't review this properly as there are so many changes due to auto-formatting.
You can follow the instructions I left on the other PR to revert the changes and reapply just your intentional changes.
Thanks !

akashgp09:
@MonkeyDo i have made the changes, thank you for the review :)

MonkeyDo (approved):
Thanks! That looks a lot better!

PR #571, opened 2021-03-14
feat(Collections): added a 'No.of Entities' column to Collections table

the-good-boy:
### Problem

https://tickets.metabrainz.org/browse/BB-580
Adding a column which displays No. of entities in a collection 

### Solution

I have created a helper function getCollectionSize() which returns the count of UserCollectionItems in each collection. I then used this function to add a property named 'noOfEntities' to each item in the collectionsJSON array which was returned from getOrderedPublicCollections()

### Areas of Impact

Now the function getOrderedPublicCollections() will return an additional property with each collection object : noOfEntities

CollectionsTableNew

### DIFF
--- a/src/client/components/pages/parts/collections-table.js
+++ b/src/client/components/pages/parts/collections-table.js
@@ -64,7 +64,7 @@ class CollectionsTable extends React.Component {
 						{_.startCase(entityType)}
 					</MenuItem>
 				))}
-				<MenuItem divider />
+				<MenuItem divider/>
 				<MenuItem
 					eventKey={null}
 					key="allTypes"
@@ -79,7 +79,7 @@ class CollectionsTable extends React.Component {
 				href="/collection/create"
 				type="button"
 			>
-				<FontAwesomeIcon icon={faPlus} />
+				<FontAwesomeIcon icon={faPlus}/>
 				&nbsp;Create Collection
 			</Button>
 		);
@@ -94,7 +94,7 @@ class CollectionsTable extends React.Component {
 						{entityTypeSelect}
 					</div>
 				</div>
-				<hr className="thin" />
+				<hr className="thin"/>
 				{
 					results.length > 0 ?
 						<Table
@@ -166,7 +166,7 @@ class CollectionsTable extends React.Component {
 
 						<div>
 							<h4> No collections to show</h4>
-							<hr className="wide" />
+							<hr className="wide"/>
 						</div>
 				}
 			</div>
--- a/src/client/components/pages/parts/collections-table.js
+++ b/src/client/components/pages/parts/collections-table.js
@@ -18,16 +18,16 @@
 
 import * as bootstrap from 'react-bootstrap';
 import * as utilsHelper from '../../../helpers/utils';
-import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
 import React from 'react';
 import _ from 'lodash';
-import {faPlus} from '@fortawesome/free-solid-svg-icons';
-import {genEntityIconHTMLElement} from '../../../helpers/entity';
+import { faPlus } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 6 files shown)


src/client/components/pages/parts/collections-table.js
```diff
@@ -106,6 +106,7 @@ class CollectionsTable extends React.Component {
 									<th className="col-sm-2">Name</th>
 									<th className="col-sm-4">Description</th>
 									<th className="col-sm-2">Entity Type</th>
+									<th className="col-sm-2">No. of Entities</th>
```
akashgp09, line 109:
```suggestion
									<th className="col-sm-2">Total Entities</th>
```
a more better name for this column would be `Total Entities`

akashgp09 (changes requested):

src/server/helpers/collections.js
```diff
@@ -104,6 +119,16 @@ export async function getOrderedPublicCollections(from, size, entityType, orm) {
...
 
+	// We need to add the no. of entities in each collection
+	collectionsJSON.forEach((collection) => {
+		collection.noOfEntities = getCollectionSize(collection.id, orm);
+	});
+	// Resolving the promises
+	await Promise.all(collectionsJSON.map(async collection => {
+		const resolvedValue = await collection.noOfEntities;
+		collection.noOfEntities = resolvedValue;
+	}));
```
MonkeyDo, line 130: After some thought, I think we can do this in a more optimized manner on the ORM side.
In the [UserCollection model](https://github.com/bookbrainz/bookbrainz-data-js/blob/master/src/models/userCollection.js), we should be able to add something like that:
```
itemCount() {
	return this.hasMany('UserCollectionItem', 'collection_id').count();
},
```

And then when fetching the collections, we can request our new `itemCount`:
```
const allCollections = await new UserCollection()
.where(…
.fetchPage({
	limit: size,
	offset: from,
	withRelated: ['owner', 'itemCount'] // <-- Adding it here
});
```

That way, the ORM will only have to do a COUNT  SQL query in the same transaction which should be faster, while also making it reusable in other parts of the codebase.

Now… we install our ORM bookbrainz-data-js as a package from NPM.
To test your changes locally there's a few more steps:
1. you'll need to clone the repo locally
2. from `bookbrainz-data-js`, run the command `npm link`
3. once that is done, from `bookbrainz-site` run `npm link bookbrainz-data`

What that does is create a local NPM package and substitutes it in the website repo.
With that, during development when you make changes to `bookbrainz-data-js` you can simply rebuild (`npm run build-js`) and then restart the web server.

src/client/components/pages/parts/collections-table.js
MonkeyDo: I would even just say "Entities". I think in the context of this table it's self explanatory.

MonkeyDo (changes requested):
Looking good :)

I'm requesting some pretty major implementation changes, but only to move the code to the ORM where it will be reusable.
Your logic works fine :)

MonkeyDo:
I deployed a new version of the ORM package. You'll need to run `npm i --save bookbrainz-data@2.9.0` and commit both the package.json and package-lock.json files.

the-good-boy:
That's great to hear, Thanks! I had made all other required changes to this pr beforehand, only updating the package was left afaik. I think this issue can be closed now. Do let me know if you want me to add any more specific tests to the UserCollections orm, I will be excited to help. My end-sem exams are on these days(will be more free after a week), but I will be glad to add more tests if necessary.

MonkeyDo (changes requested):
After testing locally, I see everything working perfectly for the collections page, nice work!

There's still one place to add that `withItemCount` to so that on the user profile collections (for example localhost:9099/editor/1245/collections) the itemCount also appears; currently the column is there but with no values.
You'll have to add it here: https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/helpers/collections.js#L64-L67

the-good-boy:
Ah! I see. I completely missed that. Will do it soon.

MonkeyDo (approved):
Great job on all that, and thank you !

PR #353, opened 2020-01-26
React fontawesome #bb375

prabalsingh24:
### Problem
https://tickets.metabrainz.org/browse/BB-375
Migrating React-FontAwesome : parts/editor-profile.js

### Areas of Impact
src/client/components/pages/parts/editor-profile.js

PR #572, opened 2021-03-14
FIX(BB-564) : Converted revision notes string url to clickable links

akashgp09:
### Problem

**This PR Fixes: [BB-564](https://tickets.metabrainz.org/projects/BB/issues/BB-564?filter=allopenissues)**

### Solution

**Used `regex` to convert any string url that has a prefix `http|https|ftp|ftps` to a  clickable `link` and then rendered the HTML string as real HTML.**

### Before

https://user-images.githubusercontent.com/55311336/111072915-a46d7e00-8502-11eb-92e8-791fd37e7fe0.mp4

### After

https://user-images.githubusercontent.com/55311336/111072922-ab948c00-8502-11eb-9089-6d97c4bdc3c2.mp4


src/client/components/pages/parts/revisions-table.js
```diff
@@ -116,11 +116,19 @@ function RevisionsTable(props) {
 														revision.notes.map(note => (
 															<div className="revision-note clearfix" key={note.id}>
 																<span className="note-content">
-																	{note.content}
+																	<span
+																		dangerouslySetInnerHTML={{
```
akashgp09, line 120: hey @MonkeyDo  is it fine here to use `dangerouslySetInnerHTML` , if not can you suggest any alternate option we can try.

MonkeyDo: I like to add the package `DOMPurify` whenever I use dangerouslySetInnerHTML , just to be on the safe side .

Here a good breakdown of how it's used: https://linguinecode.com/post/complete-guide-react-dangerouslysetinnerhtml

akashgp09: thank you, will make the changes soon :)

akashgp09:
I have added the DOMPurify package, but using DOMPurify.sanitize throws **TypeError: dompurify_1.default.sanitize is not a function**. I don't know what's causing this, so i have opened this [issue](https://github.com/cure53/DOMPurify/issues/526) in DOMPurify, let's see if we can get some insights on this. 
Related issue: [#353](https://github.com/cure53/DOMPurify/issues/353)

akashgp09:
I have added the DOMPurify package, but using DOMPurify.sanitize throws **TypeError: dompurify_1.default.sanitize is not a function**. I don't know what's causing this, so i have opened this [issue](https://github.com/cure53/DOMPurify/issues/526) in DOMPurify, let's see if we can get some insights on this. 
Related Issue: [#353](https://github.com/cure53/DOMPurify/issues/353)

MonkeyDo:
> I have added the DOMPurify package, but using DOMPurify.sanitize throws **TypeError: dompurify_1.default.sanitize is not a function**. I don't know what's causing this, so i have opened this [issue](https://github.com/cure53/DOMPurify/issues/526) in DOMPurify, let's see if we can get some insights on this.
> Related Issue: [#353](https://github.com/cure53/DOMPurify/issues/353)

It is probably because of missing typescript types.
You'll need to installe the `@types/dompurify` package: https://www.npmjs.com/package/@types/dompurify

Let me know if that fixes it.

akashgp09:
> It is probably because of missing typescript types.
> You'll need to install the `@types/dompurify` package: https://www.npmjs.com/package/@types/dompurify
> Let me know if that fixes it.

I installed this package but still getting the error.

MonkeyDo:
> > Let me know if that fixes it.
> 
> I installed this package but still getting the error.
> 
> I tried debugging it using console.log
> 
> ```
> import DOMPurify from "dompurify";
> console.log(DOMPurify);  // [Function: DOMPurify] { version: '2.2.6', removed: [], isSupported: false }
> console.log(DOMPurify.sanitize); // undefined
> ```
> 
> DOMPurify.sanitize is **undefined**

Hm. In another project I imported it like this:  `import { sanitize } from "dompurify";`  and then used simple with `sanitize(…)` . Maybe try that?

akashgp09:
> Hm. In another project I imported it like this: `import { sanitize } from "dompurify";` and then used simple with `sanitize(…)` . Maybe try that?

Tried this but same results. Can you check it on this branch on your **system**. Or what about using **[sanitize-html](https://www.npmjs.com/package/sanitize-html)** package alternative to dompurify.

MonkeyDo:
> > Hm. In another project I imported it like this: `import { sanitize } from "dompurify";` and then used simple with `sanitize(…)` . Maybe try that?
> 
> Tried this but same results. Can you check it on this branch on your **system**. Or what about using **[sanitize-html](https://www.npmjs.com/package/sanitize-html)** package alternative to dompurify.

I ran `npm install --save dompurify`(the PR is currently missing that change in package.json and package-locks.json), then in revision-table.js:
`import {sanitize} from 'dompurify';`
 and later on:
 ```
 <span
	// eslint-disable-next-line react/no-danger
	dangerouslySetInnerHTML={{
		__html: sanitize(note.content.replace(
			// eslint-disable-next-line max-len
			/((http|https|ftp|ftps)\:\/\/[a-zA-Z0-9\-\.]+\.[a-zA-Z]{2,3}(\/\S*)?)/g,
			'<a target="_blank" href="$1">$1</a>'
		))
	}}
/>
 ```

Everything works as expected, no warnings of any kind.

src/client/components/pages/parts/revisions-table.js
```diff
@@ -116,9 +117,19 @@ function RevisionsTable(props) {
 														revision.notes.map(note => (
 															<div className="revision-note clearfix" key={note.id}>
 																<span className="note-content">
-																	{note.content}
+																	<span
+																		// eslint-disable-next-line react/no-danger
+																		dangerouslySetInnerHTML={{
+																			__html: sanitize(note.content.replace(
+																				// eslint-disable-next-line max-len
+																				/((http|https|ftp|ftps)\:\/\/[a-zA-Z0-9\-\.]+\.[a-zA-Z]{2,3}(\/\S*)?)/g,
```
MonkeyDo, line 125: Regarding the url extension part, `.[a-zA-Z]{2,3}` is too restrictive. For example `http://www.test.info/` will return `http://www.test.inf` with a missing final 'o'.
I also wonder if we should also convert `www.test.info` without the protocol at the start. What do you think?

A quick google search an i found this which seems to solve both issues:
https://www.urlregex.com/

MonkeyDo (changes requested):

akashgp09:
>Regarding the url extension part, .[a-zA-Z]{2,3} is too restrictive. For example http://www.test.info/ will return http://www.test.inf with a missing final 'o'.

  hm, our current regex needs to be modified 👀 

> I also wonder if we should also convert www.test.info without the protocol at the start. What do you think?

 Yes indeed, we should also convert URLs without the protocol at the beginning.  
  will make the changes soon, thanks for the review:)

MonkeyDo (changes requested):
So I should probably have realized earlier, but we're going to want to make this code more reusable by extracting it into a utility function (probably in src/client/helpers/utils.ts).

For example on any page with revision notes (e.g. revision page: https://test.bookbrainz.org/revision/29833), as well as annotations (https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/annotation.js#L49).
In the future there will be more uses for it but let's start with those.


The utility should return the whole <span> element with the `dangerouslySetInnerHTML`.




Separately to that, something else we missed is that a short url (`www.xyz…`) will be parsed correctly, but the url is interpreted relative to the current page.
Here's an example where on the page `http://localhost:9099/editor/1245/revisions`, the url `www.bookbrainz.org` will link to `http://localhost:9099/editor/1245/revisions/www.bookbrainz.org`:
Capture d’écran 2021-04-06 à 18 58 36
If I manually turn the URL from `www.bookbrainz.org` to `https://www.bookbrainz.org` (or even the less descriptive `//www.bookbrainz.org`), it points to the right page.
Full URLs are working fine however :+1:

akashgp09:
> So I should probably have realized earlier, but we're going to want to make this code more reusable by extracting it into a utility function (probably in src/client/helpers/utils.ts).
> The utility should return the whole `<span>` element with the `dangerouslySetInnerHTML`.

Then I guess we will need to rename our **utils.ts** file to **utils.tsx** so that we can return **jsx** `<span dangerouslySetInnerHTML={{...}} />`

MonkeyDo:
> Then I guess we will need to rename our **utils.ts** file to **utils.tsx** so that we can return **jsx** `<span dangerouslySetInnerHTML={{...}} />`

Yes, that's fine. There are no ill effects with doing that.

akashgp09:
> If I manually turn the URL from www.bookbrainz.org to https://www.bookbrainz.org (or even the less descriptive //www.bookbrainz.org), it points to the right page.

Added `addHttpRegex` which will prepend `https://` protocol before` www.` if there isn't any.

src/client/components/pages/parts/revisions-table.js
```diff
@@ -26,12 +26,13 @@ import {faCodeBranch} from '@fortawesome/free-solid-svg-icons';
...
 const {Table} = bootstrap;
-const {formatDate} = utilsHelper;
+const {formatDate, stringToUrl} = utilsHelper;
 
 
 function RevisionsTable(props) {
 	const {results, showEntities, showRevisionNote, showRevisionEditor, tableHeading} = props;
-
+	// eslint-disable-next-line max-len
+	const urlRegex = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/g;
```
MonkeyDo, line 35: This line looks unused now that you've refactored.

src/client/helpers/utils.tsx
```diff
@@ -161,3 +163,17 @@ export function dateObjectToISOString(value: DateObject) {
 	}
 	return date;
 }
+
+export function stringToUrl(string: string) {
```
MonkeyDo, line 167: I think the name could be a bit more descriptive, maybe something along those lines ?
```suggestion
export function stringToHTMLWithLinks(string: string) {
```

MonkeyDo (changes requested):
Looking good ! 
Last bits before merging this useful tool.

PR #577, opened 2021-03-17
Adding mailing service to the codebase

endurance21:
### Problem
There was no mailing agent dedicated for the codebase.

### Solution
Implement one such agent.
This pr deals with setting up node-mailer as our mailing agent.

### Areas of Impact
server
global config file


src/server/helpers/mailer.js
```diff
@@ -0,0 +1,12 @@
+import nodemailer from 'nodemailer';
+import {mailConfig} '../../../config/config.json';
+
+
+async function sendEmail({from, to, subject, html}) {
+	const transporter = nodemailer.createTransport(mailConfig);
+	// returns a promise
+	return transporter.sendMail({from, to, subject, html});
```
akashgp09, line 8: nice work here, just leaving some suggestions.
as we are usng  async function we should  use the await keyword here
```suggestion
	return await transporter.sendMail({from, to, subject, html});
```
 Also the sendMail method accepts a callback function to run once the message is delivered or it failed. 
```suggestion
	return transporter.sendMail({from, to, subject, html}, (error)=>{
	  if (error) {
          return console.log(error);
      }
	}
	);
```
This callback will help us know whenever something goes wrong wilth sufficient information

akashgp09 (changes requested):

src/server/helpers/mailer.js
akashgp09: also your checks are failing because it is unable to parse `{from, to, subject, html}`. then till then you can fill it with some dummy datas to pass the checks
 {
      from: '"Nodemailer Contact" <your@email.com>',
      to: 'RECEIVEREMAILS', 
      subject: 'Node Contact Request', 
      html: 'output '
  }

endurance21: hey ! @akashgp09  thanks for your review !
first for the async part , i think it not necessary to use `await` here, async definitions are just meant to declare that this fucntion is aynchronous in nature , and returning a promise from such functions means the same.
when this sendEmail() will be invoked then we can use `await` to to see it that is resolved or not.
returning simple promise abstracts erro handlling to the invokation place and this way we get clear API .

secondly ,i dont think we need to add `error` callback function, as all those error will ultimately be cathed in promise object !

third part , i was wondering equivalent as lint error information is not adequate to trace the error, but  let me try this sugestion though!

endurance21: may be this is what i am intending ! 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#return_value
@MonkeyDo  what are your thoughts on this

MonkeyDo: Returning a promise from an async function is fine, as you mention it's equivalent to `await`ing it. Per the nodemailer docs:
>If callback argument is not set then the method returns a Promise object.

We'll end up using `sendMail` as a promise.
We can then proceed normally and wrap it in a try-catch block if we use it with async/await syntax, or with `.catch` if we're using promise syntax.

Let's start with that, and depending on how we use it we may decide to instead use a callback to be able to log error here.

src/server/helpers/mailer.js
```diff
@@ -0,0 +1,12 @@
+import nodemailer from 'nodemailer';
+import {mailConfig} '../../../config/config.json';
```
endurance21, line 2: i think the problem is here , there should be a `from` keyword

endurance21:
@MonkeyDo  lint tests are failing , and one of them is due to 
` Async function 'sendEmail' has no 'await' expression ` 
should i bypass this?
or should i remove async keyword cz anyway this function gonna return promise :thinking:

MonkeyDo:
> @MonkeyDo lint tests are failing , and one of them is due to
> `Async function 'sendEmail' has no 'await' expression`
> should i bypass this?
> or should i remove async keyword cz anyway this function gonna return promise 🤔

Removing the async keyword is the way to go here.

MonkeyDo (commented):
You'll also need to commit the package-lock.json file in order for the tests to pass.

src/server/helpers/mailer.js
```diff
@@ -0,0 +1,14 @@
+import config from '../../../config/config.json';
```
MonkeyDo, line 1: See how we import and deal with the configuration here: https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/app.js#L32

endurance21: i thougt mail configs must be at the global level , that way wont it be easy to setup the project for newbies ?

endurance21: okay i get it

endurance21:
done the changes , please have a look @MonkeyDo

endurance21:
i have run npm run lint locally but found no error ! 
how this test is not passing

src/server/helpers/mailer.js
```diff
@@ -0,0 +1,14 @@
+import config from '../../common/helpers/config.js';
+import nodemailer from 'nodemailer';
+
+
+const {mailConfig} = config;
+
+function sendEmail({from, html, subject, to}) {
```
MonkeyDo, line 7: Let's do a few things to improve this function:
1. Change the file extension to `.ts` so that we can add typescript types
2. Take each of the inputs as a separate argument, rather than requiring to know the expected structure of the object
3. Add types for each argument

We'll have something like this: `function sendEmail(from:string, html:string|JSX.Element, subject:string, to:string) {`

And finally 4. Adding JSDoc style documentation like you'll see [here](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/api/helpers/entityLoader.ts#L23-L44) for example

That way when we import the function in another typescript file we'll get full completion, type checking and documentation !

MonkeyDo (changes requested):
Not sure what's happening with the linter, but we'll see if the next push passes. If not we'll ignore them as the dependency is indeed in package.json

src/server/helpers/mailer.js
endurance21: sorry for the delay , actually i am in midle of my semester exams ;)

endurance21: i will be making the changes soon

endurance21:
@MonkeyDo  
i have done the chnages , looking forward for the process :)

src/server/helpers/mailer.ts
```diff
@@ -0,0 +1,66 @@
...
+const {mailConfig} = config;
+
+/**
+ * This is a dedicated mailing interface for bookbrainz mailing services
+ *
+ * @param {string} from - Email address of sender
+ * @param {string} to - Email address of receiver 
+ * @param {string} subject- Subject of the E-mail
+ * @param {string} html - Body of E-mail in form of HTML 
+ * @returns {object} an object containing the error message if any error will occur.
```
MonkeyDo, line 30: The function returns a promise, not an object.
However it is useful to also detail what that promise returns, as you've done.
```suggestion
 * @returns {Promise<object>} Returns a promise that resolves with an error object if message failed or an info object if the message succeeded (see https://nodemailer.com/usage/ for more details)
```

src/server/helpers/mailer.ts
```diff
@@ -0,0 +1,66 @@
...
+ * If entity is found successfully in the database this function set the entity data
+ * at res.locals.entity and return to next function.
+ * @example
+ * Promise way :
+ * 
+ * sendEmail("abc@gmail.com", "def@gmail.com", "Revision added to Entity", "some html string")
+ * .then(()=>{
```
MonkeyDo, line 37:
```suggestion
 * .then((response)=>{
     // the response object contains details of a successful response (see https://nodemailer.com/usage)
```
That could be useful to someone in the future

src/server/helpers/mailer.ts
```diff
@@ -0,0 +1,66 @@
...
+* 	  console.log("Email has been sent successfully")
+ * 	 })
+ * .catch((error)=>{
+ * 	  //handle error
+ * 	})
+ * 
+ * Async/Await way :
+ * 
+ *  try{
+ * 		await sendEmail("abc@gmail.com", "def@gmail.com", "Revision added to Entity", "some html string")
```
MonkeyDo, line 47: Similarly:
```suggestion
 * 		const response = await sendEmail("abc@gmail.com", "def@gmail.com", "Revision added to Entity", "some html string")
```

src/server/helpers/mailer.ts
```diff
@@ -0,0 +1,66 @@
...
+ * 		console.log("Email has been sent successfully")
+ *   }
+ * 	catch (error) {
+ * 		//handle the error
+ * 	}
+ * @description
+ * This helper function  sends E-mail with attributes collected from argument in given order. 
+ * It is basically using nodemailer's transporter to send the E-mail under the hood.
```
MonkeyDo, line 55:
```suggestion
 * It uses nodemailer's transporter to send the E-mail under the hood.
```

src/server/helpers/mailer.ts
```diff
@@ -0,0 +1,66 @@
...
+ * It returns a promise that should be handled using .then()/.catch() or async/await syntax.
```
MonkeyDo, line 56:
```suggestion
 * It returns a promise that should be handled using .then()/.catch() or async/await syntax wrapped in a try…catch block.
```

MonkeyDo (changes requested):
Looking pretty good !

The file mailer.js needs to be removed, now that you've created the .ts file.

endurance21:
Have to lint it properly once !

endurance21:
@MonkeyDo  i think this one is ready to go !

MonkeyDo (approved):
Great, thanks !

PR #581, opened 2021-03-22
FIX(BB-593): Overlapping buttons

akashgp09:

### Problem

This PR Fixes: [BB-593](https://tickets.metabrainz.org/projects/BB/issues/BB-593?filter=allopenissues)

### Solution

Added margin to buttons

### Before

Screenshot 2021-03-22 211501

### After

Screenshot 2021-03-22 212628

### Areas of Impact

[collections-table.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/collections-table.js)


MonkeyDo (approved):
That's good to go, thanks ! :)

PR #583, opened 2021-03-29
fix(profile): Fix bug where profile page does not load due to undefined area (BB-600)

BharatKalluri:
### Problem

Fix bug where profile page does not load due to undefined area

### Solution

Added undefined check on area

### DIFF
--- a/src/client/components/forms/profile.js
+++ b/src/client/components/forms/profile.js
@@ -105,7 +105,7 @@ class ProfileForm extends React.Component {
 	hasChanged = () => {
 		const {name, bio, areaId, titleId, genderId} = this.state;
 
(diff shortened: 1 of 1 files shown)


src/client/components/forms/profile.js
```diff
@@ -105,7 +105,7 @@ class ProfileForm extends React.Component {
 	hasChanged = () => {
 		const {name, bio, areaId, titleId, genderId} = this.state;
 
-		return this.props.editor?.area.id !== areaId ||
+		return this.props.editor?.area?.id !== areaId ||
```
endurance21, line 108: :+1:

MonkeyDo (approved):
Thanks, good catch and good simple fix !

PR #586, opened 2021-03-30
fix: Removed trailing space in getEntityLabel (BB-601)

BharatKalluri:
### Problem

Remove trailing space so that selections will not contain a space at the end

### Solution

Currently `getLabelEntity` adds a space at the end. Looks like Label is always followed by disambiguation and `getEntityDisambiguation` already [prefixes a space](https://github.com/bookbrainz/bookbrainz-site/blob/c908c2135aaed62c7448977a74ece5c6fd462524/src/client/helpers/entity.tsx#L217). So this should not be a problem :)

### Areas of Impact

Text selection in title will be easier

### DIFF
--- a/src/client/helpers/entity.tsx
+++ b/src/client/helpers/entity.tsx
@@ -165,7 +165,7 @@ export function entityToOption(entity) {
 
 export function getEntityLabel(entity, returnHTML = true) {
 	if (entity.defaultAlias) {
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
I had a look in the codebase, and it looks like indeed this trailing space can be removed without ill effects, and fixes the tickets.
Thanks !

PR #587, opened 2021-03-31
Refactor(achievement): Rewritten Achievements helpers with async/await syntax

akashgp09:
### Problem

This PR Fixes: [BB-598](https://tickets.metabrainz.org/projects/BB/issues/BB-598?filter=allissues)

### Solution

 A lot of the code uses promises in the `promise.then(…)` format, which makes the code hard to read and maintain. Rewritten it using the **async/await** syntax

### Areas of Impact

 file [src/server/helpers/achievement.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/helpers/achievement.js)

### DIFF
--- a/src/server/helpers/achievement.js
+++ b/src/server/helpers/achievement.js
@@ -40,11 +40,11 @@ async function awardUnlock(UnlockType, awardAttribs) {
 		if (award === null || typeof award === 'undefined') {
 			const unlock = await new UnlockType(awardAttribs).save(null, {method: 'insert'});
 			return unlock.toJSON();
-		}	
-		return new Promise((resolve) => resolve('Already unlocked'));	
-	} catch (err) {
-		return new Promise((resolve, reject) => reject(err))
-		
+		}
+		return new Promise((resolve) => resolve('Already unlocked'));
+	}
+	catch (err) {
+		return new Promise((resolve, reject) => reject(err));
 	}
 }
 
@@ -59,7 +59,7 @@ async function awardUnlock(UnlockType, awardAttribs) {
  */
 async function awardAchievement(orm, editorId, achievementName) {
 	const {AchievementType, AchievementUnlock} = orm;
-	const achievementTier= await new AchievementType({name: achievementName})
+	const achievementTier = await new AchievementType({name: achievementName})
 		.fetch({require: false});
 	let awardPromise;
 	if (achievementTier === null) {
@@ -74,13 +74,14 @@ async function awardAchievement(orm, editorId, achievementName) {
 				achievementId: achievementTier.id,
 				editorId
 			};
-			const unlock = await awardUnlock(AchievementUnlock, achievementAttribs);						
+			const unlock = await awardUnlock(AchievementUnlock, achievementAttribs);
 			const out = {};
 			out[achievementName] = unlock;
 			awardPromise = out;
-		} catch (err) {
-			return new Promise((resolve, reject) => reject(new error.AwardNotUnlockedError(err.message)))
-		}			
+		}
+		catch (err) {
+			return new Promise((resolve, reject) => reject(new error.AwardNotUnlockedError(err.message)));
+		}
 	}
 	return awardPromise;
 }
@@ -97,7 +98,7 @@ async function awardAchievement(orm, editorId, achievementName) {
 async function awardTitle(orm, editorId, tier) {
 	const {TitleType, TitleUnlock} = orm;
 	let titlePromise;
-	if (tier.titleName) { 
+	if (tier.titleName) {
 		const title = await new TitleType({title: tier.titleName})
 			.fetch({require: false});
 		let awardPromise;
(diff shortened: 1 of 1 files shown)


src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
-				new Promise((resolve, reject) => reject(new Error('no date attribute')));
-			}
-			return differencePromise;
-		})
-		.catch(() => new Promise((resolve, reject) => reject(new Error('no date attribute'))));
+	try {
+		const edition = await new EditionRevision({id: revisionId}).fetch();
+		const data = await edition.related('data').fetch();
+		const releaseEventSet = await data.related('releaseEventSet').fetch();
+		const releaseEvents = await releaseEventSet.related('releaseEvents').fetch();
```
MonkeyDo, line 463: I feel like we should be able to replace this with the right ORM relationships.
I wonder if 
```suggestion
		const edition = await new EditionRevision({id: revisionId}).fetch({withRelated: 'data.releaseEventSet.releaseEvents'});
		const releaseEvents = edition.related('releaseEventSet').related('releaseEvents');
```
Or something along those lines should work. I'm not sure about the syntax for the second line (I'm not 100% certain we can chain the calls like that). 

If all else fails you can use `edition.toJSON()` and then access `edition.releaseEventSet.releaseEvents`.
In that case there will be differences in the code below too; `releaseEvents.models[0].attributes`  will probably be something like `releaseEvents[0]`.

src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
+		if (releaseEvents.length > 0) {
+			const msInOneDay = 86400000;
+			const attribs = releaseEvents.models[0].attributes;
+			const date = new Date(
+				attribs.year,
+				attribs.month - 1,
+				attribs.day
+			);
+			const now = new Date(Date.now());
+			differencePromise = Math.round((date.getTime() - now.getTime()) / msInOneDay);
```
MonkeyDo, line 474: While we're in there touching the code, this can be improved:
```suggestion
			const now = Date.now();
			return Math.round((date.getTime() - now) / msInOneDay);
```

src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
+		}
+		else {
+			differencePromise =
+		new Promise((resolve, reject) => reject(new Error('no date attribute')));
```
MonkeyDo, line 478: As in other comments, you can just `throw` directly here instead of returning a Promise.

src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
+		}
+		return differencePromise;
```
MonkeyDo, line 480: With the above two comments this variable won't be needed anymore

src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
+	}
+	catch (err) {
+		return new Promise((resolve, reject) => reject(new Error('no date attribute')));
```
MonkeyDo, line 483:
```suggestion
		throw new Error('no date attribute');
```

src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
-			}
-			else {
-				achievementPromise = new Promise(resolve => resolve(
-					{'Hot Off the Press': false}
-				));
-			}
-			return achievementPromise;
-		})
-		.catch((err) => ({'Hot Off the Press': err}));
+			achievementPromise = testTiers(orm, diff, editorId, tiers);
```
MonkeyDo, line 511:
```suggestion
			return testTiers(orm, diff, editorId, tiers);
```

src/server/helpers/achievement.js
```diff
@@ -474,71 +454,72 @@ function achievementToUnlockId(achievementUnlock) {
...
+		}
+		else {
+			achievementPromise = new Promise(resolve => resolve(
+				{'Hot Off the Press': false}
+			));
```
MonkeyDo, line 516:
```suggestion
			return {'Hot Off the Press': false};
```

MonkeyDo (changes requested):
Most fo the comments below are repeated, as the code itself is repetitive.
Overall that's a good conversion !

src/server/helpers/achievement.js
akashgp09: > Or something along those lines should work. I'm not sure about the syntax for the second line (I'm not 100% certain we can chain the calls like that).

Chaining isn't working, though the second approach works perfect :)

src/server/helpers/achievement.js
```diff
@@ -34,18 +34,18 @@ import log from 'log';
...
-		})
-		.catch(err => new Promise((resolve, reject) => reject(err)));
+async function awardUnlock(UnlockType, awardAttribs) {
+	try {
+		const award = await new UnlockType(awardAttribs).fetch({require: false});
+		if (award === null || typeof award === 'undefined') {
+			const unlock = await new UnlockType(awardAttribs).save(null, {method: 'insert'});
+			return unlock.toJSON();
+		}
+		return new Promise((resolve) => resolve('Already unlocked'));
```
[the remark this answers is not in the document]
akashgp09: > On the other side in the catch block you can simply throw err

I have removed the try catch statement as the catch clause was only rethrowing the original error, which is redundant.

akashgp09:
@MonkeyDo thank you so much for the detailed review :100: . I have refactored the code as suggested, do let me know if changes are needed :)

MonkeyDo (approved):
Thanks a bunch, the code is a lot more readable now ! :)

PR #590, opened 2021-04-01
Autocomplete search limit

nynaalekhya:
Closes BB-577
### Problem

Search autocomplete endpoint responses are limited to 10 results

### Solution

Changed the search autocomplete endpoint limit by including  size property in dslQuery object body

### Areas of Impact

src/common/helpers/search.js

### DIFF
--- a/src/common/helpers/search.js
+++ b/src/common/helpers/search.js
@@ -202,7 +202,8 @@ export function autocomplete(orm, query, type) {
 
 	const dslQuery = {
 		body: {
-			query: queryBody
+			query: queryBody,
+		    size: 10000
 		},
 		index: _index
 	};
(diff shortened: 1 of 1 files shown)


src/common/helpers/search.js
```diff
@@ -202,7 +202,8 @@ export function autocomplete(orm, query, type) {
 
 	const dslQuery = {
 		body: {
-			query: queryBody
+			query: queryBody,
+			size: 10000
```
MonkeyDo, line 206: That looks like a pretty extreme number… I'm not sure what the implications would be in terms of time to run the queries, but let's not overdo it and keep that number  more reasonable.
After all, this autocomplete endpoint sis aimed at humans using the website, and it's unlikely that a human would go through a thousand autocomplete suggestions :)
Let's go with… 42?

nynaalekhya: I have set it to the max limit allowed by the API
Changed the size to 42.

MonkeyDo (approved):
Thank you !

PR #591, opened 2021-04-01
fixes badges visibility in editor's achievements tab

nynaalekhya:
Closes BB-605
### Problem

Unlocked and Locked achievements are not visible on small screen devices
achievements

### Solution

fixes badges visibility in editor's achievements tab
badge

### Areas of Impact

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -563,3 +563,8 @@ hr.wide {
     }
 }
 
+/* editor-achievements.js */
+
+.dnd-container {
+	display: flex;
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
A simple change that makes the page usable, which is a lot better than before :)
Thank you !

Longer term, we'll probably want to redesign that page a bit, as even with this improvement it's not very usable on mobile devices.

nynaalekhya:
@MonkeyDo Thank you, I would be glad to help you with the redesign.

PR #592, opened 2021-04-01
added my collections button

nynaalekhya:
Closes BB-596

### Problem

Added My collections button

### Solution

Public collections page
mycollection1

Viewing other collections
mycollection2

Clicking on my collections
collection3

### DIFF
--- a/src/client/components/pages/parts/collections-table.js
+++ b/src/client/components/pages/parts/collections-table.js
@@ -107,7 +107,7 @@ class CollectionsTable extends React.Component {
 		// 1.the user is logged in and not viewing the user's collections
 		// 2.the user is logged in and viewing public collections
 		const showMyCollectionButton = user && (user.id !== ownerId || !ownerId);
-		
+
 		return (
 			<div>
 				<div>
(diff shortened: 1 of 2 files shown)


src/client/components/pages/parts/collections-table.js
```diff
@@ -91,13 +91,32 @@ class CollectionsTable extends React.Component {
...
 		(user && !ownerId);
+
+		const myCollectionButton = (
+			<Button
+				bsStyle="success"
+				className="margin-bottom-d5"
+				href={`/editor/${user.id}/collections`}
+				type="button"
+			>
+				&nbsp;My Collections
```
MonkeyDo, line 102:
```suggestion
				My Collections
```

Since there's no icon before the text we don't need that space

MonkeyDo (changes requested):
Thanks for this addition, that looks good !
A tiny little change and we can merge the PR.

src/client/components/pages/parts/collections-table.js
nynaalekhya: I have made the suggested changes.

MonkeyDo (approved):
Thank you for this good addition !

PR #593, opened 2021-04-03
FIX(BB-606): Buttons overlapping in merge entity section

akashgp09:
### Problem

**This PR Fixes: [BB-606](https://tickets.metabrainz.org/projects/BB/issues/BB-606?filter=allissues)**

### Solution

**Added some margins to buttons** 

### Before

Screenshot from 2021-04-03 14-33-15

### After
Screenshot from 2021-04-03 14-31-29

### Areas of Impact

[merge-queue.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/merge-queue.js)

### DIFF
--- a/src/client/components/pages/parts/merge-queue.js
+++ b/src/client/components/pages/parts/merge-queue.js
@@ -98,6 +98,7 @@ class MergeQueue extends React.Component {
 				<ButtonGroup>
 					<Button
 						bsStyle="success"
+						className="margin-bottom-d5"
 						disabled={isNil(this.state.selectedOption)}
 						href={`/merge/submit/${this.state.selectedOption}`}
 						title="Merge entities"
@@ -107,6 +108,7 @@ class MergeQueue extends React.Component {
 					</Button>
 					<Button
 						bsStyle="warning"
+						className="margin-bottom-d5"
 						disabled={isNil(this.state.selectedOption)}
 						href={`/merge/remove/${this.state.selectedOption}`}
 						title="Remove from merge"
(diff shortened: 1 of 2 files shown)


src/client/components/pages/parts/merge-queue.js
```diff
@@ -98,6 +98,7 @@ class MergeQueue extends React.Component {
 				<ButtonGroup>
 					<Button
 						bsStyle="success"
+						className="margin-bottom-d5"
```
endurance21, line 101: I think if we dont assign margin on first element that would be fine ?

akashgp09: I don't think so that would be good :)
If we remove margin from the first element then this is how it's gonna look

Screenshot from 2021-04-05 12-33-19

endurance21: Oh ! I am sorry I meant that for the last button .

Is it necessary to give margin to last button ?

MonkeyDo (commented):
Since this seems to be an issue in various places, maybe it would be better to try and solve it at a higher level with css.
Something like this should be equivalent:
```
.btn-group > .btn:not(:first-child) {
    margin-bottom: 0.5em;
}
```
What do you think?

akashgp09:
> What do you think?

Yes Indeed, It would be a great solution fixing this overlapping buttons issue all at once.

MonkeyDo (approved):
Great, thanks :)

PR #596, opened 2021-04-05
Improve[UI]: added left margin to buttons on collection page for better UI

endurance21:
### Problem
Buttons on collection page are collapsing or touching each other'a boundary,  which not only looks less better but also may temp user to press wrong button as no boundary space is given. 
Screenshot from 2021-04-05 07-23-03

### Solution

Screenshot from 2021-04-05 07-22-15

### Areas of Impact

src/client/components/pages/collection.js

### DIFF
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -270,7 +270,7 @@ class CollectionPage extends React.Component {
 							<Button
 								bsSize="small"
 								bsStyle="success"
-								className="margin-bottom-d5"
+								className="margin-bottom-d5 margin-left-d5"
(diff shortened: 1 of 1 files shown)


endurance21:
@MonkeyDo  please review this pr

MonkeyDo (commented):
This sort of changes is better done on the scale of the whole website rather than separately on each page. Currently that collection page button bar will look different from the other pages button bars.
Instead you'll want to come up with a css solution.

I'm not completely convinced this is needed, but I could see how button without space could possibly be detrimental. I would think that the color difference would be enough?
Do you have any material I could read in  support of the idea?

PR #601, opened 2021-04-15
FIX(BB-615): Copy/Paste annotation text  in FireFox <= 60

akashgp09:

### Problem

**This PR Fixes: [BB-615](https://tickets.metabrainz.org/projects/BB/issues/BB-615?filter=allissues)**

### Solution

**Replaced the current `<p class="annotation-content">` with a `<pre class="annotation-content">` and overridden the existing `<pre>` css styles when className `annotation-content` is present.**

**Also Tested in some older version of Firefox browsers using BrowserStack.** 

### Areas of Impact

[annotation.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/annotation.js)
[style.less](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/stylesheets/style.less)

### DIFF
--- a/src/client/components/pages/entities/annotation.js
+++ b/src/client/components/pages/entities/annotation.js
@@ -46,7 +46,7 @@ class EntityAnnotation extends React.Component {
 				<Col md={12}>
 					<h2>Annotation</h2>
 					<Collapse in={this.state.open}>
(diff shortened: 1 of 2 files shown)


MonkeyDo (approved):
Well done, that's a great solution !

PR #602, opened 2021-04-18
FIX(Revision): "New" badge appears in entity deletion revision

akashgp09:
### Problem

**This PR Fixes: [BB-523](https://tickets.metabrainz.org/projects/BB/issues/BB-523?filter=allopenissues)**

### Solution

Refactored the `isNew` property logic in **server/routes/revision.js** by taking advantage of `dataId`( which is **null** for a deleted entity)
 
### Before

Screenshot from 2021-04-18 21-39-58

### After
Screenshot from 2021-04-18 21-39-19

### Areas of Impact

server/routes/revision.js

### DIFF
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -171,8 +171,10 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
 					(parent) => {
 						const dataId = revision.get('dataId');
 						let isNew = false;
+						let isDeletion = false;
 						if (!parent) {
 							isNew = dataId;
+							isDeletion = !dataId;
 						}
 						return makePromiseFromObject({
 							changes: revision.diff(parent),
@@ -183,6 +185,7 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
 									orm, entityType, revision.get('bbid')
 								),
 							isNew,
+							isDeletion,
 							revision
 						});
 					},
@@ -196,6 +199,7 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
 								orm, entityType, revision.get('bbid')
 							),
 						isNew: true,
+						isDeletion,
 						revision
 					})
 				)
@@ -283,7 +287,7 @@ router.get('/:id', async (req, res, next) => {
 			revision: revision.toJSON(),
 			title: 'RevisionPage'
 		});
-
+		
 		const markup = ReactDOMServer.renderToString(
 			<Layout {...propHelpers.extractLayoutProps(props)}>
 				<RevisionPage
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -173,7 +173,7 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
 						let isNew = false;
 						let isDeletion = false;
 						if (!parent) {
-							isNew = dataId;
+							isNew = Boolean(dataId);
 							isDeletion = !dataId;
 						}
 						return makePromiseFromObject({
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -169,22 +169,23 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
 			revision.parent()
 				.then(
 					(parent) => {
-						let dataId = revision.get('dataId');
+						const dataId = revision.get('dataId');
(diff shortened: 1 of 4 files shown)


src/server/routes/revision.js
```diff
@@ -168,17 +168,24 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
...
+					(parent) => {
+						const dataId = revision.get('dataId');
+						let isNew = false;
+						if (!parent) {
+							isNew = dataId;
+						}
+						return makePromiseFromObject({
+							changes: revision.diff(parent),
+							entity: revision.related('entity'),
+							entityAlias: revision.get('dataId') ?
```
MonkeyDo, line 180: Since we've already extracted this as a constant:

```suggestion
							entityAlias: dataId ?
```

MonkeyDo (commented):
Great, that's a good improvement!

I think it would be great to take that opportunity to mark deleted entities with a `Deleted` badge with a red background, now that you've got that piece of info in your logic.
I guess the cleanest would be to pass a new `isDeletion` property.

What do you think?

src/server/routes/revision.js
akashgp09: > Great, that's a good improvement!
> 
> I think it would be great to take that opportunity to mark deleted entities with a `Deleted` badge with a red background, now that you've got that piece of info in your logic.
> I guess the cleanest would be to pass a new `isDeletion` property.
> 
> What do you think?

akashgp09:
Added the deleted badge to mark deleted entitites. Let me know if any changes are required :)

Screenshot from 2021-04-19 19-08-31

src/server/routes/revision.js
```diff
@@ -168,17 +168,27 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
...
-							),
-						isNew: !parent,
-						revision
-					}),
+					(parent) => {
+						const dataId = revision.get('dataId');
+						let isNew = false;
+						let isDeletion = false;
+						if (!parent) {
+							isNew = dataId;
```
MonkeyDo, line 176:
```suggestion
							isNew = Boolean(dataId);
```

I guess this is a bit clearer and makes sure we use a boolean.

src/server/routes/revision.js
```diff
@@ -168,17 +168,27 @@ function diffRevisionsWithParents(orm, entityRevisions, entityType) {
...
+							isDeletion = !dataId;
```
MonkeyDo, line 177: I think this could be moved above to the variable definition:
`const isDeletion = !dataId;`

MonkeyDo (changes requested):
Yes, that's nice, thanks for adding that :)

MonkeyDo (commented):
OK, this is starting to look really nice!
I went and deployed this PR to test.bookbraine.org so that we can do some testing on a common space.
It's working pretty well and showing the right badges where expected most of the time, but I did find a couple of issues:

If the entity has been merged into another entity, it will display the "deleted" badge, because for merges we also remove the dataId of any merged entity (apart from the one we are merging into).
I think it should be relatively easy to make another "merged" badge (and perhaps that will entirely replace the current way we display merged entities in a revision. That's for another PR)
An example of that: https://test.bookbrainz.org/revision/29798

In another case, I deleted an entity that had some previous revisions, and it appears with the "new" badge instead of "deleted". Not sure what is happening here:
https://test.bookbrainz.org/revision/29849
I deleted the entity `6ef597ce-9f15-4c91-ae8d-83c18a0c6797` and nothing else in that revision and the database seems to reflect that correctly:
image

akashgp09:
> I think it should be relatively easy to make another "merged" badge (and perhaps that will entirely replace the current way we display merged entities in a revision. That's for another PR)

Ok sure will open a PR for a`merged`badge to mark merged entities once this PR gets merged.

> In another case, I deleted an entity that had some previous revisions, and it appears with the "new" badge instead of "deleted". Not sure what is happening here:
https://test.bookbrainz.org/revision/29849
I deleted the entity 6ef597ce-9f15-4c91-ae8d-83c18a0c6797 and nothing else in that revision and the database seems to reflect that correctly.

Yeah thanks for reporting it. I have made required changes and it seems to work fine now

Screenshot from 2021-04-20 08-37-29

MonkeyDo (approved):
All working fine now, thank you ! :)

PR #603, opened 2021-04-22
FIX(Revision): Alias language change not shown in revisions

akashgp09:
### Problem

This PR Fixes: [BB-592](https://tickets.metabrainz.org/browse/BB-592)

### Solution

- Used the **loadLanguages middleware** to get all the languagesOptions. 
- Converted the alias **languageId** to corresponding language **name** utilizing the languageOptions array of objects.
- Implemented an **isLanguage** property. If the revision includes change in **alias language** then **isLanguage** will be turned **true** and convert the alias **languageId** to corresponding language **name** before displaying.  

### Before

Screenshot from 2021-04-22 17-23-59

### After

Screenshot from 2021-04-22 08-53-41

### Areas of Impact

- [server/routes/revision.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/routes/revision.js)
- [server/helpers/diffFormatters/base.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/helpers/diffFormatters/base.js)
- [server/helpers/diffFormatters/entity.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/helpers/diffFormatters/entity.js)
- [client/components/pages/revision.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/revision.js)

### DIFF
--- a/src/client/components/pages/revision.js
+++ b/src/client/components/pages/revision.js
@@ -49,7 +49,15 @@ class RevisionPage extends React.Component {
 		);
 	}
 
-	static formatChange(change) {
+	static formatChange(change,languageOptions) {	
+		if(change.isLanguage){
+			if(change.lhs){
+				const lhs = languageOptions.find(language => language.id === change.lhs[0]);
+				change.lhs = [lhs.name];
+			}
+			const rhs = languageOptions.find(language => language.id === change.rhs[0]); 
+			change.rhs = [rhs.name];
+		}
 		const isChangeADate = change.key.toLowerCase().match(/\bdate\b/);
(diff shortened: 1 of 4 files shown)


MonkeyDo:
While reviewing this PR I realized that all the code is there to compare the alias language changes without the need to load all the languages from the DB.
That seems a bit overkill when part of the issue is solved with minimal changes in #646 .

I'm still seeing another issue with the revision diff linked in the ticket description, so it is not fully fixed.
However I think this PR can be closed as I believe we do not need to load the languages.

PR #613, opened 2021-04-29
feat(search) : show a work's author in inline search

akashgp09:
### Problem

Bookbrainz Ticket : [BB-528](https://tickets.metabrainz.org/browse/BB-528)

### Solution

Fetched all relationships related to **Work** entity then filtered the relationships array using `getRelationshipSourceByTypeId()` helper function .

### Before

Screenshot from 2021-04-29 20-02-36

### After
Screenshot from 2021-04-29 20-03-17

### Areas of Impact

- [common/helpers/search.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/common/helpers/search.js)
- [common/entity-search-field-option.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/common/entity-search-field-option.js)
- [common/entity.tsx](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/common/entity.tsx)
- [common/linked-entity.tsx](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/common/linked-entity.tsx)

### DIFF
--- a/src/client/entity-editor/common/entity-search-field-option.js
+++ b/src/client/entity-editor/common/entity-search-field-option.js
@@ -25,6 +25,7 @@ import React from 'react';
 import {Async as SelectAsync} from 'react-select';
 import ValidationLabel from '../common/validation-label';
 import _ from 'lodash';
+import {getRelationshipSource, getRelationshipSourceByTypeId} from '../../helpers/entity';
 import makeImmutable from './make-immutable';
 import request from 'superagent';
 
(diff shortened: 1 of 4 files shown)


MonkeyDo (changes requested):
I've been mulling over this, and I think we should try to somehow add the author's alias(es) as a searchable field when indexing the entities.
That way, we would also improve the quality of search results by matching works by author name (with a lower importance).
This would also work say with the API without needing any extra implementation.

Let me know if you need some help deciphering the search indexing. I think overall the logic you used to fetch a work's author can be mostly reused.

Adding `'relationships.source','relationships.type'` to [this line](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/common/helpers/search.js#L358) should save you having to fetch the relationshipSet manually, and then processing Editions to fetch and add the author like you did, somewhere in [_fetchEntityModelsForESResults](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/common/helpers/search.js#L36).

Let me know how it goes and if you need some help, the search system takes some getting used to !

tr1ten:
Hey @akashgp09 are you still working on this PR ? , will it okay if i work on this?

akashgp09:
Sure @tr1ten, I’m a little tied up with some other work at the moment, you can take this up.

PR #615, opened 2021-04-30
FIX: proptypes mismatch and invalid DOM property warning

akashgp09:

### Problem

Screenshot from 2021-04-30 20-05-49

### Solution

updated **propTypes** and fixed invalid DOM property warning

### Areas of Impact

- [components/entity-link.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/entity-link.js)
- [pages/revision.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/revision.js)


MonkeyDo (approved):
That's an easy merge, thank you !

PR #616, opened 2021-05-02
[BB-619]: Search results should not contain private collections

akashgp09:
### Problem

This PR Fixes: [BB-619](https://tickets.metabrainz.org/browse/BB-619)

### Solution

Filter out all the private collections before showing in the search results. 

### Areas of Impact

- [helpers/search.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/common/helpers/search.js)

### DIFF
--- a/src/client/components/pages/parts/search-results.js
+++ b/src/client/components/pages/parts/search-results.js
@@ -143,13 +143,7 @@ class SearchResults extends React.Component {
 	}
 
 	render() {
-		let resultsArray = this.props.results;
-		if (resultsArray.length) {
-			if (resultsArray[0].type === 'Collection') {
-				resultsArray = resultsArray.filter(result => result.public);
-			}
-		}
-		const noResults = !resultsArray || resultsArray.length === 0;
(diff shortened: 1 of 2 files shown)


MonkeyDo (changes requested):
Similarly to PR #599 we should make sure we do the filtering on the server side instead of the client-side.

I even wonder if we should index private collections. The simplest solution might be to filter out private collections when creating the index here: https://github.com/bookbrainz/bookbrainz-site/blob/master/src/common/helpers/search.js#L420-L421
Adding a `.where({public: true})` as described [in the BookshelfJS docs](https://bookshelfjs.org/api.html#Model-instance-where) should be enough by itself to solve the issue.
Let me know if that works out !

akashgp09:
> Adding a `.where({public: true})` as described in the BookshelfJS docs should be enough by itself to solve the issue.
Let me know if that works out !

Yeah !! quick and simple. Works great, thanks :)

MonkeyDo (approved):
Good to go! Thanks for fixing this !

PR #619, opened 2021-05-04
BB-454: Should not be able to create relationship with itself

akashgp09:
### Problem

This PR Fixes: [BB-454](https://tickets.metabrainz.org/browse/BB-454)

### Solution

Filtered the search field options using entity bbid

### Before
Screenshot from 2021-05-04 16-13-20

### After

Screenshot from 2021-05-04 15-56-01

### Areas of Impact

- [common/entity-search-field-option.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/common/entity-search-field-option.js)
- [relationship-editor.tsx](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/relationship-editor/relationship-editor.tsx)

### DIFF
--- a/src/client/entity-editor/common/entity-search-field-option.js
+++ b/src/client/entity-editor/common/entity-search-field-option.js
@@ -96,8 +96,9 @@ class EntitySearchFieldOption extends React.Component {
 				q: manipulatedQuery,
 				type: this.props.type
 			});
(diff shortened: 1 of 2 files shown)


MonkeyDo (approved):
:rocket:  Good fix, thanks !

PR #621, opened 2021-05-05
BB-432: move duplicate entity warning below sort name

akashgp09:

### Problem

This PR Fixes: [BB-432](https://tickets.metabrainz.org/browse/BB-432)

### Solution

Moved Duplicate Entity Warning below sort name

Screenshot from 2021-05-05 16-46-43

### Areas of Impact

- [name-section.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/entity-editor/name-section/name-section.js)

### DIFF
--- a/src/client/entity-editor/name-section/name-section.js
+++ b/src/client/entity-editor/name-section/name-section.js
@@ -171,7 +171,22 @@ class NameSection extends React.Component {
 							onChange={this.handleNameChange}
 						/>
 					</Col>
+				</Row>
+				<Row>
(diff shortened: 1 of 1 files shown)


MonkeyDo:
I think this need to be reconsidered more thoroughly.
If the possible duplicates are in the way in the name section, then moving them one field lower isn't really a solution.

We should think about how to use the space on the right of the name section and/or a popover to display the possible duplicates rather than just insert it between existing fields.

@akashgp09 do let me know if this is something you might be interested in working on, otherwise I will close the PR.

PR #622, opened 2021-05-06
fix(BB-405):Search: Tabulated results columns have dynamic width

the-good-boy:
### Problem

This PR deals with https://tickets.metabrainz.org/browse/BB-405

### Solution

Added some css to fix the layout of all tables, as this problem also occurs in other places where table is used.

BEFORE:
searchcolumn-initial

AFTER:
SearchColumnsImproved

### Areas of Impact

Modified: src/client/stylesheets/style.less

### DIFF
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -97,8 +97,7 @@ table {
 	table-layout: fixed;
 }
 table td {
-	text-overflow: ellipsis;
-	overflow: hidden;
+	word-break: break-word;
 }
 .whole-page-form {
 	margin-top: 3em;
--- a/src/client/stylesheets/style.less
+++ b/src/client/stylesheets/style.less
@@ -93,6 +93,13 @@ body {
 
 /* Site-wide styling */
 /* ================= */
+table {
+	table-layout: fixed;
+}
+table td {
+	text-overflow: ellipsis;
+	overflow: hidden;
+}
 .whole-page-form {
 	margin-top: 3em;
 }


MonkeyDo (commented):
Setting styles for the whole website like that requires to be very sure we want to affect the same changes on other pages.

I had a look at the front page table of recent revisions and saw this issue with the `table-layout` property:
image
Compare to the original with the table-layout rule disabled:
image

Seem to be doing the same on the revisions page when the window size is small.
I'm not sure that solution is going to work, but maybe there's another way, or something more to add so that lines don't overlap like that?

the-good-boy:
Whenever there is an overlong word in an alias, the column tries to shrink all other columns in order to fit the word. I have given an example of this on the screenshot above. Although, a word that big is unlikely to actually be used as an alias

Looking at the ticket, I am failing to reproduce the exact issue mentioned there. 
However, even if the name of an entity is not as big as the one I have shown, we can still run into this type of column-shrinking in other tables also. I tried to use the 'word-break' to deal with overlong words in columns.

I used a 'fixed' table-layout in order to deal with this dynamic shifting of column-sizes in tables:

Screenshot from 2021-05-22 12-16-10
Screenshot from 2021-05-22 12-16-16

I agree that the table doesn't look that bad in this case, but we can easily imagine cases where it can. 

About the overlap, I think doing `overlap: hidden` fixes the issue. Admittedly, I did not try changing the size of the window to check the changes before making the pr.

the-good-boy:
Hmm. Now that I think about it, some might even argue that column resizing is actually better than giving a fixed width to each column equally (like in this case, each column has been assigned a width of 25%): 
Screenshot from 2021-05-22 12-37-56

MonkeyDo:
> Hmm. Now that I think about it, some might even argue that column resizing is actually better than giving a fixed width to each column equally (like in this case, each column has been assigned a width of 25%):


On idea would be to use a `width` attribute with a percentage on the column headers like so:
image

MonkeyDo:
I think this will require another approach

PR #626, opened 2021-05-08
BB-585: Revisions should clearly mark removed entities

akashgp09:
### Problem

This PR Fixes: [BB-585](https://tickets.metabrainz.org/browse/BB-585)

### Solution

- utilized **entity dataId** to mark a deleted entity.
- deleted entity should have **parentAlias** and existing entity should have **defaultAlias**

### Before
Screenshot from 2021-05-08 23-05-10

### After

Screenshot from 2021-05-08 20-27-38

### Areas of Impact

- [routes/revision.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/routes/revision.js)
- [diffFormatters/entity.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/helpers/diffFormatters/entity.js)


MonkeyDo (commented):
This works well for the most part. Seems to display deleted entities correctly in all cases I checked.

I'm seeing a regression in the way merged entities are displayed though, showing the BBID of the entity they are merged *into* rather than their own BBID :
Compare this screenshot from bb.org:
image

… to this one running this PR locally:
image

A clue might lie in this comment in the deleted code in diffFormatters/entity.js : 
`For entities without data (deleted or merged), we use getEntityParentAlias instead which returns a JSON object`

MonkeyDo (commented):
I think this is almost ready. Deleted entities are shown as expected, so that part seems solved.

However I believe that now a revision where we create a new entity will appear like the entity was deleted:

Screenshot from this revision mentioned in the BB-585 ticket: http://localhost:9099/revision/3212
image

MonkeyDo (approved):
Thanks for this good feature !
Now it will be a lot clearer when an entity has been deleted, even looking at its history.

I took the liberty to make the modifications I suggested on IRC; I added hover text for the strikethrough names and for the "new", "deleted" and "merged" badges to help avoid possible confusion.

Hovering on the strikethrough name:
image

Hovering on the "new" badge:
image

I also implemented a tiny refactor I had suggested in my previous review.

PR #649, opened 2021-06-14
feat(series): Created series entity create/edit routes and display page

akashgp09:
### Summary

- Created Series cretae/edit routes
- Set up  series entity-editor 
- Set up  series display page
- Set up attributes in relationship-editor
- Created new middlewares specific to series entity
- Enabled sorting relationships in relationship-section for series-entity
- Created new action and reducer for sorting relationships
- Implemented relationship attribute saving mechanism

### DIFF
--- a/src/client/components/pages/entities/series.js
+++ b/src/client/components/pages/entities/series.js
@@ -73,7 +73,7 @@ function SeriesDisplayPage({entity, identifierTypes, user}) {
 	const EntityTable = getEntityTable(entity.entityType);
 	const entityKey = getEntityKey(entity.entityType);
 	const propsForTable = {
-		[entityKey]: entity.entities,
+		[entityKey]: entity.seriesItems,
 		showAdd: false,
 		showAddedAtColumn: false,
 		showCheckboxes: false
--- a/src/client/components/pages/entities/series.js
+++ b/src/client/components/pages/entities/series.js
@@ -58,6 +58,12 @@ function SeriesAttributes({series}) {
 						<dd>{series.seriesOrderingType.label}</dd>
 					</dl>
 				</Col>
+				<Col md={3}>
+					<dl>
+						<dt>Total Items</dt>
+						<dd>{series.seriesItems.length}</dd>
+					</dl>
+				</Col>
 			</Row>
 		</div>
 	);
--- a/src/client/components/pages/entities/series.js
+++ b/src/client/components/pages/entities/series.js
@@ -18,6 +18,7 @@
 
 import * as bootstrap from 'react-bootstrap';
 import * as entityHelper from '../../../helpers/entity';
+import {formatDate, getEntityKey, getEntityTable} from '../../../helpers/utils';
 import EntityAnnotation from './annotation';
 import EntityFooter from './footer';
 import EntityImage from './image';
@@ -26,7 +27,7 @@ import EntityRelatedCollections from './related-collections';
 import EntityTitle from './title';
 import PropTypes from 'prop-types';
 import React from 'react';
-import {formatDate, getEntityTable, getEntityKey} from '../../../helpers/utils';
+
 
 const {deletedEntityMessage, getTypeAttribute, getEntityUrl, ENTITY_TYPE_ICONS, getSortNameOfDefaultAlias} = entityHelper;
 const {Col, Row} = bootstrap;
@@ -51,7 +52,7 @@ function SeriesAttributes({series}) {
 						<dd>{series.entityType}</dd>
 					</dl>
 				</Col>
-				<Col md={3}> 
+				<Col md={3}>
 					<dl>
 						<dt>Ordering Type</dt>
 						<dd>{series.seriesOrderingType.label}</dd>
@@ -93,7 +94,7 @@ function SeriesDisplayPage({entity, identifierTypes, user}) {
 				</Col>
 			</Row>
 			<EntityAnnotation entity={entity}/>
-		
+
 			{!entity.deleted &&
 			<React.Fragment>
 				<EntityTable {...propsForTable}/>
--- a/src/client/components/pages/entities/series.js
+++ b/src/client/components/pages/entities/series.js
@@ -18,7 +18,7 @@
 
 import * as bootstrap from 'react-bootstrap';
 import * as entityHelper from '../../../helpers/entity';
-import {formatDate, getEntityKey, getEntityTable} from '../../../helpers/utils';
+import {getEntityKey, getEntityTable} from '../../../helpers/utils';
 import EntityAnnotation from './annotation';
 import EntityFooter from './footer';
 import EntityImage from './image';
@@ -29,7 +29,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 
-const {deletedEntityMessage, getTypeAttribute, getEntityUrl, ENTITY_TYPE_ICONS, getSortNameOfDefaultAlias} = entityHelper;
+const {deletedEntityMessage, getEntityUrl, ENTITY_TYPE_ICONS, getSortNameOfDefaultAlias} = entityHelper;
 const {Col, Row} = bootstrap;
 
 function SeriesAttributes({series}) {
--- a/src/client/components/pages/entities/series.js
+++ b/src/client/components/pages/entities/series.js
@@ -1,0 +1,129 @@
+/*
+ * Copyright (C) 2021  Akash Gupta
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import * as entityHelper from '../../../helpers/entity';
+import EntityAnnotation from './annotation';
+import EntityFooter from './footer';
+import EntityImage from './image';
+import EntityLinks from './links';
+import EntityRelatedCollections from './related-collections';
+import EntityTitle from './title';
+import PropTypes from 'prop-types';
+import React from 'react';
+import {formatDate, getEntityTable, getEntityKey} from '../../../helpers/utils';
+
+const {deletedEntityMessage, getTypeAttribute, getEntityUrl, ENTITY_TYPE_ICONS, getSortNameOfDefaultAlias} = entityHelper;
+const {Col, Row} = bootstrap;
+
+function SeriesAttributes({series}) {
+	if (series.deleted) {
+		return deletedEntityMessage;
+	}
+	const sortNameOfDefaultAlias = getSortNameOfDefaultAlias(series);
+	return (
+		<div>
+			<Row>
+				<Col md={3}>
+					<dl>
+						<dt>Sort Name</dt>
+						<dd>{sortNameOfDefaultAlias}</dd>
+					</dl>
+				</Col>
+				<Col md={3}>
+					<dl>
+						<dt>Series Type</dt>
+						<dd>{series.entityType}</dd>
+					</dl>
+				</Col>
+				<Col md={3}> 
+					<dl>
+						<dt>Ordering Type</dt>
+						<dd>{series.seriesOrderingType.label}</dd>
+					</dl>
+				</Col>
+			</Row>
+		</div>
+	);
+}
+SeriesAttributes.displayName = 'SeriesAttributes';
+SeriesAttributes.propTypes = {
+	series: PropTypes.object.isRequired
+};
+
+
+function SeriesDisplayPage({entity, identifierTypes, user}) {
+	const urlPrefix = getEntityUrl(entity);
+	const EntityTable = getEntityTable(entity.entityType);
+	const entityKey = getEntityKey(entity.entityType);
+	const propsForTable = {
+		[entityKey]: entity.entities,
+		showAdd: false,
+		showAddedAtColumn: false,
+		showCheckboxes: false
+	};
+	return (
+		<div>
+			<Row className="entity-display-background">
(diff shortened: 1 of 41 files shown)


src/client/entity-editor/relationship-editor/actions.ts
```diff
@@ -51,6 +53,44 @@ export function addRelationship(data: Relationship): Action {
 	};
 }
 
+export function sortRelationships(oldIndex, newIndex):any {
```
MonkeyDo, line 56: This function could do with inline comments. It's pretty hard to follow what's happening and a new contributor would need to know for example what `orderTypeValue = 1` means.
It would be good to precede each step in the process with a line or two explaining what/why we're doing it.
I don't understand all the code in this method, like why it seems we're going from arrays to objects and back

src/client/entity-editor/relationship-editor/relationship-editor.tsx
```diff
@@ -289,9 +333,15 @@ class RelationshipModal
...
+		const {baseEntity, relationshipTypes, seriesType} = this.props;
 		const {targetEntity} = this.state;
-		const types = getValidOtherEntityTypes(relationshipTypes, baseEntity);
+		let types;
+		if (baseEntity.type === 'Series') {
+			types = [seriesType];
+		}
+		else {
+			types = getValidOtherEntityTypes(relationshipTypes, baseEntity);
+		}
```
MonkeyDo, line 344: Are you sure this is needed? Doesn't `getValidOtherEntityTypes` return the one valid relationship type (X is part of Series)?

src/client/entity-editor/series-section/series-section.tsx
```diff
@@ -0,0 +1,159 @@
...
+	relationships,
+	onOrderTypeChange,
+	onSeriesTypeChange
+}: Props) {
+	const relationshipsObject = relationships.toJS();
+	const relationshipsArray = Object.values(relationshipsObject);
+	const seriesOrderingTypesForDisplay = seriesOrderingTypes.map((type) => ({
+		label: type.label,
+		value: type.id
+	}));
```
MonkeyDo, line 89: All this we should do in mapStateToProps instead

src/client/entity-editor/series-section/series-section.tsx
```diff
@@ -0,0 +1,159 @@
...
+	);
+}
+SeriesSection.displayName = 'SeriesSection';
+
+function mapStateToProps(rootState): StateProps {
+	const state = rootState.get('seriesSection');
+
+	return {
+		orderTypeValue: state.get('orderType'),
+		relationships: rootState.getIn(['relationshipSection', 'relationships']),
```
MonkeyDo, line 142: You should be able to do something like this if I'm not mistaken:
```suggestion
		relationships: rootState.getIn(['relationshipSection', 'relationships']).toArray(),
```

src/server/routes/entity/series.js
```diff
@@ -0,0 +1,241 @@
...
+			relationship[`${attribute.type.name}`] = attribute.value.textValue || -Infinity;
+		});
+	});
+	/* eslint-disable no-nested-ternary */
+	if (series.seriesOrderingType.label === 'Manual') {
+		series.relationships.sort((a, b) => (a.position > b.position ? 1 : b.position > a.position ? -1 : 0));
+	}
+	else {
+		series.relationships.sort((a, b) => (a.number > b.number ? 1 : b.number > a.number ? -1 : 0));
+	}
```
MonkeyDo, line 192: This code is also in another file, isn't it? If so we should extract it as a separate util, considering it's pretty dense code that we only want to change in one place if we ever need to change it

src/server/helpers/middleware.ts
```diff
@@ -72,6 +74,37 @@ export const loadLanguages = makeLoader('Language', 'languages', (a, b) => {
 	return a.name.localeCompare(b.name);
 });
 
+export function loadEntities(req: $Request, res: $Response, next: NextFunction) {
```
MonkeyDo, line 77: The function name is a bit confusing; it promises loading entities but then the code is all about relationships and sorting attributes.

I'm not entirely sure from reading the method why it is needed. Maybe you could explain the steps you took?

src/server/helpers/middleware.ts
```diff
@@ -72,6 +74,37 @@ export const loadLanguages = makeLoader('Language', 'languages', (a, b) => {
...
+	try {
+		const {entity} = res.locals;
+		if (entity.dataId) {
+			const {relationships} = entity;
+			relationships.forEach((relationship) => {
+				relationship.attributeSet.relationshipAttributes.forEach(attribute => {
+					relationship[`${attribute.type.name}`] = attribute.value.textValue || -Infinity;
+				});
+			});
```
MonkeyDo, line 86: Setting the relationship attributes looks like it could maybe be done in `constructRelationships` ?

src/client/entity-editor/relationship-editor/types.ts
```diff
@@ -54,17 +72,23 @@ export type RelationshipWithLabel = {
 };
 
 export type RelationshipForDisplay = {
+	attributes: Array<Attribute>,
 	relationshipType: RelationshipType,
 	sourceEntity: Entity,
 	targetEntity: Entity,
-	rowID: number
+	rowID: string
```
MonkeyDo, line 79: This was necessary for sorting the relationships array in redux, if I understand correctly?

MonkeyDo (changes requested):
It's looking good so far!
I'll deploy this to test.bookbrainz.org as well once the PR in data-js is merged.

There are a couple of unresolved ESLint warnings in `pages/entities/series.js`  (you can see them inlined in the PR "files" tab)

I've got some ideas of how we could improve the editor UI for series, but that'll be for another PR.
I'll make some sketches and we can discuss it.

src/client/entity-editor/entity-editor.tsx
```diff
@@ -81,13 +81,13 @@ const EntityEditor = (props: Props) => {
...
 					<NameSection {...props}/>
 					<ButtonBar {...props}/>
-					<RelationshipSection {...props}/>
 					{
 						React.cloneElement(
 							React.Children.only(children),
 							{...props}
 						)
 					}
+					<RelationshipSection {...props}/>
```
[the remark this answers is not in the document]
akashgp09: props.children renders the EntitySection (WorkSection, AuthorSection, SeriesSection etc).
Untill now it was fine to have EntitySection below RelationshipSection but after Implementing SeriesSection it was very odd to have SeriesSection below RelationshipSection because it have some fields(seriesType, orderingType) which are needed to be filled before moving to RelationshipSection.

Screenshot from 2021-06-23 09-02-07

src/client/entity-editor/relationship-editor/actions.ts
```diff
@@ -51,6 +53,44 @@ export function addRelationship(data: Relationship): Action {
...
+		const relObject = relationships.toJS();
+		const array = Object.entries(relObject);
+		const automaticSort = [];
+		let automaticSortedArr: [string, Relationship][];
+
+		if (orderTypeValue === 1) {
+			array.forEach((relationship:[string, Relationship]) => {
+				relationship[1].attributes.forEach((attribute:_Attribute) => {
+					if (attribute.attributeType === 2) {
+						automaticSort.push({number: attribute.value.textValue || -Infinity, relationshipArray: relationship});
```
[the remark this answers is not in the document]
akashgp09: I was using `Infinity` to sort the relationship array where the number or position values were null. But now in this recent commit [1d7f9ea](https://github.com/bookbrainz/bookbrainz-site/pull/649/commits/1d7f9eaf587b7624e9090359068fe173b92c2304) I have removed any reference to `Infinity` used in the codebase and as well as updated the sorting technique as i encountered some cases where the previous sorting techniques wasn't giving the desired results.

src/client/entity-editor/relationship-editor/relationship-editor.tsx
akashgp09: **getValidOtherEntityTypes** will return all the valid relationship type for series entity i.e `["Author", "Edition", "EditionGroup", "Publisher", "Work"]` 
This will then show all entities in relationship search results
Example: 
Screenshot from 2021-06-23 15-03-06

But we don't want this. Instead we want to restrict the search results to just one entity type. For Example If the user select series type as 'Author' then show only authors in the search results:
Example: 
Screenshot from 2021-06-23 15-06-21
 Hence I think it is necessary to use seriesType value (from the SeriesSection) here.

src/client/entity-editor/relationship-editor/types.ts
akashgp09: Yes exactly !!

src/server/helpers/middleware.ts
akashgp09: Yes indeed it's very confusing. I have renamed it now to `loadSeriesItems` .
- This function first extracts the deeply nested relationship **number** and **position** attribute value and attaches directly to the relationship object. so that we can easily access it (e.g: relationship.number or relationship.position) https://github.com/bookbrainz/bookbrainz-site/pull/649/files#diff-8f13ac13a037cddb521ab9b4829f038c0021e5e761d544b826cb539b70682c77R84
- Secondly we perform sort operations on either number or position attribute of the relationship object depending on the ordering type. https://github.com/bookbrainz/bookbrainz-site/pull/649/files#diff-8f13ac13a037cddb521ab9b4829f038c0021e5e761d544b826cb539b70682c77R84
- At last we extract the required properties from the relationship object to an array **seriesItems** used to display the series items in the entity-table. https://github.com/bookbrainz/bookbrainz-site/pull/649/files#diff-8f13ac13a037cddb521ab9b4829f038c0021e5e761d544b826cb539b70682c77R94-R99

src/client/entity-editor/series-section/series-section.tsx
akashgp09: Line 86-89 is fine here 🤔

src/server/helpers/middleware.ts
akashgp09: I don't think we can use **constructRelationships**.
Here we are extracting the deeply nested number and position attribute value in the relationship object and then attaching directly to the relationship object. so that we can easily access it (e.g: relationship.number or relationship.position)

**Screenshot: Before Vs After**

Screenshot from 2021-06-23 20-00-04

src/client/components/pages/entities/series.js
```diff
@@ -0,0 +1,130 @@
...
+						<dd>{series.entityType}</dd>
+					</dl>
+				</Col>
+				<Col md={3}>
+					<dl>
+						<dt>Ordering Type</dt>
+						<dd>{series.seriesOrderingType.label}</dd>
+					</dl>
+				</Col>
+			</Row>
```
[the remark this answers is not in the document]
akashgp09: do you mean to count the total number of items in the series and display in the frontend something like this:
```
<dl>
	<dt>Total Items</dt>
	<dd>{SomeValue}</dd>
</dl>
```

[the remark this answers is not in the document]
MonkeyDo: Yes, that's what I'm talking about.
I'm not sure what the best way of counting the items is or where that code should live, but it might not be too hard to add.

What I originally had in mind was probably much too complicated (it involved a `count` property on the SeriesData ORM model…) but I think there should be an easy enough solution.

src/client/entity-editor/entity-editor.tsx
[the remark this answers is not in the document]
MonkeyDo: Makes sense, thanks.

src/client/entity-editor/relationship-editor/relationship-editor.tsx
MonkeyDo: Ah, yes, of course there is a relationship type "part of series" for each entity type, I forgot about that.

That makes sense. However I'm worried that we won't be able to have any other type of relationship for series.
For example imagine we want to add a new relationship type "Series was inspired by Work" (quick example: This sci-fi series  by Dan Simmons was inspired by Homer's Illiad: https://en.wikipedia.org/wiki/Ilium/Olympos).
With the current setup we won't be able to search for a work or add a relationship to it.

It's not something we need to solve in this PR, but let's keep in mind that we'll probably have to change that.
The UI changes I have in mind might solve the issue by having a separate component in the SeriesSection where we can search for and add entities to the series, while the relationship editor would be used for the other types of relationships. But let's cross that bridge when we get to it :)

src/client/components/pages/entities/series.js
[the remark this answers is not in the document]
akashgp09: Thanks for clarifying. I think we can simply use **seriesItems.length** to display the total number of items in the series. https://github.com/bookbrainz/bookbrainz-site/pull/649/files#diff-db9f5ec20bca1f2e5812e2fdd4066313ca3ed4dcfeb86ba21b055f1125143c3dR76

src/client/entity-editor/series-section/series-section.tsx
MonkeyDo: Better to do it in mapStateToProps so that it only get recalculated if the state changes, rather than on every re-render of the component( which might rerender for reasons other than a state change)

MonkeyDo: Although, let's be honest, for the two items in this array it won't make a difference :D

src/server/helpers/middleware.ts
MonkeyDo: That was the wrong name, sorry. I was thinking of `loadEntityRelationships` defined in the same file.
The reason I'm suggesting to move the code to a more common method is that we are expecting other relationship attributes such as dates that won't have anything to do with Series and hence won't use the loadSeriesItems middleware.
Since you've already coded it to be generic (```relationship[`${attribute.type.name}`]```) it won't hurt to use it in loadEntityRelationships. I'd only suggest to wrap it in a condition just in case (`if(relationship.attributeSet?.relationshipAttributes)`)

src/client/entity-editor/relationship-editor/actions.ts
```diff
@@ -54,33 +55,43 @@ export function addRelationship(data: Relationship): Action {
...
 }
 
+/**
+ * This action creator first sorts the relationship object and then pass the
+ * sorted object in the payload while dispatching the action.
+ *
+ * @param {number} oldIndex - Old Position of the relationship.
+ * @param {number} newIndex - New Position of the relationship.
+ * @returns {void}
+ */
```
MonkeyDo, line 65: :heart:  Love to see some docs, it's a great habit to have!

src/client/entity-editor/relationship-editor/relationship-editor.tsx
akashgp09: sounds great. Ready to Get it :100: :100:

MonkeyDo (approved):
Nice work!

PR #658, opened 2021-06-30
feat(collection,series): Extend Collection to include Series Entity.

akashgp09:
### Summary 
- Created SeriesTable component.
- Extended collections to include series entity.
### Screenshot
Screenshot from 2021-06-24 15-18-28

### DIFF
--- a/src/server/routes/collection.js
+++ b/src/server/routes/collection.js
@@ -70,6 +70,12 @@ function getEntityRelations(entityType) {
 		'area'
 	];
 
+	const seriesRelations = [
+		'defaultAlias.language',
+		'disambiguation',
+		'seriesOrderingType'
+	];
+
 	const workRelations = [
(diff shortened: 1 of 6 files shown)


src/server/routes/collection.js
```diff
@@ -70,6 +70,12 @@ function getEntityRelations(entityType) {
 		'area'
 	];
 
+	const seriesRelations = [
+		'defaultAlias.language',
+		'disambiguation',
+		'seriesOrderingType'
+	];
```
MonkeyDo, line 77: As a possible future improvement, we could think of a way to load the number of items in a series from the ORM model.
That would require loading related relationships through the relationshipSet and filtering them by relationship type, which would probably be quite a costly process.
Let's keep it in mind for another day.

MonkeyDo (approved):
Looks good, ready to merge !

PR #660, opened 2021-07-02
fix(revision): show type changes in revision page

akashgp09:
### Problem

Revisions for **workType, authorType, publisherType, editionGroupType, editionFormat and editionStatus** are not shown in the revision page.

### Solution

- Fixed some incorrect path name.
- Modified formatChange function and conditional statement.

### Screenshot:
<details>

<summary>1</summary>

# New

### Before
Screenshot from 2021-07-02 11-51-59

### After
Screenshot from 2021-07-02 11-49-39

</details>

<details>

<summary>2</summary>

# Edited

### Before
Screenshot from 2021-07-02 11-52-07

### After
Screenshot from 2021-07-02 11-57-00

</details>

### Areas of Impact

- [diffFormatters/base.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/helpers/diffFormatters/base.js)
- [routes/revision.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/routes/revision.js)

### DIFF
--- a/src/server/helpers/diffFormatters/base.js
+++ b/src/server/helpers/diffFormatters/base.js
@@ -76,7 +76,15 @@ export function formatEndedChange(change) {
 }
 
 export function formatTypeChange(change, label) {
-	return formatChange(change, label, (side) => (typeof side === 'string' ? [side] : side && [side.label]));
+	// eslint-disable-next-line consistent-return
+	return formatChange(change, label, (side) => {
+		if (typeof side === 'string') {
+			return [side];
+		 }
+		 else if (side) {
+			 return [side.label];
+		 }
+	});
 }
 
 export function formatScalarChange(change, label) {
--- a/src/server/helpers/diffFormatters/base.js
+++ b/src/server/helpers/diffFormatters/base.js
@@ -76,7 +76,7 @@ export function formatEndedChange(change) {
 }
 
 export function formatTypeChange(change, label) {
-	return formatChange(change, label, (side) => side && [side.label]);
+	return formatChange(change, label, (side) => (typeof side === 'string' ? [side] : side && [side.label]));
 }
 
 export function formatScalarChange(change, label) {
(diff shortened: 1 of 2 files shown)


src/server/helpers/diffFormatters/base.js
```diff
@@ -76,7 +76,7 @@ export function formatEndedChange(change) {
 }
 
 export function formatTypeChange(change, label) {
-	return formatChange(change, label, (side) => side && [side.label]);
+	return formatChange(change, label, (side) => (typeof side === 'string' ? [side] : side && [side.label]));
```
MonkeyDo, line 79: This is getting a bit hard to read, how about breaking it down again like this?
```suggestion
	return formatChange(change, label, (side) => {
	if(typeof side === 'string'){
	    return [side];
	}
	else if(side) {
	    return [side.label]
	}
	});
```

src/server/routes/revision.js
```diff
@@ -58,8 +58,8 @@ function formatAuthorChange(change) {
 	if (_.isEqual(change.path, ['ended'])) {
 		return baseFormatter.formatEndedChange(change);
 	}
-
-	if (_.isEqual(change.path, ['type'])) {
+	if (_.isEqual(change.path, ['authorType']) ||
+			_.isEqual(change.path, ['authorType', 'label'])) {
```
MonkeyDo, line 62: I wonder if this would be simpler and essentially work the same; what do you think?
This would apply to the other entity types as well.
```suggestion
	if (change.path?.[0] === 'authorType']) {
```

MonkeyDo (changes requested):
Nice catch, and thanks for fixing that !

src/server/routes/revision.js
akashgp09: this will also match the path `['authorType',  'id' ]` leading to a Type Error.

Screenshot from 2021-07-15 19-44-59

MonkeyDo: Ah yes, thanks :)

MonkeyDo (approved):
Good to go !

PR #666, opened 2021-07-16
API: add browse requests for series

akashgp09:
**Summary**

- [x] Added browse requests for series.
- [x] Added series to search. [ `GET /search` ]


src/api/routes/series.js
```diff
@@ -227,4 +230,93 @@ router.get('/:bbid/relationships',
...
+ *         schema:
+ *           type: string
+ *           format: uuid
+ *       - name: seriesOrderingType
+ *         in: query
+ *         description: filter by Series Ordering Type
+ *         required: false
+ *         schema:
+ *           type: string
+ *           enum: [Automatic, Manual]
```
MonkeyDo, line 284: What use cases did you have in mind or this?
I'm not really convinced is would be particularly useful, but I also don't think it hurts in any way to have it there.
I would also propose shortening it to `orderingType` to match what we do for other entities.

src/api/routes/series.js
```diff
@@ -227,4 +230,93 @@ router.get('/:bbid/relationships',
...
+ *               $ref: '#/components/schemas/BrowsedSeries'
+ *       404:
+ *         description: author/work/edition/edition-group/publisher (entity entity) not found
+ *       400:
+ *         description: Invalid BBID passed in the query params OR Multiple browsed entities passed in parameters
+ */
+
+router.get('/',
+	formatQueryParameters(),
+	validateBrowseRequestQueryParameters(['edition', 'author', 'edition-group', 'work', 'publisher']),
```
MonkeyDo, line 300: If we are keeping the `seriesOrderingType` parameter then I assume it will need to be in this list too.

src/api/routes/series.js
```diff
@@ -227,4 +230,93 @@ router.get('/:bbid/relationships',
...
+	makeEntityLoader(null, utils.relationshipsRelations, 'Entity not found', true),
+	loadEntityRelationshipsForBrowse(),
+	async (req, res) => {
+		function relationshipsFilterMethod(relatedEntity) {
+			if (req.query.seriesorderingtype) {
+				return toLower(relatedEntity.seriesOrderingType) === toLower(req.query.seriesorderingtype);
```
MonkeyDo, line 306: There is a mismatch of the capitalization here. Does the query parameter name get automatically transformed from `seriesOrderingType` (as defined in the docs on line 278) to `seriesorderingtype` as seen here on the query object?

The filtering seems to me to be working fine, so we should probably correct the docs?

MonkeyDo (commented):
All working nicely !

And thanks for the extra fixes not related to series that you added in there :)

src/api/routes/edition-group.js
```diff
@@ -271,21 +278,22 @@ router.get('/:bbid/relationships',
  *             schema:
  *               $ref: '#/components/schemas/BrowsedEditionGroups'
  *       404:
- *         description: Edition not found
+ *         description: edition/series (entity entity) not found
```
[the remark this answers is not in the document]
akashgp09: > Is (entity entity) on purpose?

Oh no. It should be `(other entity)`. My bad.

> Also I would keep the capitals on Edition and Series for consistency.

Sure

src/api/routes/series.js
akashgp09: Not really I can think of a scenario where it would be useful but yeah, there isn't any harm to have it there.

src/api/routes/series.js
MonkeyDo: Actually I think I'm wrong there, it's working without it.

src/api/routes/series.js
akashgp09: > Does the query parameter name get automatically transformed from seriesOrderingType (as defined in the docs on line 278) to seriesorderingtype as seen here on the query object ?

Yes, the middleware [formatQueryParameters](https://github.com/bookbrainz/bookbrainz-site/blob/66a1b8706579eb1bfa8b598e0071bf239c408fa9/src/api/helpers/middleware.js#L79-L84) transforms all the query parameters to lowerCase.

MonkeyDo (approved):

PR #676, opened 2021-07-22
FIX(BB-627): set badges in achievement section

akashgp09:
### Problem

This PR Fixes: [#627](https://tickets.metabrainz.org/projects/BB/issues/BB-627)

Two response were sent to the client side upon a request leading to **Uncaught Exception: Error [ERR_HTTP_HEADERS_SENT]  Cannot set headers after they are sent to the client**

First Response :

https://github.com/bookbrainz/bookbrainz-site/blob/66a1b8706579eb1bfa8b598e0071bf239c408fa9/src/server/routes/editor.js#L506
Second Response:
https://github.com/bookbrainz/bookbrainz-site/blob/66a1b8706579eb1bfa8b598e0071bf239c408fa9/src/server/routes/editor.js#L511
https://github.com/bookbrainz/bookbrainz-site/blob/66a1b8706579eb1bfa8b598e0071bf239c408fa9/src/server/helpers/handler.js#L27

### Solution

return the response statement and refactor code.
https://github.com/bookbrainz/bookbrainz-site/blob/66a1b8706579eb1bfa8b598e0071bf239c408fa9/src/server/routes/editor.js#L506

### Areas of Impact

- [server/routes/editor.js](https://github.com/bookbrainz/bookbrainz-site/blob/master/src/server/routes/editor.js)

### DIFF
--- a/src/server/routes/editor.js
+++ b/src/server/routes/editor.js
@@ -485,30 +485,17 @@ async function rankUpdate(orm, editorId, bodyRank, rank) {
 }
 
 
-router.post('/:id/achievements/', auth.isAuthenticated, (req, res) => {
+router.post('/:id/achievements/', auth.isAuthenticated, async (req, res) => {
 	const {orm} = req.app.locals;
 	const {Editor} = orm;
 	const userId = parseInt(req.params.id, 10);
-
-	async function runAsync() {
-		if (!isCurrentUser(userId, req.user)) {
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
That looks better, thanks !

I looked at how this endpoint is used, and we don't actually need to have the `rankJSON` returned as a response (which from reading it is what the code was previously trying to achieve), so it's all good !

PR #679, opened 2021-07-30
Revert relationship editor changes

akashgp09:
**Summary:** 
Reverted some of the relationship editor changes.


MonkeyDo (approved):
Looks good. The more I look at it, the more it makes sense to have all this series-specific logic and presentation in a separate component.
Glad you have already got the bulk of the logic done, and it just needs to be moved :)

PR #682, opened 2021-08-03
filter relationships when baseEntity is Series

akashgp09:
**Summary:**
When the base entity is a Series, we need to filter out relationshipType 70 - 74. As this relationships will be added via the series editor, we don't want to show them in the relationship modal.

### DIFF
--- a/src/client/entity-editor/relationship-editor/relationship-editor.tsx
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.tsx
@@ -355,6 +355,12 @@ class RelationshipModal
 		const relationships = generateRelationshipSelection(
 			relationshipTypes, baseEntity, otherEntity
 		);
+		if (baseEntity.type === 'Series') {
+			// when the base entity is Series, we need to remove relationshipType 70 - 74.
+			// As this relationships will be added via the series editor,
+			// we don't want to show them in the relationship modal.
+			_.remove(relationships, (relationship) => relationship.relationshipType.id > 69 && relationship.relationshipType.id < 75);
+		}
 
 		return (
 			<FormGroup>
--- a/src/client/entity-editor/relationship-editor/relationship-editor.tsx
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.tsx
@@ -300,6 +300,11 @@ class RelationshipModal
 	renderEntitySelect() {
 		const {baseEntity, relationshipTypes} = this.props;
 		const {targetEntity} = this.state;
+		if (baseEntity.type === 'Series') {
+			// When the base entity is Series, remove relationship type 70 - 74.
+			// We don't want to generate entity types corresponding to these relationship type.
+			_.remove(relationshipTypes, (type) => type.id > 69 && type.id < 75);
+		}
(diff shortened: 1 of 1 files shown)


src/client/entity-editor/relationship-editor/relationship-editor.tsx
```diff
@@ -300,6 +300,11 @@ class RelationshipModal
 	renderEntitySelect() {
 		const {baseEntity, relationshipTypes} = this.props;
 		const {targetEntity} = this.state;
+		if (baseEntity.type === 'Series') {
+			// When the base entity is Series, remove relationship type 70 - 74.
+			// We don't want to generate entity types corresponding to these relationship type.
+			_.remove(relationshipTypes, (type) => type.id > 69 && type.id < 75);
+		}
```
MonkeyDo, line 307: I feel this would be better in the `getValidOtherEntityTypes` helper instead, but the logic is what I'm expecting

src/client/entity-editor/relationship-editor/relationship-editor.tsx
```diff
@@ -355,6 +360,12 @@ class RelationshipModal
 		const relationships = generateRelationshipSelection(
 			relationshipTypes, baseEntity, otherEntity
 		);
+		if (baseEntity.type === 'Series') {
+			// When the base entity is Series, we need to remove relationshipType 70 - 74.
+			// As these relationships will be added via the series editor,
+			// we don't want to show them in the relationship modal.
+			_.remove(relationships, (relationship) => relationship.relationshipType.id > 69 && relationship.relationshipType.id < 75);
+		}
```
MonkeyDo, line 368: Since this is essentially the same code as above, to avoid duplication I wonder if it can be extracted to a filter function similar to the existing `isValidRelationship` function in the same file.
That filter function can then be used both in `getValidOtherEntityTypes` and like this in `generateRelationshipSection`:

```
let validRels = candidateRelationships.filter(isValidRelationship);
if (baseEntity.type === 'Series') {
validRels = validRels.filter(isSeriesRelationship); /* <- The new reusable filtering method*/
}
```

akashgp09: @MonkeyDo the reason to write this code explicitly like this was because now the `generateRelationshipSection` is also being used here https://github.com/bookbrainz/bookbrainz-site/pull/680/files#diff-ce680346708de8b052297af757840fc1df4f57226f9223ddc9c02d8c7af71d47R92-R94

If I add the following code to `generateRelationshipSection` function then it will remove relationshipType 70 - 74 for the new series editor component as well. 
```
let validRels = candidateRelationships.filter(isValidRelationship);
if (baseEntity.type === 'Series') {
validRels = validRels.filter(isSeriesRelationship); /* <- The new reusable filtering method*/
}
```

MonkeyDo: I see, that makes sense, thanks :)

I also see now that the structure of the array we are filtering  is different (type.id vs relationship.relationshipType.id), so not an ideal candidate for refactoring.
In any case, the comments you added do help understand what is happening.

MonkeyDo (approved):
👍

PR #683, opened 2021-08-03
added new relationship type

akashgp09:
**Summary:**

Added new relationship type for series.


MonkeyDo (changes requested):
I would like to see this new relationship added along with the other new series relationships in sql/migrations/relationship-attributes/up.sql 
That will mean one less migration to run (= one less for me to forget to run :) )

MonkeyDo (approved):
Thanks !

PR #684, opened 2021-08-04
feat: set relationship attributes value in relationship-editor

akashgp09:
**Summary:**

Set relationship attributes value in relationship-editor.

Screenshot from 2021-08-04 07-37-13

### DIFF
--- a/src/client/entity-editor/relationship-editor/relationship-editor.tsx
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.tsx
@@ -288,13 +288,8 @@ class RelationshipModal
 			attributeType: 2,
 			value: {textValue: value}
 		};
-		const attributePosition = {
-			attributeType: 1,
-			value: {textValue: null}
-		};
 		this.setState({
-			attributeNumber,
-			attributePosition
+			attributeNumber
 		});
 	};
 
--- a/src/client/entity-editor/relationship-editor/relationship-editor.tsx
+++ b/src/client/entity-editor/relationship-editor/relationship-editor.tsx
@@ -38,6 +38,7 @@ import type {
 	Relationship as _Relationship
 } from './types';
 import {faExternalLinkAlt, faPlus, faTimes} from '@fortawesome/free-solid-svg-icons';
+import {getInitAttribute, setAttribute} from './helper';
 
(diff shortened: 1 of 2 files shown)


src/client/entity-editor/relationship-editor/relationship-editor.tsx
```diff
@@ -267,11 +277,31 @@ class RelationshipModal
...
+	handleNumberAttributeChange = ({target}) => {
+		const value = target.value === '' ? null : target.value;
+		const attributeNumber = {
+			attributeType: 2,
+			value: {textValue: value}
+		};
+		const attributePosition = {
+			attributeType: 1,
+			value: {textValue: null}
+		};
```
MonkeyDo, line 289: Is it required to set the position attribute in this method? Does that mean changing the number attribute resets the position attribute to null?
From the name, I would expect it to only change the number attribute.

What will happen once we implement other relationship attributes like date? Will we have to set each attribute not concerned by the handler to null (i.e. in handleNumberAttributeChange set position and date to null, in handleDateAttributeChange set number and position to null, etc.) ? If so we should try to avoid that situation.

MonkeyDo (commented):
You are paving the way for more relationship attributes! :)

src/client/entity-editor/relationship-editor/relationship-editor.tsx
akashgp09: > What will happen once we implement other relationship attributes like date? Will we have to set each attribute not concerned by the handler to null (i.e. in handleNumberAttributeChange set position and date to null, in handleDateAttributeChange set number and position to null, etc.) ? If so we should try to avoid that situation.

My bad, actually I should give it a different name,  maybe **handleOrdinalAttributeChange.**(This will update both number and position)

And for clarity by no means **handleDateAttributeChange** will have to set number or position inside  it to null.

> Is it required to set the position attribute in this method? Does that mean changing the number attribute resets the position attribute to null?

Hmm, this is a problem. I need to think about it.

akashgp09:
Alright, @MonkeyDo I realized some mistakes in my Implementation. We don't need to set position value inside `handleNumberAttributeChange`. I have removed it now 😅

MonkeyDo (approved):
Thanks for looking it over !

PR #690, opened 2021-08-16
feat(series): drag and drop to order series items.

akashgp09:
**Summary:**

- Added drag and drop feature to order the items in the series in any way you want.

series-10

### DIFF
--- a/src/client/entity-editor/series-section/actions.ts
+++ b/src/client/entity-editor/series-section/actions.ts
@@ -18,6 +18,8 @@
 
 
 import type {Attribute, Relationship} from '../relationship-editor/types';
+import arrayMove from 'array-move';
+import {sortRelationshipOrdinal} from '../../../common/helpers/utils';
 
 
 export const UPDATE_ORDER_TYPE = 'UPDATE_ORDER_TYPE';
@@ -25,6 +27,8 @@ export const UPDATE_SERIES_TYPE = 'UPDATE_SERIES_TYPE';
 export const ADD_SERIES_ITEM = 'ADD_SERIES_ITEM';
 export const EDIT_SERIES_ITEM = 'EDIT_SERIES_ITEM';
 export const REMOVE_SERIES_ITEM = 'REMOVE_SERIES_ITEM';
+export const SORT_SERIES_ITEM = 'SORT_SERIES_ITEM';
+
 export type Action = {
 	type: string,
 	payload: {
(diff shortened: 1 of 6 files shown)


src/client/entity-editor/series-section/actions.ts
```diff
@@ -81,6 +86,52 @@ export function addSeriesItem(data: Relationship): Action {
...
+
+		if (orderTypeValue === 1) { // OrderType 1 for Automatic Ordering
+			seriesItemsArray.forEach((seriesItem:[string, Relationship]) => {
+				seriesItem[1].attributes.forEach((attribute: Attribute) => {
+					if (attribute.attributeType === 2) { // Attribute Type 2 for number
+						automaticSort.push({number: attribute.value.textValue, seriesItem});
+					}
+				});
+			});
+			automaticSort.sort(sortRelationshipOrdinal('number')); // sorts the array of series items on number attribute
```
MonkeyDo, line 116: We'll have to have another look at this sorting function to implement a natural sort order; currently it will do a language sort, so it will sort for example 1 > 10 > 100 > 2

image

This can be done in another PR

src/client/entity-editor/series-section/actions.ts
```diff
@@ -81,6 +86,52 @@ export function addSeriesItem(data: Relationship): Action {
...
+ * @param {number} newIndex - New Position of the series item.
+ * @returns {void}
+ */
+export function sortSeriesItems(oldIndex, newIndex):any {
+	return (dispatch, getState) => {
+		const state = getState();
+		const seriesItems = state.get('seriesSection').get('seriesItems');
+		const orderTypeValue = state.get('seriesSection').get('orderType');
+		const seriesItemsObject = seriesItems.toJS();
+		const seriesItemsArray = Object.entries(seriesItemsObject);
```
MonkeyDo, line 104: It seems to me a bit cumbersome to have to use entries and end up with [ [string , Relationship] ] and having to get the relationship properties with `seriesItem[1].attributes`, instead of using Object.values() and getting an array of Relationship objects directly (considering the rowID is also in the Relationship object).
Does it seem easier to you as well to work with a Relationship[] instead?

I do see that further down you do `Object.fromEntries(new Map([...sortedSeriesItems]))`, maybe that's the reason you opted for this?
In that case, perhaps a simple one-liner array.reduce can give us the same result:
`const payload = sortedSeriesItems.reduce((accumulator,rel:RelationshipForDisplay)=> (accumulator[rel.rowID]=rel,accumulator),{});`

I'm using the type RelationshipForDisplay instead of Relationship as it has the rowID defined.

src/client/entity-editor/series-section/series-editor.tsx
```diff
@@ -74,7 +78,64 @@ type SeriesItemsProps = {
  * @returns {ReactElement} React Component as a modular section within the entity editor.
  *
  */
-function SeriesEditor({baseEntity, relationshipTypes, seriesType, onRemove, onAdd, onEdit, seriesItemsArray}:SeriesItemsProps) {
+
+const SeriesList = ({value, baseEntity, handleNumberAttributeChange, onRemove, dragHandler}) => (
```
MonkeyDo, line 82: I think the name is misleading if this component is for one item rather than the list of items. Maybe SeriesItem or SeriesListItem?

src/client/entity-editor/series-section/series-editor.tsx
```diff
@@ -123,45 +185,40 @@ function SeriesEditor({baseEntity, relationshipTypes, seriesType, onRemove, onAd
...
-									onChange={handleNumberAttributeChange.bind(this, value.rowID)}
+				<> {
+					orderType === 1 ?
+						<>
+							{seriesItemsArray.map((value) => (
+								<SeriesList
+									baseEntity={baseEntity}
+									dragHandler={false}
+									handleNumberAttributeChange={handleNumberAttributeChange}
+									key={`${value.rowID}`}
```
MonkeyDo, line 202:
```suggestion
									key={value.rowID}
```

src/client/entity-editor/series-section/series-editor.tsx
```diff
@@ -123,45 +185,40 @@ function SeriesEditor({baseEntity, relationshipTypes, seriesType, onRemove, onAd
...
+							))
+							}
+						</> :
+						<SortableList distance={1} onSortEnd={onSort}>
+							{seriesItemsArray.map((value, index) => (
+								<SortableItem
+									baseEntity={baseEntity}
+									handleNumberAttributeChange={handleNumberAttributeChange}
+									index={index}
+									key={`${value.rowID}`}
```
MonkeyDo, line 215:
```suggestion
									key={value.rowID}
```

MonkeyDo (commented):
This is working very well, nothing to say about the implementation apart from good job :)
I do have some comments about code styling below:

src/client/entity-editor/series-section/actions.ts
akashgp09: Yeah looks good, I will refactor it :)

MonkeyDo (approved):
Great!

src/client/entity-editor/series-section/actions.ts
MonkeyDo: See PR #691

PR #691, opened 2021-08-18
FIX: Allow sorting with numeric value

akashgp09:
### Problem

https://github.com/bookbrainz/bookbrainz-site/pull/690#discussion_r690578719

### Solution

Pass **`{numeric: true}`** option to **localeCompare.**

#### Numeric sorting
```
// by default, "2" > "10"
console.log("2".localeCompare("10")); // 1

// numeric using options:
console.log("2".localeCompare("10", undefined, {numeric: true})); // -1
```

### Areas of Impact

https://github.com/bookbrainz/bookbrainz-site/blob/series-entity/src/common/helpers/utils.ts


MonkeyDo (approved):
👍

PR #696, opened 2021-08-31
Notifications Service

prabalsingh24:
### Problem
Adding a feature to subscribe entities and get notified if and when any changes are made to those entities

### Solution
https://community.metabrainz.org/t/gsoc-2021-notification-service-bookbrainz-prabal/526184

### DIFF
--- a/sql/migrations/2021-09-notification-service/up.sql
+++ b/sql/migrations/2021-09-notification-service/up.sql
@@ -1,0 +1,28 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS bookbrainz.notification (
+    id UUID PRIMARY KEY DEFAULT public.uuid_generate_v4(),
+    subscriber_id INT NOT NULL,
+    read BOOLEAN NOT NULL DEFAULT FALSE,
+    notification_text TEXT NOT NULL CHECK (notification_text <> ''),
+    notification_redirect_link TEXT NOT NULL CHECK (notification_redirect_link <> ''),
+    timestamp TIMESTAMP WITHOUT TIME ZONE NOT NULL DEFAULT timezone('UTC'::TEXT, now())
+);
+ALTER TABLE bookbrainz.notification ADD FOREIGN KEY (subscriber_id) REFERENCES  bookbrainz.editor (id);
+
+CREATE TABLE bookbrainz.entity_subscription (
+    bbid UUID,
+    subscriber_id INT
+);
+ALTER TABLE bookbrainz.entity_subscription ADD FOREIGN KEY (bbid) REFERENCES bookbrainz.entity (bbid);
+ALTER TABLE bookbrainz.entity_subscription ADD FOREIGN KEY (subscriber_id) REFERENCES bookbrainz.editor (id);
+
+
+CREATE TABLE bookbrainz.collection_subscription (
+    collection_id UUID,
+    subscriber_id INT
+);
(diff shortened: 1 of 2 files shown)


sql/migrations/2021-09-notification-service/up.sql
```diff
@@ -0,0 +1,28 @@
...
+    bbid UUID,
+    subscriber_id INT
+);
+ALTER TABLE bookbrainz.entity_subscription ADD FOREIGN KEY (bbid) REFERENCES bookbrainz.entity (bbid);
+ALTER TABLE bookbrainz.entity_subscription ADD FOREIGN KEY (subscriber_id) REFERENCES bookbrainz.editor (id);
+
+
+CREATE TABLE bookbrainz.collection_subscription (
+    collection_id UUID,
+    subscriber_id INT
```
prabalsingh24, line 23: I think it'll be better if I make `PRIMARY KEY (collection_id, subscriber_id)` what do you think?

src/client/components/pages/collection.js
```diff
@@ -224,6 +232,59 @@ class CollectionPage extends React.Component {
 		}, this.pagerElementRef.triggerSearch);
 	}
 
+	handleSubscribe() {
```
MonkeyDo, line 235: As far as I can tell, all this code (`handleSubscribe`, `handleUnsubscribe`, `setIsSubscribed` and the buttons that go with it) is duplicated for each display page. Instead it would be better if the whole subscription part was a separate component that can then be added to the relevant pages.
You'll probably want to pass the submissionUrl as a prop to make it more flexible.

We might end up with separate CollectionSubscription  and an EntitySubscription components if they require very different implementations.

src/client/components/pages/entities/author.js
```diff
@@ -109,6 +109,38 @@ AuthorAttributes.propTypes = {
...
+			.send({bbid})
+			.then((res) => {
+				setIsSubscribed(false);
+			}, (error) => {
+				// eslint-disable-next-line no-console
+				console.log('error thrown');
+			});
+	}
+	function handleSubscribe(bbid) {
+		const submissionUrl = '/subscription/subscribe/entity';
```
MonkeyDo, line 132: I think this URL structure would be more appropriate and "RESTy":
`/${entityType}/${bbid}/subscribe`  and associated `/unsubscribe` and `/subscribed`
I guess that means we don't need to send any information with that request anymore: BBID is in the route and the user needs to be logged in to access the route so we have their info in the route handler.

That would also work for collections: `/collection/${collectionId}/subscribe`…

That means a fair amount of refactoring on the routes, I know, sorry about that :/
I think you'll end up with a middleware or two (one for entities, one for collections) which should be reusable for each entity's routes. In theory, in these routes you should have access to the user information

src/client/components/pages/collection.js
```diff
@@ -224,6 +232,59 @@ class CollectionPage extends React.Component {
...
+		const submissionUrl = '/subscription/subscribe/collection';
+		const collectionId = this.props.collection.id;
+		const subscriberId = this.props.userId;
```
MonkeyDo, line 238: Would this not be accessible in the route handler on the server?
How come you don't need to send the userId for the entity subscribe routes but you do send them for the collections subscribe route?

MonkeyDo (changes requested):
Looking good so far.

My main comments below aim to refactor and consolidate as much code as possible so that we only have to make change in one place if the code is the same or similar enough.

I'll do another round of more thorough review of the SQL bits, since I'm not super well versed in SQL it takes me extra time to make sure I'm not missing anything, although I think there won't be much to change anyway, it looks good already!

PR #697, opened 2021-09-01
Small refactor of revision page code

MonkeyDo:
Forgot to commit this before merging #626 
Remove the use of the costly .toJSON() function on ORM model, since we only need the BBID from the model.


akashgp09 (approved):
Looks clean and working fine. Thank you for modifications @MonkeyDo :rocket:

PR #698, opened 2021-09-02
FIX: Disallow to add dissimilar entity to series

akashgp09:
**Summary:**

Do not show **series** entities whose **`entityType`** is not same as the **`baseEntity.type`** in the search results of the relationship editor.

### DIFF
--- a/src/client/entity-editor/common/entity-search-field-option.js
+++ b/src/client/entity-editor/common/entity-search-field-option.js
@@ -97,6 +97,11 @@ class EntitySearchFieldOption extends React.Component {
 				type: this.props.type
 			});
 		const options = response.body.filter(entity => entity.bbid !== this.props.bbid);
+		if (this.props.isRelationshipEditor) {
+			const {baseEntity} = this.props;
+			// If series entities entityType doesn't match with the baseEntity type, remove them from the options.
+			_.remove(options, entity => entity.type === 'Series' && baseEntity.type !== entity.entityType);
+		}
 		return {
 			options: options.map(this.entityToOption)
 		};
@@ -122,9 +127,11 @@ class EntitySearchFieldOption extends React.Component {
 
 EntitySearchFieldOption.displayName = 'EntitySearchFieldOption';
 EntitySearchFieldOption.propTypes = {
+	baseEntity: PropTypes.object,
 	bbid: PropTypes.string,
 	empty: PropTypes.bool,
 	error: PropTypes.bool,
+	isRelationshipEditor: PropTypes.bool,
 	label: PropTypes.string.isRequired,
 	languageOptions: PropTypes.array,
 	tooltipText: PropTypes.string,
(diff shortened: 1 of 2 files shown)


MonkeyDo (commented):
Although this works as intended, I was wondering if we could come up with a more generic way to filter results.
I was uncomfortable with the separation of concerns here: the search field component shouldn't have to know about the relationshipEditor or about the Series entity type really.

I had a little plunge into the elasticsearch docs to see if we could pass some filters with the query, but I'm not certain of my results and it seems to bring complications and also tie us to ES syntax.

However, there is a way we can separate the filtering logic:
1. Modify the entity-search-field-option component to accept an optional filters array defaulting to an empty array
2. Move the logic to determine if entity is a series and needs a filter on entityType in the relationshipEditor component
3. Pass that filter to the EntitySearchFieldOption component
4. in that component, refactor the bbid filtering function and concatenate with the filters array from props

How does that sound to you?
We won't need the isRelationshipEditor or baseEntity props anymore and have better separation of concerns I think


It's a soft filtering, meaning that if the

akashgp09:
> 2. Move the logic to determine if entity is a series and needs a filter on entityType in the relationshipEditor component

On which array will I perform this filtering logic in the relationshipEditor component? like right now I am filtering the [**options**](https://github.com/bookbrainz/bookbrainz-site/pull/698/files#diff-a165f7370710f87d009788cd008d3a230a0657af6f54156ebb100b1aac9437fdR103) array in the entity-search-field-option component.

MonkeyDo:
> > 1. Move the logic to determine if entity is a series and needs a filter on entityType in the relationshipEditor component
> 
> Then on which array will I have to perform the filtering logic in the relationshipEditor component? like right now I am filtering the [**options**](https://github.com/bookbrainz/bookbrainz-site/pull/698/files#diff-a165f7370710f87d009788cd008d3a230a0657af6f54156ebb100b1aac9437fdR103) array in the entity-search-field-option component.

You'll still filter the options array; it's the filters that we'll be passing down as props.
Something like this should work:

In the RelationshipEditor component:

```
const bbid = this.props.baseEntity?.bbid;
// We create a function to filter out Series of a different entityType than the current entity
// This ensures we don't add a Work to an Author Series, for example
const filterDifferentTypeSeries = (entity) => entity.type === 'Series' && baseEntity.type !== entity.entityType;
const additionalFilters = [filterDifferentTypeSeries];
return (
<EntitySearchFieldOption
				isRelationshipEditor
				baseEntity={baseEntity}
				bbid={bbid}
				filters={additionalFilters}
…
```

In the EntitySearchFieldOption:
```

const isSameBBIDFilter = (entity) => entity.bbid !== this.props.bbid;
const combineFilters = (...filters) => (item) => {
  return filters.map((filter) => filter(item)).every((x) => x === true);
};
//  default this.props.filters to an empty array
const combinedFilters = combineFilters(isSameBBIDFilter, ...(this.props.filters || []) )
const filteredOptions = response.body.filter(combinedFilters );
```

MonkeyDo (approved):
This works great now, and the optional filters might very well be useful in other ways in the future. Thanks !

MonkeyDo:
I deployed this PR to test.bookbrainz.org

PR #718, opened 2021-11-26
Fix[BB-458]: Showing error conflict page on deleting entity twice

tr1ten:
### Problem

fixing BB-458 bug

### Solution

show  error conflict page when user tries to delete same entity twice.

### Areas of Impact

1. src/server/routes/entity/author.js
2. src/server/routes/entity/edition-group.js
3. src/server/routes/entity/edition.js
4. src/server/routes/entity/publisher.js
5. src/server/routes/entity/series.js
6. src/server/routes/entity/work.js

image

### DIFF
--- a/src/server/routes/entity/author.js
+++ b/src/server/routes/entity/author.js
@@ -21,6 +21,7 @@ import * as auth from '../../helpers/auth';
 import * as entityRoutes from './entity';
 import * as middleware from '../../helpers/middleware';
 import * as utils from '../../helpers/utils';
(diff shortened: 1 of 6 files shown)


MonkeyDo (approved):
That's great, now it's a bit safer and we can't double-delete entities.

Thanks for your efforts !
In comparison to the other PRs, this one was a bit more involved and I think you did absolutely great finding your way through this complicated codebase !
Well done 🎉

PR #719, opened 2021-11-27
fix[BB-630]: fix regex to match ~ in url

tr1ten:
### Problem

BB-630 : URLs with ~ in edit notes do not get parsed properly

### Solution

added '~' match in urlRegex

### Areas of Impact

1. src/client/helpers/utils.tsx

### DIFF
--- a/src/client/helpers/utils.tsx
+++ b/src/client/helpers/utils.tsx
@@ -180,7 +180,7 @@ export function dateObjectToISOString(value: DateObject) {
 export function stringToHTMLWithLinks(string: string) {
 	const addHttpRegex = /(^| )www\./ig;
 	// eslint-disable-next-line max-len
-	const urlRegex = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%~@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/g;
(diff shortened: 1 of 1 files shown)


src/client/helpers/utils.tsx
```diff
@@ -180,7 +180,7 @@ export function dateObjectToISOString(value: DateObject) {
 export function stringToHTMLWithLinks(string: string) {
 	const addHttpRegex = /(^| )www\./ig;
 	// eslint-disable-next-line max-len
-	const urlRegex = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/g;
+	const urlRegex = /((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%~@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/g;
```
MonkeyDo, line 183: Works as promised !

I did see however that in the same URL described in the ticket (`https://lib.mau.se/record=b1127537~S5*swe`) there is another character that is not currently recognized: "*" and the following characters.

So the URL parsing went from 
`https://lib.mau.se/record=b1127537`
to
`https://lib.mau.se/record=b1127537~S5`
and we would expect in the end the entire URL:
`https://lib.mau.se/record=b1127537~S5*swe`

I think it should be enough for now to add the *  character as you did for the ~ 
If anyone runs into other special cases in the future we can open another ticket.

MonkeyDo (changes requested):
Yes! Simple solution, thank you for looking into this !

src/client/helpers/utils.tsx
tr1ten: Is "*" also considered a valid url character? , i looked at some regex for url matching most  of them exclude * so i thought it would be bettter without it . nonetheless i will add * match in regex then if you say so

MonkeyDo: I was looking into it and found that it is a "reserved" character, but nonetheless valid and as we see sometimes used by some creative developers…

MonkeyDo (approved):
Thanks a lot !

PR #722, opened 2021-11-30
Fix[BB-513] : added missing entity option for every search and prefilled the name

tr1ten:
### Problem

BB-513 :  Improve "Are we missing an entry?" section.

### Solution

Added missing entity option in every search result and prefilled the new entity with searched name using query parameters.

### Areas of Impact

1. src/client/components/pages/parts/call-to-action.js
2. src/client/components/pages/parts/search-results.js
3. src/client/components/pages/search.tsx
4. src/server/routes/entity/author.js
5. src/server/routes/entity/edition-group.js
6. src/server/routes/entity/edition.js
7. src/server/routes/entity/publisher.js
8. src/server/routes/entity/work.js


src/server/routes/entity/author.js
```diff
@@ -96,11 +96,20 @@ router.get(
...
 			'author', req, res, {
 				genderOptions: res.locals.genders
 			}
-		));
+		);
+		markupProps.initialState.nameSection = {
+			disambiguation: '',
+			exactMatches: null,
+			language: null,
+			name: req.query.name,
```
MonkeyDo, line 108:
```suggestion
			name: req.query?.name ?? '',
```

This ensures there is a default empty string if query or query.name are undefined.
Without that, creating an entity without an initial name results in a server error in [isAliasEmpty](https://github.com/bookbrainz/bookbrainz-site/blob/88ce6fe4a61cf755ff14f5ad764d48a8070976cf/src/client/entity-editor/helpers.ts#L61) and breaking the page:
```
× TypeError: Cannot read property 'length' of undefined
     at isAliasEmpty (/home/bookbrainz/bookbrainz-site/lib/client/entity-editor/helpers.js:109:22)
     at NameSection.render (/home/bookbrainz/bookbrainz-site/lib/client/entity-editor/name-section/name-section.js:207:42)
```
     
The same applies for the other routes too.

src/client/components/pages/parts/call-to-action.js
```diff
@@ -41,35 +43,35 @@ function CallToAction() {
 				<ButtonGroup id="searchpage-button-group">
 					<Button
 						className="padding-bottom-1 padding-sides-2 padding-top-1"
-						href="/author/create"
+						href={props.query ? `/author/create?name=${props.query}` : '/author/create'}
```
MonkeyDo, line 46: A suggestion to make this a bit more readable:
Since the name will be the same for each route, we can create a variable once at the start of this component function:
```const nameQueryParameter = props.query ? `?name=${props.query}` : "";```

And then use it like so for each route below:
```href={`/author/create${nameQueryParameter}`}```

**Review by MonkeyDo — CHANGES_REQUESTED** [review]
Nice work so far.
I have some suggestions to improve a few things but generally you're going in the right direction, nicely done :)

File: src/client/components/pages/search.tsx [hunk 3]
```diff
@@ -157,6 +160,26 @@ class SearchPage extends React.Component<Props, State> {
...
+					</div>}
+
+					<Row>
+						{results.length === 0 &&
+							<small>Make sure the spelling is correct, and that
+								 you have selected the correct type in the search bar.
+							</small>}
+						<hr className="wide"/>
+						<h3>Are we missing an entry?</h3>
+						<CallToAction query={this.props.query}/>
```
MonkeyDo, line 179:
```suggestion
						<CallToAction query={query}/>
```

Just realized when I tried locally that we need to use the query from state and not props, otherwise changes won't be taken into account (search for 'abc' -> CallToAction set with query='abc' -> search for 'def' -> CallToAction still set to query='abc')

MonkeyDo (changes requested):
Last issue I found while testing, other than that it looks great, thanks !

MonkeyDo:
I took the liberty to fix a merge conflict, since it was not super trivial (nothing complicated but required knowing what moved where).

MonkeyDo (approved):

PR #723, opened 2021-11-30
Use Github actions to run CI tests

MonkeyDo:
Travis is dead, long live CI !

Travis CI stopped their OSS-friendly programs, and as a consequence we can't use it for testing anymore.

Instead, we're moving testing to Github actions, like we have been doing successfully with [ListenBrainz](https://github.com/metabrainz/listenbrainz-server/tree/master/.github/workflows)


.github/workflows/test-report.yml
```diff
@@ -0,0 +1,16 @@
+name: Test Report
+on:
+  workflow_run:
+    workflows: ['CI']                     # runs after CI workflow
+    types:
+      - completed
+jobs:
```
amCap1712, line 7: Why not just run this as an extra step after exporting test results?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,80 @@
...
+      # For more information, see https://docs.npmjs.com/cli/ci.html
+      - name: Install dependencies
+        run: npm ci
+      
+      - name: Run tests
+        run: npm run test-ci                    # run tests (configured to use mocha's json reporter)
+      
+      - name: Export test results
+        uses: actions/upload-artifact@v2  # upload test results
+        if: success() || failure()        # run this step even if previous step failed
```
amCap1712, line 77: There's always too ;) https://docs.github.com/en/actions/learn-github-actions/expressions#always

.github/workflows/test-report.yml
MonkeyDo: Following the instructions on the test-reporter action:
https://github.com/dorny/test-reporter#recommended-setup-for-public-repositories

I haven't tested it for real yet, so not 100% sure it's necessary, but it certainly won't hurt to try as advertised first.

.github/workflows/ci.yml
MonkeyDo: I suppose that doesn't include the `cancelled()` status, which `always()` does include according to the docs. Probably best not to run if the job is cancelled.

.github/workflows/test-report.yml
MonkeyDo: I'll keep this comment unresolved and remember to try it out.

amCap1712 (approved):

.github/workflows/test-report.yml
MonkeyDo: Re-read this article and confirming the recommended "safe" way to check out PRs from forked repos without allowing the forked repo to have write access to our repo is to run any task that requires write access as a separate workflow where the context will be safe

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

PR #730, opened 2021-12-01
Replace NPM with Yarn

MonkeyDo:
The Yarn package manager is a lot faster when it comes to installing dependencies, making a real and big difference for local development and Docker image builds in particular.

Yarn comes bundled with our node base docker image, so we might as well use it.


MonkeyDo:
Not sure why CI tests are now failing, will need to have a closer look

PR #731, opened 2021-12-02
Fix[BB-604]: making datepicker and link button responisve on small screen devices.

tr1ten:
### Problem

BB-604: The date picker popper is hidden on small screen devices
( i also notice overflow issue with link button so fixed it as well )

### Solution

add some css to fix the responsive problem  

### Areas of Impact

1. src/client/stylesheets/style.less
2. src/client/entity-editor/common/new-date-field.js
3. src/client/entity-editor/button-bar/button-bar.js

### DIFF
--- a/src/client/entity-editor/common/new-date-field.js
+++ b/src/client/entity-editor/common/new-date-field.js
@@ -140,10 +140,11 @@ class DateField extends React.Component {
 					groupClassName={groupClassName}
 					label={labelElement}
 				>
-					<InputGroup style={{width: '18em'}}>
+					<InputGroup className="res-dob">
 						<FormControl
 							maxLength={isCommonEraDate ? 4 : 5}
 							placeholder="YYYY"
+							style={{width: '4em'}}
 							type="text"
 							value={this.state.year}
 							onBlur={this.handleYearInputBlur}
@@ -182,6 +183,10 @@ class DateField extends React.Component {
 								dateFormat="uuuuuu-MM-dd"
 								disabled={!isCommonEraDate}
 								dropdownMode="select"
+								popperModifiers={
+									{preventOverflow: {
+									  enabled: true
+									}}}
 								popperPlacement="top-end"
 								selected={isValid(selectedDate) ? selectedDate : null}
 								timeFormat="false"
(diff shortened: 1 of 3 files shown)


src/client/stylesheets/style.less
```diff
@@ -339,6 +339,11 @@ a.contact-text:visited {
 
 /* react-datepicker styling */
 .form-group {
+	.res-dob{
```
MonkeyDo, line 342: A more descriptive name for this class would be better. I have no idea what this means 😅 
res(ponsive)-somethingsomething?

src/client/entity-editor/common/new-date-field.js
```diff
@@ -140,10 +140,11 @@ class DateField extends React.Component {
 					groupClassName={groupClassName}
 					label={labelElement}
 				>
-					<InputGroup style={{width: '18em'}}>
+					<InputGroup className="res-dob">
 						<FormControl
 							maxLength={isCommonEraDate ? 4 : 5}
 							placeholder="YYYY"
+							style={{width: '4em'}}
```
MonkeyDo, line 147: It was a good move to remove the inline style from the InputGroup, let's make sure we continue with the cleaner solution and put this styling in the css file instead of inlined

src/client/entity-editor/common/new-date-field.js
```diff
@@ -182,6 +183,10 @@ class DateField extends React.Component {
 								dateFormat="uuuuuu-MM-dd"
 								disabled={!isCommonEraDate}
 								dropdownMode="select"
+								popperModifiers={
+									{preventOverflow: {
```
MonkeyDo, line 187: Good find ! ⭐

MonkeyDo (commented):
Great, thanks a lot !
Fixes are working fine.

src/client/stylesheets/style.less
tr1ten: responsive date of birth ...

src/client/entity-editor/common/new-date-field.js
tr1ten: what about other FormControl components , should i shift their inline styles to style.less as well?

MonkeyDo: If they have a classname that allows them to be targeted with CSS, sure, why not.

src/client/stylesheets/style.less
MonkeyDo: I'd rather see the long form .responsive-date-field or something rather than the cryptic .res-dob

MonkeyDo (approved):
Thank a lot for these fixes !

PR #733, opened 2021-12-03
Fix(BB-610): Using query params to show correct label in dropdown

tr1ten:
### Problem

BB-610: In collections page, entity type URL parameters don't work on refresh or navigation

### Solution

use query params to show correct label in dropdown.

### Areas of Impact

1. src/client/components/pages/collections.js
2. src/client/components/pages/parts/collections-table.js
3.  src/server/routes/collections.js

### DIFF
--- a/src/client/components/pages/collections.js
+++ b/src/client/components/pages/collections.js
@@ -26,10 +26,10 @@ class CollectionsPage extends React.Component {
 	constructor(props) {
 		super(props);
 		this.state = {
-			querySearchParams: '',
+			querySearchParams: this.props.type,
 			results: this.props.results
 		};
-
+		// console.log('here type in page', this.props.type);
 		this.handleTypeChange = this.handleTypeChange.bind(this);
 		this.searchResultsCallback = this.searchResultsCallback.bind(this);
 		this.paginationUrl = './collections/collections';
@@ -56,6 +56,7 @@ class CollectionsPage extends React.Component {
 					showOwner={this.props.showOwner}
 					showPrivacy={this.props.showPrivacy}
 					tableHeading={this.props.tableHeading}
+					type={this.props.type ? this.props.type.split('=')[1] : ''}
 					user={this.props.user}
 					onTypeChange={this.handleTypeChange}
 				/>
@@ -86,6 +87,7 @@ CollectionsPage.propTypes = {
 	showPrivacy: PropTypes.bool,
 	size: PropTypes.number,
 	tableHeading: PropTypes.string,
+	type: PropTypes.string,
 	user: PropTypes.object
 };
 CollectionsPage.defaultProps = {
@@ -98,6 +100,7 @@ CollectionsPage.defaultProps = {
 	showPrivacy: false,
 	size: 20,
 	tableHeading: 'Collections',
+	type: '',
 	user: null
 
 };
--- a/src/client/components/pages/collections.js
+++ b/src/client/components/pages/collections.js
@@ -45,6 +45,13 @@ class CollectionsPage extends React.Component {
 		this.setState({querySearchParams, type});
 	}
 
+	searchParamsChangeCallback = (searchParms) => {
+		const type = searchParms.get('type') ? searchParms.get('type') : '';
+		if (type !== this.state.type) {
+			this.setState({querySearchParams: `?${searchParms.toString()}`, type});
+		}
+	};
+
 	render() {
 		return (
 			<div id="pageWithPagination">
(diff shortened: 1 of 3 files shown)


MonkeyDo (commented):
Looking good so far !

We fixed one thing which is great.
I just realized we also broke something else XD

If you change the entity type a couple of times using the dropdown, then use your browser navigation arrows to go back a couple of times, you'll see the entity type changes in the URL but not on the page/dropdown.

We need to add something to have the CollectionsPage interact with the PagerElement changes.
In short, when PagerElement reacts to a browser navigation event, it needs to send the new query parameters back up the the parent CollectionsPage which will set it in its state (the same way we do with CollectionsTable's `onTypeChange`)

There is already a mechanism for this: PagerElement accepts an optional `searchParamsChangeCallback` prop:
https://github.com/bookbrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/pager.js#L81-L83
13:05:38 

CollectionsPage will  need a new method that accepts those searchParams sent back by PagerElement, and extract the `type` to set it in the CollectionsPage state.

`searchParamsChangeCallback` is called with a [`URLSearchParams` object](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), and in CollectionsPage we can simply do `type = searchParams.get("type") ?? ""` (get the type if it exists, default to empty string otherwise) and set that on the state.

tr1ten:
Oh glad we discover it early ,  i'ma fix that real quick !

src/client/components/pages/collections.js
```diff
@@ -26,10 +26,11 @@ class CollectionsPage extends React.Component {
...
 		super(props);
 		this.state = {
-			querySearchParams: '',
-			results: this.props.results
+			querySearchParams: props.type ? `type=${props.type}` : '',
+			results: this.props.results,
+			type: props.type
 		};
-
+		// console.log('here type in page', this.props.type);
```
MonkeyDo, line 33: Some clean up here

src/client/components/pages/collections.js
```diff
@@ -41,9 +42,16 @@ class CollectionsPage extends React.Component {
 
 	handleTypeChange(type) {
 		const querySearchParams = type ? `type=${type}` : '';
-		this.setState({querySearchParams});
+		this.setState({querySearchParams, type});
 	}
 
+	searchParamsChangeCallback = (searchParms) => {
+		const type = searchParms.get('type') ? searchParms.get('type') : '';
```
MonkeyDo, line 49:
```suggestion
		const type = searchParms.get('type') ?? '';
```

The `??` operator means "if left side is null or undefined, return the right side", which basically means a default value.
In this case, searchParams.get will return either the parameter or `null` if there isn't one, so it's exactly appropriate (your code works, just suggesting a more succinct approach)

MonkeyDo (changes requested):
This is working great!
Thanks again for the good work, it was nice to get to the bottom of the issue(s) together.

src/client/components/pages/collections.js
tr1ten: ?? operator would be perfect for this , thanks !

src/client/components/pages/collections.js
```diff
@@ -41,9 +41,16 @@ class CollectionsPage extends React.Component {
...
 		const querySearchParams = type ? `type=${type}` : '';
-		this.setState({querySearchParams});
+		this.setState({querySearchParams, type});
 	}
 
+	searchParamsChangeCallback = (searchParms) => {
+		const type = searchParms.get('type') ?? '';
+		if (type !== this.state.type) {
+			this.setState({querySearchParams: `?${searchParms.toString()}`, type});
+		}
```
MonkeyDo, line 51: I forgot about this little detail, sorry:

setState has a lot of smarts built-in, and won't re-render the page if the values haven't changed. So you should be able to simply do this and not worry about whether type has changed, if the rest of the search paramas have changed or not, etc. :


```suggestion
		this.setState({querySearchParams: `?${searchParms.toString()}`, type});
```

tr1ten: Just to be sure , won't it cause the PagerElement to render twice ? , like it will also run setState when querySearchParms changes ( cause of modified from ,size parameter) which change on next and back button.

MonkeyDo: Fair enough, I'm not 100% sure of what would happen. 😅 
I don't think it would be a big impact, but might as well keep the condition here. :+1:

MonkeyDo (approved):

MonkeyDo (approved):
Works great, thanks !

PR #734, opened 2021-12-05
Fix[BB-635]: adding check for deleted entity

tr1ten:
### Problem

BB-635: Website crashes due to not handling already deleted entity case.  

### Solution

return 409 response code

### Areas of Impact

1. src/server/routes/entity/entity.tsx


MonkeyDo (approved):
Thank you ! :rocket:

PR #737, opened 2021-12-07
[BB-622]: Remove dimensions if format is eBook

the-good-boy:
### Problem

Physical data section should be disabled if eBook is chosen


MonkeyDo:
Thank you ! :ship:

PR #739, opened 2021-12-07
Homepage and Footer Upgrade 

akshaaatt:
- Removed Bookshelf.jpg and replaced it with MeB gradient style.
- Footer upgraded as per the other MeB projects.


MonkeyDo (commented):
Thanks for getting on that, and sorry I'm only looking at it now 😅 

I still love the footer. However it looks like all the images are missing:
image

Regarding the gradient, we have a dark overlay over that section that changes what I think is the original intent.
Without the dark overlay this is what it looks like:
image
Makes it pretty hard to read.


I have a suggestion that keeps the background image we previously had but overlay the gradient with a multiply blend mode, something like that, what do you think?
I still think there should be a dark shadow under the "BookBrainz" text to make it more readable.
image

akshaaatt:
I must have left something out here maybe! Will fix this. Thanks for checking!

akshaaatt:
> Regarding the gradient, we have a dark overlay over that section that changes what I think is the original intent. Without the dark overlay this is what it looks like: image Makes it pretty hard to read.

Agreed. In the latest commit, I improved the gradient situation. Let me know if that works fine for you. I believe we should get rid of an image background since the planned MusicBrainz revamp also follows the same.

0PrashantYadav0:
i would love to make changes @MonkeyDo

PR #740, opened 2021-12-10
BB-528: Index author in ElasticSearch

tr1ten:
### Problem

[BB-528](https://tickets.metabrainz.org/browse/BB-528) ticket for adding author name for work 

### Solution

index the author in ES and replace author with disambiguation for work having author

Before
before

After
after

### Areas of Impact

1. generateIndex to index author property.
2. dropdown related component for relationship.


src/common/helpers/search.js
```diff
@@ -374,8 +384,22 @@ export async function generateIndex(orm) {
...
-	for (const entityList of entityLists) {
+	for (const [index, entityList] of entityLists.entries()) {
 		const listArray = entityList.toJSON();
+		if (index === (entityLists.length - 1)) {
+			for (const workEntity of listArray) {
+				if (workEntity.relationshipSetId) {
+					for (const relationship of workEntity.relationshipSet.relationships) {
+						if (relationship.typeId === 8) {
+							// eslint-disable-next-line no-await-in-loop
+							const source = await getEntityWithAlias(orm, relationship.sourceBbid, 'Author');
```
MonkeyDo, line 395: Considering we already load all the Authors, I wonder if we can find the relevant author by bbid from the entityLists instead of loading it again from the DB

src/common/helpers/search.js
```diff
@@ -374,8 +384,22 @@ export async function generateIndex(orm) {
 	const entityLists = await Promise.all(behaviorPromise);
 
 	const listIndexes = [];
-	for (const entityList of entityLists) {
+	for (const [index, entityList] of entityLists.entries()) {
 		const listArray = entityList.toJSON();
+		if (index === (entityLists.length - 1)) {
```
MonkeyDo, line 389: It would be better and more future-proof to check by entity type name rather than position in an array.
If the order changes in the future this will break.

src/client/entity-editor/common/entity.tsx
```diff
@@ -41,7 +42,8 @@ function Entity(
 			}
 			{nameComponent}
 			{
-				disambiguation &&
+				author ? <span className="disambig margin-left-0-3"><small>({author})</small></span> :
+					disambiguation &&
```
MonkeyDo, line 46: It's not going to be ideal display-wise, but it would be best to show both the disambiguation followed by the author name.

src/client/entity-editor/common/linked-entity.tsx
```diff
@@ -91,7 +91,8 @@ class LinkedEntity extends React.Component<any, any> {
 				{nameComponent}
 				&nbsp;&nbsp;
 				{
-					disambiguation &&
+					author ? <span className="disambig"><small>({author})</small></span> :
+						disambiguation &&
```
MonkeyDo, line 95: Same comment as above, we want to show the disambiguation.
I'm not 100% certain we want to show the author in this component.

I wonder how that will look like considering one of the main relationships we show for works is "author wrote work".
This is what I envision which doesn't add much:
image

MonkeyDo (commented):
Thanks for working on that!
I'm looking forward to this feature, i think it's very helpful.

I was entering a book just yesterday and wanted that exact feature !

src/common/helpers/search.js
tr1ten: I doubt iterating over all authors would be better than single db query.

src/client/entity-editor/common/entity.tsx
tr1ten: How about name -author (disambiguation)?

MonkeyDo:
I took the liberty to finish things up, better late than never !
Here's where I got to for showing the author in the autocomplete search results, keeping the disambiguation as it was before and styling the author names:
image

And as for the search page, authors are now used as search terms for work entities, which is super useful. For example I searched for "lovecraft" and I get both results with lovecraft in the name, but also work entities with the "written by" relationship toH. P. Lovecraft.
We display the authors in the same way as above:
image

PR #749, opened 2021-12-23
Maintenance updates

leftmostcat:
### Problem

Several dependencies are out-of-date and have transitive dependencies with security issues.

### Solution

Update `swagger-ui-express`, `connect-redis`, and `log-node`.

### Areas of Impact

App connectors.


package.json
```diff
@@ -73,6 +73,7 @@
     "react-sortable-hoc": "^2.0.0",
     "react-sticky": "^6.0.1",
     "react-virtualized-select": "^3.0.1",
+    "redis": "^3.1.2",
```
leftmostcat, line 76: `connect-redis` v4+ no longer bundle `redis`.

src/api/app.js
```diff
@@ -49,6 +50,10 @@ app.use(express.json());
 app.use(express.urlencoded({extended: false}));
 app.use(compression());
 
+const redisClient = createClient({
```
leftmostcat, line 53: Due to `connect-redis` no longer bundling `redis`, we now have to create the client ourselves and pass it in, here and in the server app.

MonkeyDo (approved):
Thanks a lot ! I do like dropping the requirement for redis in dev.

We should have a follow-up PR that removes redis from the docker-compose file (we use it for local dev but not for prod) and associated docs.

PR #752, opened 2021-12-30
fix[BB-637]: Fixing typo in variable name?

tr1ten:
### Problem

BB-637: When editing edition width, height, depth, and weight fields are locked

### Solution

I suppose there was typo in variable name  so renamed physicalVisible -> physicalEnable

### Areas of Impact

src/server/routes/entity/edition.js


MonkeyDo (approved):
Thanks for finding this !
Such an easy fix, but the issue is hard to find in the first place !

PR #754, opened 2022-01-03
fix(CollectionPage): Added Horizontal rule in CollectionPage

the-good-boy:
### Problem

The Collection Page Name Section does not have a horizontal rule between Collection Title and Collection Attributes. This makes the design of the page appear somewhat inconsistent.
Here's how the Collection Page looks like: 

Screenshot from 2022-01-03 21-38-20

Here's how other entities look like: 

Screenshot from 2022-01-03 21-57-22

Here's how the Collection Page should look like:

Screenshot from 2022-01-03 21-57-17

### Solution
Added a Horizontal Rule between the title and attributes of the Collection in collection page

### DIFF
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -232,6 +232,7 @@ class CollectionPage extends React.Component {
 					</Col>
 					<Col md={10}>
 						<h1>{this.props.collection.name}</h1>
+						<hr/>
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Thanks ! I hadn't even realized it was missing, definitely looks better now !

PR #757, opened 2022-01-12
Fix(BB-636): displaying show more button only when necessary

tr1ten:
### Problem

BB-636: Show more link should be hidden in small annotations.

### Solution

Compare span height with parent container for toggling show more button.

### Areas of Impact

1. src/client/components/pages/entities/annotation.js

### DIFF
--- a/src/client/components/pages/entities/annotation.js
+++ b/src/client/components/pages/entities/annotation.js
@@ -25,10 +25,24 @@ import React from 'react';
 class EntityAnnotation extends React.Component {
 	constructor(props) {
 	  super(props);
-
+	  this.preRef = React.createRef();
 	  this.state = {
-			open: false
+			open: false,
+			show: true
 	  };
+	}
+
+
+	componentDidMount() {
+		const {annotation} = this.props.entity;
+		if (!annotation || !annotation.content) {
+			return;
+		}
+		const spanElement = document.querySelector('pre span');
+		if (spanElement.offsetHeight < this.preRef.current.offsetHeight) {
+			// eslint-disable-next-line react/no-did-mount-set-state
+			this.setState({open: false, show: false});
+		}
 	}
 
 	handleToggleCollapse = () => {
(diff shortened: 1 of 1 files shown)


src/client/components/pages/entities/annotation.js
```diff
@@ -25,12 +25,26 @@ import React from 'react';
...
 	  };
 	}
 
+
+	componentDidMount() {
+		const {annotation} = this.props.entity;
+		if (!annotation || !annotation.content) {
+			return;
+		}
+		const spanElement = document.querySelector('pre span');
```
MonkeyDo, line 41: It would probably be better to target the element with more specificity, like using the `.annotation-content` class.
Maybe this?
```suggestion
		const spanElement = document.querySelector('.annotation-content span');
```

src/client/components/pages/entities/annotation.js
```diff
@@ -25,12 +25,26 @@ import React from 'react';
 class EntityAnnotation extends React.Component {
 	constructor(props) {
 	  super(props);
-
+	  this.preRef = React.createRef();
 	  this.state = {
-			open: false
+			open: false,
+			show: true
```
MonkeyDo, line 31: For more clarity:

```suggestion
			showButton: true
```

src/client/components/pages/entities/annotation.js
```diff
@@ -25,12 +25,26 @@ import React from 'react';
 class EntityAnnotation extends React.Component {
 	constructor(props) {
 	  super(props);
-
+	  this.preRef = React.createRef();
```
MonkeyDo, line 28:
```suggestion
	  this.annotationContentRef = React.createRef();
```

MonkeyDo (changes requested):
Nice, that's a straightforward way to solve the issue!
Thanks for fixing this, I've been annoyed by it for a while !
A couple of suggestions below:

MonkeyDo:
OK, the code looks great.
I noticed something, a small detail but it should be easy to fix:

When we have a text block that is supposed to be collapsed, we show a white gradient to cover the bottom and suggest there is more to show.
This is done with CSS:
image

What we're going to want to do is very simple:
In your componentDidMount function, if there is enough height to show all the text, then set `open` to `true` in the state, instead of false:
image

MonkeyDo (approved):
Great, thank you !

PR #782, opened 2022-02-10
Fix(BB-643): Setting initialState in correct place

tr1ten:
### Problem

BB-643: Beta: "Add edition" is not pre-populated from work.

### Solution

setting initialState in correct place

### Areas of Impact

src/server/routes/entity/edition.js


MonkeyDo (approved):
Thanks for the fix !

PR #785, opened 2022-02-25
Fix(BB-644): Auto append new isbn identifier row

tr1ten:
### Problem

BB-644: Automatically generate ISBN-13 from ISBN-10 and vice versa

### Solution

append new identifier row on adding any isbn

### Areas of Impact

common utils, identifier row component


MonkeyDo (commented):
Looks sane at a quick glance.

I'd love to see unit tests for the conversion methods, and throwing all the valid and invalid ISBN formats we can think of

MonkeyDo (approved):
Looks great, thanks !

PR #789, opened 2022-03-01
fix(bs4): Place tooltips to the right in modal header

MonkeyDo:
### Problem
Tooltips are rendered by default on top of their OverlayTrigger, and for modal headers the tooltips appear mostly off-screen:
image

### Solution
Adds `placement="right"` property to OverlayTrigger components that appear in modal headers.
Other OverlayTriggers remain untouched as we don't have the issue mentioned above.

image

### Areas of Impact
```
src/client/components/forms/deletion.js
src/client/entity-editor/alias-editor/alias-editor.js
src/client/entity-editor/identifier-editor/identifier-editor.js
```

### DIFF
--- a/src/client/components/forms/deletion.js
+++ b/src/client/components/forms/deletion.js
@@ -185,6 +185,7 @@ class EntityDeletionForm extends React.Component {
 											<OverlayTrigger
 												delay={50}
 												overlay={deletionTooltip}
+												placement="right"
 											>
 												<FontAwesomeIcon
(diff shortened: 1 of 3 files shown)


MonkeyDo:
Last fixes before deploying the Bootstrap4 update to prod, along with https://github.com/LordSputnik/lobes/pull/1

LordSputnik:
Looks good to me! I think I did something like this locally at one point, but didn't keep it for some reason...

PR #790, opened 2022-03-03
React-Select v4 Upgrade

tr1ten:
upgrading react-select from v1 to v4

### Areas of Impact
all components using react-select


src/client/components/input/select-wrapper.js
```diff
@@ -86,7 +86,12 @@ class SelectWrapper extends React.Component {
 		const Child = base;
 
 		const childValue = _.isNil(value) ? this.state.value : value;
-
+		function getOptionLabel(option) {
```
LordSputnik, line 89: If you use `getOptionLabel = (option) => { ... }` instead, then `getOptionLabel={this.getOptionLabel}`, you don't get the react/jsx-no-bind warning

src/client/entity-editor/common/entity-search-field-option.js
```diff
@@ -138,14 +137,19 @@ class EntitySearchFieldOption extends React.Component {
...
+			return option.text;
+		}
 		const wrappedSelect = (
 			<ImmutableAsyncSelect
+				className="Select"
+				classNamePrefix="react-select"
+				components={{Option: LinkedEntity, SingleValue: Entity}}
 				filterOptions={false}
+				// eslint-disable-next-line react/jsx-no-bind
+				getOptionLabel={getOptionLabel}
```
LordSputnik, line 150: Can avoid using arrow func, as in select-wrapper

src/client/components/input/entity-search.js
```diff
@@ -18,7 +18,8 @@
  */
 
 import {entityToOption, genEntityIconHTMLElement} from '../../helpers/entity';
-import {Async} from 'react-select';
+// eslint-disable-next-line import/no-internal-modules
```
LordSputnik, line 21: Let's configure the rule in `/.eslintrc.js` to allow imports from `**/react-select/*` - see example here:

https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-internal-modules.md

src/client/stylesheets/style.scss
```diff
@@ -28,7 +28,6 @@ $disabled-color: lighten($disabled-bg,15%);
 
 @import "lobes4/scss/bootstrap.scss";
 
-@import "react-select/scss/default.scss";
 @import "react-virtualized/styles.css";
```
LordSputnik, line 31: Can probably get rid of these too, once LanguageField is migrated.

LordSputnik (commented):
Nice work so far! Nothing suprising here really, it's along the same lines as my change. Added a few comments here and there as suggestions

MonkeyDo (commented):
Server crashed when I try to edit my profile locally, with this error:
```
TypeError: Cannot read properties of null (reading 'unlockId')
    at new SelectWrapper (bookbrainz-site/lib/client/components/input/select-wrapper.js:75:38)
    ```

File: src/client/entity-editor/relationship-editor/relationship.tsx [hunk 5]
```diff
@@ -52,9 +53,8 @@ type RelationshipProps = {
 	relationshipType: RelationshipType
 };
 
-function Relationship({
-	contextEntity, link, relationshipType, sourceEntity, attributes, showAttributes, targetEntity
-}: RelationshipProps) {
+function Relationship(props: SingleValueProps<Relationship> | OptionProps<Relationship, any> | {data:Relationship}) {
```
MonkeyDo, line 56: Same as other places, we want to avoid nesting props into an object

src/client/entity-editor/relationship-editor/relationship.tsx
```diff
@@ -74,29 +74,42 @@ function Relationship({
...
+			<Entity data={targetObject}/>
+			{' '}
+			<RelationshipAttribute attributes={attributes} showAttributes={showAttributes}/>
+		</div>
+	);
+	let parentContainer;
+	if (props.getStyles) {
+		parentContainer = props.innerProps ? <components.Option {...props as OptionProps<any, any>}> {child}</components.Option> : props.getStyles &&
+		<components.SingleValue {...props}>{child}</components.SingleValue>;
+	}
```
MonkeyDo, line 91: I can't tell what that does, can you explain?
It would be good to have some comments here.

I'm not sure we should be tying this component to react-select, since it doesn't use it directly
, and it is also use with nothing to do with react-select (for example the EntityRelationships component)

Where are props.getStyles and props.innerProps from? They're not mentioned anywhere else in the component

src/client/stylesheets/style.scss
```diff
@@ -282,15 +280,30 @@ a.contact-text:visited {
...
+	background-color: white !important;
+	&:hover{
+		background-color: #deebff !important;
+		color: #333;
+	}
+}
+.react-select__menu {
+	z-index: 100 !important;
+}
+.react-select__control, .Select-control{
```
MonkeyDo, line 294: It would be interesting to see why the `!important` is needed here.
What are they being overwritten by? Can we supersede them simply by having more selectors instead?
Usually it's not a good sign if we have to use !important and its use should be limited as much as possible.

src/client/entity-editor/relationship-editor/relationship.tsx
tr1ten: Sorry for the mess, these are the [props](https://react-select.com/props#components) react-select pass to custom components. since we are using same component for value, option and other purposes i have to do this to ensure it work in all cases. i will add some comments about the same

src/client/entity-editor/relationship-editor/relationship-editor.tsx
```diff
@@ -405,17 +406,28 @@ class RelationshipModal
 			// Name of the attribute type belonging to the relationship type. EX: ['position', 'number]
 			attributes = attributeTypes.map(attribute => attribute.name);
 		}
+		const colourStyles = {
+			option: (styles) => ({
+				...styles,
+				backgroundColor: 'transparent',
+				color: 'inherit'
+			  })
+		  };
```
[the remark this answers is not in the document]
tr1ten: This is not needed since we are using classes now to style components, i used it as promoted by [react-select](https://react-select.com/upgrade-to-v2#new-styles-api)  but later realized classes would be more convenient here.

src/client/stylesheets/style.scss
tr1ten: Since react-select use css in js to style its component, i don't think we can get rid of !important flag.

src/client/entity-editor/relationship-editor/relationship.tsx
```diff
@@ -52,9 +53,8 @@ type RelationshipProps = {
 	relationshipType: RelationshipType
 };
 
-function Relationship({
-	contextEntity, link, relationshipType, sourceEntity, attributes, showAttributes, targetEntity
-}: RelationshipProps) {
+function Relationship(props: SingleValueProps<Relationship> | OptionProps<Relationship, any> | Relationship) {
+	const {contextEntity, link, relationshipType, sourceEntity, attributes, showAttributes, targetEntity} = props.data ?? props;
```
MonkeyDo, line 57: It still looks wrong to me that props would be passed nested inside a `data` object.
We should avoid doing that, it's an anti-pattern.
Why is this necessary?

src/client/entity-editor/relationship-editor/relationship.tsx
MonkeyDo: To me it looks like we shouldn't be using this component directly, if all these changes are required in order to make it work for use in the select.
The reason is that this component isn't specific to react-select and is used in other contexts. It shouldn't have to know anything about react-select.

Even with the comment you added, I can't tell what is happening in this code block because I don't know how the new version of react-select works; that's a red flag: we're tying this component to a specific version of a library.


I think it is probably time to have a RelationshipOption component specifically for use with react-select, that will manage all these implementation details and rely on the unmodified Relationship component internally for displaying.

Does that seem reasonable to you? As I said I'm unclear on react-select details, but happy to help.

src/client/entity-editor/common/linked-entity.tsx
```diff
@@ -77,13 +79,15 @@ class LinkedEntity extends React.Component<any, any> {
 	}
 
 	render() {
-		const option = this.getSafeOptionValue(this.props.option);
+		const option = this.getSafeOptionValue(this.props.data);
+		const styles = this.props.getStyles('option', this.props);
```
MonkeyDo, line 83: This is going to throw an error if getStyles is not defined.
For example when used in `src/client/entity-editor/edition-section/edition-section.tsx`.

Another case where we're tying a component to react-select implementation details when we shouldn't be.
I'd rather see a separate component for use with react-select.

tr1ten: I also believe we might as well have separate components for react-select.

PR #791, opened 2022-03-03
fix(bs4): Make exactMatches ListGroupItems links

MonkeyDo:
### Problem
As reported [in the forums](https://community.metabrainz.org/t/bug-after-changes/575062) the possible duplicate entities are not clickable anymore.
image

introduced in the Bootstrap v4 migration: 30518cacd058d63f09d54d1b5fff625c254ce294

### Solution
Simply adding the `action` boolean prop turns the list group item into an anchor tag.
From [react-bootstrap docs](https://react-bootstrap-v4.netlify.app/components/list-group/):
>Toggle the action prop to create actionable list group items, with disabled, hover and active styles. List item actions will render a <button> or <a> (depending on the presence of an href) by default 

This component is the only ListGroup.Item that has an `href` and that we use as a link.

### Areas of Impact
src/client/entity-editor/name-section/name-section.js

### DIFF
--- a/src/client/entity-editor/name-section/name-section.js
+++ b/src/client/entity-editor/name-section/name-section.js
@@ -185,6 +185,7 @@ class NameSection extends React.Component {
 									{exactMatches.map((match) =>
 										(
 											<ListGroup.Item
+												action
 												href={`/${_.kebabCase(entityType)}/${match.bbid}`}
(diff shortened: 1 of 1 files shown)


LordSputnik (approved):
Simple - looks good to me!

PR #792, opened 2022-03-04
fix(static): Update entity-relationships diagram on the Help Page

the-good-boy:
The [Help Page](https://bookbrainz.org/help) must briefly describe the new Series entity. The entity-relationships diagram should also be updated.
I have updated the diagram as follows:
entity_relationships


MonkeyDo (approved):
Thank you so much ! 
This has been bugging me, but I wasn't finding the time to fix it :)

PR #793, opened 2022-03-04
fix(Collection): conditionally render author name column in the Work Table component

the-good-boy:
### Problem

If authorsData is undefined (in case of collections, it will be undefined), the author name column should not be rendered.

### Solution

check if authorData is defined before trying to render it in the WorkTableRow component


MonkeyDo (approved):
Thanks for the quickfix !

As discussed we're gonna want to think about how to refactor the existing code to show authors on collection pages workTables (for work collections), but now we can take our time :)

PR #799, opened 2022-03-08
feat(LanguageField): Index languages for search

MonkeyDo:
With upcoming work to improve the select dropdown used in the LanguageField component, we need to index languages in ElasticSearch.

### Areas of Impact
src/common/helpers/search.js


LordSputnik (approved):
Looks good to me, simple and can't see anything that would break.

MonkeyDo:
This should not be necessary now, as #790 fixes the immediate issue with the searchable select for languages (older browsers struggle with large lists of options [not related to virtualization, it's a filtering issue])

PR #802, opened 2022-03-10
using word boundary in addHttpRegex

tr1ten:
### Problem

Fixing broken links in annotation section

### Solution

use word boundary rather than space in adding http prefix

### Areas of Impact

client utils


MonkeyDo (approved):
Thanks ! 💯

PR #805, opened 2022-03-13
feat(Collections): added AuthorName column in Collection of Works

the-good-boy:
In case of a collection of works, we need to load the authors of the works in order to display then in the WorkTable component. Reused some of the functions and logic we used in the loadWorkTableAuthors middleware in this [PR](https://github.com/metabrainz/bookbrainz-site/pull/617).

### DIFF
--- a/src/server/routes/collection.js
+++ b/src/server/routes/collection.js
@@ -23,7 +23,9 @@ import * as handler from '../helpers/handler';
 import * as middleware from '../helpers/middleware';
 import * as propHelpers from '../../client/helpers/props';
 import * as search from '../../common/helpers/search';
+import _ from 'lodash';
 import {escapeProps, generateProps} from '../helpers/props';
+import { addAuthorsDataToWorks, getListofBBIDs } from '../../client/helpers/entity';
 import CollectionPage from '../../client/components/pages/collection';
 import Layout from '../../client/containers/layout';
 import React from 'react';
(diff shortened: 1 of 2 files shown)


src/server/routes/collection.js
```diff
@@ -132,11 +134,16 @@ router.get('/:collectionId', auth.isAuthenticatedForCollectionView, async (req,
...
 			...await orm.func.entity.getEntity(orm, collection.entityType, item.bbid, relations)
 		}));
-		const entities = await Promise.all(entitiesPromise);
+		let entities = await Promise.all(entitiesPromise);
 		const isOwner = req.user && parseInt(collection.ownerId, 10) === parseInt(req.user?.id, 10);
 		const isCollaborator = req.user && collection.collaborators.filter(collaborator => collaborator.id === req.user.id).length;
 		const userId = req.user ? parseInt(req.user.id, 10) : null;
-
+		if (collection.entityType === 'Work') {
+			const workBBIDs = getListofBBIDs(entities);
```
MonkeyDo, line 142: I don't know if we need a utility function for that, it should be as simple as 
```suggestion
			const workBBIDs = entities.map(entity => entity.bbid);
```

MonkeyDo (changes requested):
Thanks, good continuation of the previous PR !

src/server/routes/collection.js
the-good-boy: you're right, my bad.

MonkeyDo (approved):
Looks good, thanks a lot !

PR #807, opened 2022-03-15
Fix(editor): Add comparision for genderId

tr1ten:
while working on react-select in profile, i realized changing gender in editor doesn't considered as change thus giving no change error, i added check for the same in editor route.


MonkeyDo (approved):
Well done spotting this !

PR #813, opened 2022-03-16
fix(language-select): Fix filtering of language select options

MonkeyDo:
### Problem
The select is crazy slow on older browsers due to issues with filtering the gigantic list of options.
The default filtering of react-select is inefficient

### Solution
Using the library react-select-fast-filter-options fixes the issue.
It uses js-search under the hood which is more efficient.

### Areas of Impact
src/client/entity-editor/common/language-field.tsx
package.json and yarn lockfile

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,6 +71,7 @@
     "react-hot-loader": "^4.13.0",
     "react-redux": "^7.2.6",
     "react-select": "^1.1.0",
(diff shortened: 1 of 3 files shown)


tr1ten:
seems good to me :)

tr1ten (approved):

PR #815, opened 2022-03-17
Fix(BB-651): Error handling for not found editor

tr1ten:
### Problem

Since function execution doesn't stop after next(err) thus causing issues

Relevant ticket BB-651: 502 when trying to load an editor by name

### Areas of Impact

editor route

### DIFF
--- a/src/server/routes/editor.js
+++ b/src/server/routes/editor.js
@@ -257,30 +257,26 @@ function achievementColToEditorGetJSON(achievementCol) {
 	};
 }
 
-router.get('/:id', (req, res, next) => {
+router.get('/:id', async (req, res, next) => {
 	const {AchievementUnlock, Revision} = req.app.locals.orm;
 	const userId = parseInt(req.params.id, 10);
 	if (!userId) {
 		return next(new error.BadRequestError('Editor Id is not valid!', req));
 	}
-	const editorJSONPromise = getIdEditorJSONPromise(userId, req);
+	try {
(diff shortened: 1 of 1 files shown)


src/server/routes/editor.js
```diff
@@ -257,10 +257,12 @@ function achievementColToEditorGetJSON(achievementCol) {
...
 }
 
-router.get('/:id', async (req, res, next) => {
+router.get('/:id', (req, res, next) => {
 	const {AchievementUnlock, Revision} = req.app.locals.orm;
 	const userId = parseInt(req.params.id, 10);
-
+	if (!userId) {
+		return next(new error.BadRequestError('Editor Id is not valid!', req));
+	}
```
MonkeyDo, line 265: Yes, let's keep this !

MonkeyDo (changes requested):
We're slowly moving the codebase to use async/await wherever we can, so I don't want us to revert to using promise chains.

I believe what is missing is a `try {…} catch(err) {next(err)}`  block around the meat of the query, wrapping around the lines from `const editorJSONPromise`  to `res.send(target({…}));`

Finally in this route we should remove the instances of `.catch(next)`; since we wrap the whole thing, we want errors to bubble up to our main try-catch block and call `next` in the catch part.

Do let me know if I wasn't clear :)

tr1ten:
I wonder, in general what would be more efficient- awaiting on each promise or awaiting on promise.all, any difference performance wise?

MonkeyDo (approved):
Thank you !
That fixes that issue !

PR #822, opened 2022-03-26
Added area_id column to alias table

Asymmentric:
### Problem
BB-221 It should be possible to identify aliases by locale as well as language

#### Work Done
1. Added `area_id` column in `alias` table of `bookbrainz` database
2. Referenced the `area_id` column to `area` table of `musicbrainz` database
#### Work To be done
1. Adding an input drop-down field `Region/Country` in `Alias Editor` pop-up UI which displays the region name.
2. Reference the value of `area_id` to `Region/Country` field.
### Areas of Impact
1. Impact on UI - Addition of a drop-down field to `Alias Editor` pop-up.
2. Impact on Backend - Modification of `bookbrainz` database.


MonkeyDo:
Thanks for giving this a shot, but areas are not the same as locales.
The area table does not have any information on locales.

I would recommend reading up on the difference between country codes and locales, and seeing how MusicBrainz goes about storing them.

PR #827, opened 2022-03-30
BB-491: Show more information in merge queue

Devik225:
Right now only entity name appears in the merge queue list. To make it easy for the user we need a way to add more data.
I created one layout which is suitable for this query.

**This is how the card looks in normal view.**
Screenshot from 2022-03-30 00-45-22

**This is how the card looks in expanded view.**
Screenshot from 2022-03-30 00-45-25

If this works then we can add more information in this way.


MonkeyDo (changes requested):
Hello!
Thanks for submitting your PR :)
Sorry I'm reviewing this so late, I thought I left a review comment ages ago but I must have not posted it !

Your changes remove a file that is crucial for local development (config/config.json.example), so that commits/change will need to be reverted.


As for the feature, the main issue I see currently is that the EntityLink component is used across the website and not just for the merge queue preview, so your changes currently will affect unrelated areas.

I think considering we have a bit of space available, it would be nice to see *some* details under each entity name without having to click on a button to show more changes.

Not exactly sure what that would look like, but I guess we would move away from using a table format with columns, to preserve space.
Would you like to give it another try?

I can help with UI mockups otherwise if you would like.

MonkeyDo:
Closing for now due to inactivity, but again happy to help if you want to finish the PR up :)

PR #829, opened 2022-04-01
MEB-135: Standardizing MetaBrainz Logos

Aerozol:
Updating logos to current version: Taller caps, Roboto Black font


MonkeyDo (commented):
It looks like [BookBrainz_logo_short_vertical.svg](https://github.com/metabrainz/bookbrainz-site/pull/829/files#diff-417436886bdd9f62de75c521ec2c9eaa9dd69cae7cb4d0835f009039022b0ef3) and [BookBrainz_logo_small_short_vertical.svg](https://github.com/metabrainz/bookbrainz-site/pull/829/files#diff-bf135d04c38bc4c10dc71b9fb8fedd1d383c6883922959fec52fd865abe3af42) are the same as before.

We don't use any of the vertical logos or any of the logo_small versions anywhere on the website, so we could possibly even delete the files.
No reason to keep copies here if they're not used and will need to be maintained (like we're doing now)

Apart from that everything looks good, thank you !

Aerozol:
Thanks Monkey! Ah yes, I the vertical stack logos didn't get any changes, indeed.
BB was the only site that had all the logos, regardless of useage. But I updated all/these logo variations in the master logos folder anyway, so it was only an extra drag and drop.

PR #832, opened 2022-04-05
Fix(Identifier-Editor): Improve identifier editor behavior for ISBNs

tr1ten:
In this PR, I am making ISBN detection more strict by verifying its check digit, disabling switching type after user manually selected it and disabling new ISBN row addition(temporary).


MonkeyDo (approved):
Thank you !

PR #833, opened 2022-04-06
Fix(languages): Sort language options after fast filter

tr1ten:
### Problem

Languages are not sorted according to frequency after using fast-filter

### Solution

Sort options manually after filtering

### Areas of Impact

language-field

### DIFF
--- a/src/client/entity-editor/common/language-field.tsx
+++ b/src/client/entity-editor/common/language-field.tsx
@@ -24,6 +24,7 @@ import ValidationLabel from './validation-label';
 import VirtualizedSelect from 'react-virtualized-select';
 import createFilterOptions from 'react-select-fast-filter-options';
 import {faQuestionCircle} from '@fortawesome/free-solid-svg-icons';
+import {isNumber} from 'lodash';
 
 
 type Props = {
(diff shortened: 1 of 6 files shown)


src/client/entity-editor/common/language-field.tsx
```diff
@@ -61,6 +61,17 @@ function LanguageField({
 	const filterOptions = createFilterOptions({
 		options
 	});
+	const sortFilterOptions = (opts, input, selectOptions) => {
+		const newOptions = filterOptions(opts, input, selectOptions);
+		const sortLang = (a, b) => {
+			if (a.frequency !== b.frequency) {
```
MonkeyDo, line 67: I think we should guard against a potential missing frequency property, just in case.
Maybe using the isNumber utility from lodash?
```suggestion
			if (isNumber(a.frequency) && isNumber(B.frequency) && a.frequency !== b.frequency) {
```

MonkeyDo (commented):
That looks great, thanks.
That's exactly what I had in mind.

Tested with Firefox 44, works fast still. 👍

MonkeyDo (approved):
Great, I think this solves the last of our issues with the language field, nice work !
:rocket:

PR #834, opened 2022-04-06
fix(entity-editor): [BB-661] Remove slow deserialization in redux state

MonkeyDo:
https://community.metabrainz.org/t/lag-between-typing-and-appearance-of-information-in-annotation-section-of-editions/579426

Introduced in commit 7d02163, using an Immutable.List instead of a plain array to store `matchingNameEditionGroups` in the redux state introduced a  huge slowness in state deserializing which impacted the whole entity editor (when creating/editing Editions only).
The page slows down to a crawl if you trigger lots of state updates, for example writing text in the annotation textarea.
Reverting these changes to fix the issue.

### DIFF
--- a/src/client/entity-editor/edition-section/reducer.ts
+++ b/src/client/entity-editor/edition-section/reducer.ts
@@ -45,7 +45,7 @@ function reducer(
 	state: State = Immutable.Map({
 		format: null,
 		languages: Immutable.List([]),
-		matchingNameEditionGroups: Immutable.List([]),
+		matchingNameEditionGroups: [],
 		physicalEnable: true,
 		publisher: null,
 		releaseDate: '',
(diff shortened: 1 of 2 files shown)


src/client/entity-editor/edition-section/reducer.ts
```diff
@@ -85,9 +85,9 @@ function reducer(
 			return state.set('depth', payload);
 		case UPDATE_WARN_IF_EDITION_GROUP_EXISTS:
 			if (!Array.isArray(payload) || !payload.length) {
-				return state.set('matchingNameEditionGroups', Immutable.List());
+				return state.set('matchingNameEditionGroups', []);
```
tr1ten, line 88: I'm not sure but is it good idea to get rid of immutability here?
can't we just deserialize it in the component itself?

tr1ten (commented):
Apart from one doubt, i think we are good to go

src/client/entity-editor/edition-section/reducer.ts
MonkeyDo: The whole issue with the slowing down came from doing just that: https://github.com/metabrainz/bookbrainz-site/commit/7d02163d09246a27a159dbcdb3c6d54f60bfebb7#diff-2142be03733089165f1b3d5ce09055a7f3cd9a70a1559ac326acb9b8c57c26faL88

I didn't think it would have such a big impact, but obviously deserializing in the component is expensive (maybe even for an empty List).

I couldn't identify any upside in using a List over an array, so my original commit didn't make much sense in the first place.

PR #840, opened 2022-05-02
Remove hyphen from isbns

Asymmentric:
### Problem
When an ISBN is added to an edition, it can be stored with or without hyphens. This creates inconsistency across editions. It also creates problems when searching by ISBN; if the ISBN was added with hyphens and you search without hyphens in the proper positions, you won't find the edition. The best way is most likely to remove the hyphens when the identifier is validated and automatically remove them when searching.

### Solution
Remove `-`(hyphens) from ISBN identifier when constructing identifiers

### Areas of Impact
Store ISBN without hyphens to maintain consistency and ease  in search

### DIFF
--- a/sql/migrations/alias/up.sql
+++ b/sql/migrations/alias/up.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE bookbrainz.alias ADD area_id INTEGER;
+ALTER TABLE bookbrainz.alias ADD CONSTRAINT alias_area_id_fkey FOREIGN KEY (area_id) REFERENCES musicbrainz.area(id);
+
+COMMIT ;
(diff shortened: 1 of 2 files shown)


leftmostcat:
We should probably remove them for searching as a normalization process, but we should continue storing them in the database. ISBNs are structured; each segment represents a different piece of information, but we don't have enough information to add hyphens when they aren't already there.

For example, let's look at 978-0-615-44640-0. 978 is the prefix element, 0 is the registration group element, and we run into a problem at 615, the registrant element. This element can differ in length, as can the following publication element. Registrants are assigned by each country's designated ISBN agency, and there is no standard across all of them. Just for the English-speaking world, we'd need [this table](https://en.wikipedia.org/wiki/ISBN#English_language_pattern), and we would need one of those for every other country, assuming that each country even standardizes the length of the registrant element in such a way. It's not practical for us, and so we consider 9780615446400 acceptable in order to allow users to enter the information they have, but 978-0-615-44640-0 is preferred because it contains more information.

If this is reworked to add hyphen removal as part of search indexing and to remove hyphens from search terms, that would be an acceptable improvement.

PR #841, opened 2022-05-02
[BB-666]: Relationship Attributes should be shown on entity display pages too

the-good-boy:
### Problem

This PR addresses BB-666.


src/client/components/pages/entities/relationships.js
```diff
@@ -33,8 +33,10 @@ function EntityRelationships({contextEntity, relationships}) {
 					>
 						<Relationship
 							link
+							attributes={relationship.attributeSetId !== null ? relationship.attributeSet.relationshipAttributes : null}
 							contextEntity={contextEntity}
 							relationshipType={relationship.type}
+							showAttributes={relationship.attributeSetId !== null}
```
MonkeyDo, line 39: Let's make it a bit more lax, for example accept `undefined` as well:
(We don't expect an attributeSetId would ever be the number `0`, which would evaluate to false)

```suggestion
							showAttributes={Boolean(relationship.attributeSetId)}
```

MonkeyDo (commented):
Thanks for another good and simple PR :)

If I could ask a favor, in this PR or in another (as you prefer) could you also please refactor
the [`entity-editor/relationship-editor/relationship.tsx` ](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/entity-editor/relationship-editor/relationship.tsx#L90) so that instead of passing down the `showAttributes` prop down to the `RelationshipAttribute` component, we use it directly to optionally render the component like so?
```
{showAttributes &&
    <RelationshipAttribute attributes={attributes}/>
}
```

PR #842, opened 2022-05-03
Feat(EG):  Copy Edition name to Edition Group

tr1ten:
### Problem

BB-663: Add option to copy Edition name to Edition Group

### Solution

Added checkbox for copying edition name to edition-group

### Areas of Impact

nameSection component, createOrEditEnity function

### DIFF
--- a/src/client/entity-editor/name-section/actions.ts
+++ b/src/client/entity-editor/name-section/actions.ts
@@ -26,7 +26,7 @@ export const UPDATE_NAME_FIELD = 'UPDATE_NAME_FIELD';
 export const UPDATE_SORT_NAME_FIELD = 'UPDATE_SORT_NAME_FIELD';
 export const UPDATE_WARN_IF_EXISTS = 'UPDATE_WARN_IF_EXISTS';
 export const UPDATE_SEARCH_RESULTS = 'UPDATE_SEARCH_RESULTS';
-
+export const UPDATE_COPY_CHECKBOX = 'UPDATE_COPY_CHECKBOX';
 export type Action = {
 	type: string,
 	payload?: unknown,
@@ -201,5 +201,19 @@ export function searchName(
 					type: UPDATE_SEARCH_RESULTS
 				});
 			});
+	};
+}
+
+/**
+ * Produces an action indicating that the checkbox
+ * should be updated with the provided value.
+ *
+ * @param {boolean} isCheck - The new value to be used for the checkbox.
+ * @returns {Action} The resulting UPDATE_COPY_CHECKBOX action.
+ */
+export function updateCheckbox(isCheck:boolean): Action {
+	return {
+		payload: isCheck,
+		type: UPDATE_COPY_CHECKBOX
 	};
 }
--- a/src/client/entity-editor/name-section/name-section.js
+++ b/src/client/entity-editor/name-section/name-section.js
@@ -16,13 +16,14 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import {Alert, Col, ListGroup, Row} from 'react-bootstrap';
+import {Alert, Col, Form, ListGroup, Row} from 'react-bootstrap';
 import {
 	checkIfNameExists,
 	debouncedUpdateDisambiguationField,
 	debouncedUpdateNameField,
 	debouncedUpdateSortNameField,
 	searchName,
+	updateCheckbox,
 	updateLanguageField
 } from './actions';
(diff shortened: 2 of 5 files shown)


src/server/routes/entity/entity.tsx
```diff
@@ -1085,6 +1086,21 @@ export function handleCreateOrEditEntity(
 			let allEntities = [...otherEntities, mainEntity]
 				.filter(entity => entity.get('dataId') !== null);
 
+			// copy name to the edition group
+			if (!isNew && entityType === 'Edition' && body.copyToEG) {
+				const currentEditionGroup = currentEntity.editionGroup;
+				const entityDefaultAlias = body.aliases.find((alias) => alias.default);
+				if (currentEditionGroup.defaultAlias.name !== entityDefaultAlias.name) {
+					const tempBody = {aliases: [{...currentEditionGroup.defaultAlias, default: true, name: entityDefaultAlias.name}]};
```
MonkeyDo, line 1094: Reading this aliases array, I think if the EG before modification has multiple aliases, this tempBody will always be a different array of aliases, since it onyl contains one item.
You'll have to use the whole `currentEditionGroup.aliases` and modify the default alias (I guess by filtering by `id`).

I don't think we have access to the aliases array on currentEditionGroup and so the `const EGEntity = await fetchOrCreateMainEntity` call should probably be done beforehand.

MonkeyDo: About that last detail: I've suggested removing the `currentEditionGroup.defaultAlias.name !== entityDefaultAlias.name` check and letting `getNextAliasSet` do the comparison, but now I realize this adds some DB calls if we have to fetch the Edition Group first to get the aliases (and we don't want that).

I *think* we can get away with fetching all the aliases for the EG in makeEntityLoader: https://github.com/metabrainz/bookbrainz-site/blob/master/src/server/routes/entity/edition.js#L299 
`'editionGroup.defaultAlias'` -> `'editionGroup.aliases'` probably does the trick?

If none of this works then we'll want to keep an alias comparison condition, but make it similar to what `updateAliasSet` ORM function uses: https://github.com/metabrainz/bookbrainz-data-js/blob/master/src/func/alias.ts#L41-L48

src/client/entity-editor/name-section/actions.ts
```diff
@@ -26,7 +26,7 @@ export const UPDATE_NAME_FIELD = 'UPDATE_NAME_FIELD';
 export const UPDATE_SORT_NAME_FIELD = 'UPDATE_SORT_NAME_FIELD';
 export const UPDATE_WARN_IF_EXISTS = 'UPDATE_WARN_IF_EXISTS';
 export const UPDATE_SEARCH_RESULTS = 'UPDATE_SEARCH_RESULTS';
-
+export const UPDATE_COPY_CHECKBOX = 'UPDATE_COPY_CHECKBOX';
```
MonkeyDo, line 29: Same remark as my other comment, let's rename this `UPDATE_COPY_NAME_CHECKBOX` or `UPDATE_COPY_NAME_TO_EDITION_GROUP` to make way for `UPDATE_COPY_AUTHOR_CREDIT_…`

MonkeyDo (changes requested):
Nice going !
Thanks a lot, that's going to be very useful now that I have to implement a similar checkboc for copying Author Credits.
I might just copy your homework :p


When I first tried the feature, I got an error on submission: `Submission Error: No Updated Field`.
Having the checkbox as the only change in the form isn't considered a change to the Edition, which sort of makes sense, but I think we should allow this workflow.

src/client/entity-editor/name-section/name-section.js
```diff
@@ -172,6 +176,17 @@ class NameSection extends React.Component {
 							onChange={this.handleNameChange}
 						/>
 					</Col>
+					{action === 'edit' && entityType === 'edition' &&
+					<Col className="md-2" lg={{offset: 3, span: 6}}>
+						<Form.Check
+							className="margin-bottom-d8"
+							id="name"
+							label="Copy the edition name to edition-group"
```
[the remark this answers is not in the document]
tr1ten: > Maybe below the disambiguation?

This does sound like a reasonable place for checkbox

tr1ten:
> I think we should allow this workflow.

Not sure about this, if only thing user wants is to change EG's default name wouldn't it be better to do it directly?
else we would have to deal with cases like if they already have same default name? should we throw error or let them submit?
also if EG is the only thing that got changed, should we include Edition in the revision as well?

PR #843, opened 2022-05-21
Feat(editor): User account deletion endpoint

tr1ten:
### Problem

BB-260: Implement the user account deletion endpoint

### Solution

Added post route /delete-user to remove account from BB project

### DIFF
--- a/src/server/routes/auth.js
+++ b/src/server/routes/auth.js
@@ -18,6 +18,7 @@
 
 import express from 'express';
 import passport from 'passport';
+import request from 'superagent';
 import status from 'http-status';
 
 
@@ -72,5 +73,46 @@ router.get('/logout', (req, res) => {
 	req.logOut();
 	res.redirect(status.SEE_OTHER, '/');
 });
+
+router.post('/delete-user/:uname', async (req, res) => {
+	// Special Account on MB
+	const USER_DELETER = 'UserDeleter';
+	const USER_DELETER_MBID = 2007538;
+	const {access_token: accessToken} = req.query;
+	const {uname} = req.params;
+	try {
+		const rBody = await request.get('https://musicbrainz.org/oauth2/userinfo').set('Authorization', `Bearer ${accessToken}`).then((rs) => rs.body);
+		if (rBody.metabrainz_user_id !== USER_DELETER_MBID || rBody.sub !== USER_DELETER) {
+			return res.status(401).send();
+		}
+		const {orm} = req.app.locals;
+		const {Editor} = orm;
+		const editor = await Editor.forge({name: uname}).fetch({require: false});
+		if (!editor) {
+			return res.status(404).send();
+		}
+		// deleting all user info
+		const deletedUser = `Deleted User#${editor.get('id')}`;
+		editor.set('bio', '');
+		editor.set('gender_id', null);
+		editor.set('area_id', null);
+		editor.set('revisions_applied', 0);
+		editor.set('revisions_reverted', 0);
+		editor.set('total_revisions', 0);
+		editor.set('metabrainz_user_id', null);
+		editor.set('cached_metabrainz_name', null);
+		editor.set('title_unlock_id', null);
+
+
+		editor.set('name', deletedUser);
+		await editor.save();
+		// eslint-disable-next-line camelcase
+		return res.send({metabrainz_user_id: USER_DELETER_MBID, sub: USER_DELETER});
+	}
+	catch (err) {
+		return res.status(err.status ?? 500).send();
+	}
+});
+
 
 export default router;


MonkeyDo (commented):
Looks great, thanks for doing this !

Considering the importance of this endpoint,  we definitely want some tests to ensure that the authorization is covered, in case of changes in the future we don't want this to break silently.

It means mocking the http request in the tests with [superagent](https://visionmedia.github.io/superagent/)

We also want a separate test to fetch a deleted user after the endpoint was called succesfully, and ensure all their info is wiped.

superagent-mock-config.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable camelcase */
```
MonkeyDo, line 1: This file is currently placed in the repo root, we'll want it in `test/test-helpers` instead.

test/src/server/routes/auth.js
```diff
@@ -0,0 +1,65 @@
...
+describe('delete user', () => {
+	const adminEditorId = 1;
+	const adminEditorName = 'UserDeleter';
+	const userEditorId = 2;
+	const userEditorName = 'NormalUser1';
+	const otherEditorId = 3;
+	const otherEditorName = 'NormalUser2';
+	let agent;
+	before(async () => {
+		await createEditor(adminEditorId, {name: adminEditorName});
```
MonkeyDo, line 45: This user shouldn't be necessary.
The UserDeleter special user is in the MusicBrainz database, not in the BookBrainz one, and we don't really interact with MB other than authenticate it

src/server/routes/auth.js
```diff
@@ -73,4 +83,45 @@ router.get('/logout', (req, res) => {
...
+			return res.status(401).send();
+		}
+		const {orm} = req.app.locals;
+		const {Editor} = orm;
+		const editor = await Editor.forge({name: uname}).fetch({require: false});
+		if (!editor) {
+			return res.status(404).send();
+		}
+		// deleting all user info
+		const deletedUser = `Deleted User#${editor.get('id')}`;
```
MonkeyDo, line 104:
```suggestion
		const deletedUser = `Deleted User #${editor.get('id')}`;
```

src/server/routes/auth.js
```diff
@@ -73,4 +83,45 @@ router.get('/logout', (req, res) => {
...
+	const {access_token: accessToken} = req.query;
+	const {uname} = req.params;
+	try {
+		const rBody = await request.get('https://musicbrainz.org/oauth2/userinfo').set('Authorization', `Bearer ${accessToken}`).then((rs) => rs.body);
+		if (rBody.metabrainz_user_id !== USER_DELETER_MBID || rBody.sub !== USER_DELETER) {
+			return res.status(401).send();
+		}
+		const {orm} = req.app.locals;
+		const {Editor} = orm;
+		const editor = await Editor.forge({name: uname}).fetch({require: false});
```
MonkeyDo, line 99: I had another look at the implementation of the same mechanism in ListenBrainz because I wasn't sure, but it confirmed that we want to delete users by their metabrainz_user_id instead of by name, because the name might be different between website while the meb_user_id will be the same for an account across the websites
```suggestion
		const editor = await Editor.forge({metabrainz_user_id: mbRowId}).fetch({require: false});
```

That'll also require changing the query param name to `mbRowId` as well as modifying the tests accordingly.

src/server/routes/auth.js
```diff
@@ -73,4 +83,45 @@ router.get('/logout', (req, res) => {
...
+		}
+		// deleting all user info
+		const deletedUser = `Deleted User#${editor.get('id')}`;
+		editor.set('bio', '');
+		editor.set('gender_id', null);
+		editor.set('area_id', null);
+		editor.set('revisions_applied', 0);
+		editor.set('revisions_reverted', 0);
+		editor.set('total_revisions', 0);
+		editor.set('metabrainz_user_id', null);
```
MonkeyDo, line 111: I need to think about this, not sure if we want to keep this to be able to link it back to an MB account.
I'm thinking for example for abusive accounts that would get deleted by admins across the platforms, it could be necessary to have this information.

src/server/routes/auth.js
```diff
@@ -73,4 +83,45 @@ router.get('/logout', (req, res) => {
...
+		editor.set('cached_metabrainz_name', null);
+		editor.set('title_unlock_id', null);
+
+
+		editor.set('name', deletedUser);
+		await editor.save();
+		// eslint-disable-next-line camelcase
+		return res.send({metabrainz_user_id: USER_DELETER_MBID, sub: USER_DELETER});
```
MonkeyDo, line 119: Probably no need to return any actual data. Looking at the description of https://tickets.metabrainz.org/browse/MBS-9680 and the comments there, MB only expects a 200 status code if all went well.

test/src/server/routes/auth.js
```diff
@@ -0,0 +1,65 @@
...
+		await createEditor(userEditorId, {name: userEditorName});
+		await createEditor(otherEditorId, {name: otherEditorName});
```
MonkeyDo, line 47: I'm not sure I understand why we need two more users here.
If I remember correctly, when you  use `agent = await chai.request.agent(app);` below it should run the superagent queries authenticated as our mock-user with id 123456.

Meaning your test "Normal User should not be able to delete other users" is actually testing that user 123456 is not authorized, which is completely valid but does not require the extraneous users

test/src/server/routes/auth.js
```diff
@@ -0,0 +1,65 @@
...
+		agent = chai.request.agent(app);
+		await agent.get('/cb');
+	});
+	after(truncateEntities);
+	it('Normal User should not be able to delete other users', async () => {
+		const res = await agent.post(`/delete-user/${otherEditorName}?access_token=${userEditorId}`);
+		expect(res.ok).to.be.false;
+		// unauthorized
+		expect(res.status).to.be.equal(401);
+	});
```
MonkeyDo, line 57: I'd also love to see, for good measure, another test to ensure we get a 401 for an unauthenticated user.
Basically the same test as the above, but using `await chai.request(app).post(…`  <- this will not use a user session because we won't call `/cb` as we do in the `before` function

test/test-helpers/create-entities.js
```diff
@@ -121,7 +121,7 @@ export function createEditor(editorId) {
 		editorAttribs.id = editorId || random.number();
 		editorAttribs.genderId = gender.id;
 		editorAttribs.typeId = editorType.id;
-		editorAttribs.name = internet.userName();
+		editorAttribs.name = otherEditorAttribs?.name || internet.userName();
```
MonkeyDo, line 124: Considering I'm already making you change something here (using metabrainzUserId instead of name) let's make it easier to be reusable:

After defining sane defaults, we can merge editorAttribs and otherEditorAttribs and pass the result to new Editor(…
Maybe the simple `new Editor({...editorAttribs,…otherEditorAttribs})` ?

MonkeyDo (changes requested):
OK, I did a thorough review, and good on us because I missed a bunch of things ! :p

Thanks a lot for writing tests, that allowed me to think about the code more thoroughly and project into specific scenarios.
I took special care with the review on this one, just to ensure nothing can go wrong :)

src/server/routes/auth.js
```diff
@@ -16,11 +16,21 @@
...
+import mock from 'superagent-mock';
 import passport from 'passport';
+import request from 'superagent';
 import status from 'http-status';
 
 
+// Setting up mocking agent for test
+// eslint-disable-next-line node/no-process-env
+if (process.env.NODE_ENV === 'test') {
+	mock(request, config);
```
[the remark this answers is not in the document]
tr1ten: Placing it inside the test file also seems to work, though i initially thought it needs to be in same lexical scope.

test/src/server/routes/auth.js
tr1ten: Yep, you are correct. we only need one user ( that we test for deletion) else only those ids would suffice.

PR #845, opened 2022-05-23
Fix(EntityEditor): Optimize Entity Editor's components

tr1ten:
### Problem

Slow behavior of entity editor form 

### Solution

minimize the re-render cycle for different components of entity editor


MonkeyDo (approved):
Awesome improvement, and super clean in the end !

We should keep that in mind if we ever encounter another state update bottleneck!

PR #847, opened 2022-05-26
Feat(uf-route): POST endpoint for submitting multiple entities

tr1ten:
Adding POST route "/create/handler" for unified form

### TO-DO

- [x] Added /create/handler endpoint
- [x] Added Tests for /create/handler


src/server/routes/entity/entity.tsx
```diff
@@ -1312,3 +1341,186 @@ export function displayPreview(req:PassportRequest, res:$Response, next) {
...
+					.then((unlock) => {
+						if (unlock.alert) {
+							entityJSON.alert = unlock.alert;
+						}
+						return entityJSON;
+					}));
+			}
+			return Promise.all(entitiesAchievementsPromise).catch(err => { throw err; });
+		}
+	);
```
MonkeyDo, line 1520: Now that we're duplicating this code, this could do with 1. Being refactored into a separate reusable function and 2. Being refactored to use async/await syntax

src/server/routes/entity/entity.tsx
```diff
@@ -1312,3 +1341,186 @@ export function displayPreview(req:PassportRequest, res:$Response, next) {
...
+			/* fetch and reindex EditionGroups that may have been created automatically by the ORM and not indexed */
+			if (savedMainEntity.get('type') === 'Edition') {
+				await indexAutoCreatedEditionGroup(orm, savedMainEntity, transacting);
+			}
+			bbidMap[entityKey] = savedMainEntity.get('bbid');
+			savedMainEntities[entityKey] = savedMainEntity.toJSON();
+		}
+		try {
+			// bookshelf's transaction have issue with Promise.All
+			await Object.keys(body).reduce((promise, entityKey) => promise.then(() => processEntity(entityKey)), Promise.resolve());
```
MonkeyDo, line 1481: This line is quite hard to read.
Can we look at fixing the underlying issue? Can you explain what the symptom was that you had when using Promise.all?
If we can't fix it, at least it would be good to explain in a comment what the code is doing

MonkeyDo (changes requested):
Looks very promising !
Let's start with these suggested changes; I'll do another full review after that, but generally I think everything works as expected, nicely done :)

In general, when there is a need to copy existing code from a method, I'd like to see the unchanged parts refactored into a function so that they can be reused without duplication.
That way, we avoid having to apply changes in multiple places which can lead to bugs and bitrot.

The tests are also a bit thin currently.
For the tests that check submissions work, checking a 200 response is a start but it would be ideal to then fetch the entities that were supposedly created and check they do exist and have the attributes we sent.
Same when testing adding relationships between the entities.
No need to do that for tests that expect an error code.

src/server/routes/entity/entity.tsx
tr1ten: Not sure if we can fix that, it seems to be common problem with bookshelf when using transaction with promise.all as described [here](https://github.com/bookshelf/bookshelf/issues/1498) 
I will add this issue link in the comment as well

src/client/entity-editor/validators/edition.ts
```diff
@@ -66,8 +66,8 @@ export function validateEditionSectionPages(value: any): boolean {
 	return validatePositiveInteger(value);
 }
 
-export function validateEditionSectionEditionGroup(value: any, editionGroupRequired: boolean | null | undefined): boolean {
-	return validateUUID(get(value, 'id', null), editionGroupRequired);
+export function validateEditionSectionEditionGroup(value: any, editionGroupRequired: boolean | null | undefined, isCustom = false): boolean {
+	return isCustom ? !editionGroupRequired || Boolean(get(value, 'id', null)) : validateUUID(get(value, 'id', null), editionGroupRequired);
```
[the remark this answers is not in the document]
tr1ten: Like other entities which are created  on the fly, they are first assign with temporary ids which later map to actual bbids on the db.
`validateUUID(get(value, 'id', null), editionGroupRequired)` may not work for temporary id.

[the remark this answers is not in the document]
MonkeyDo: Got it, thanks for the explanation, I forgot the temporary ids are not UUIDs

MonkeyDo:
I tried merging the updated new-creation-form branch into this PR in order to merge it, but there's a merge conflict with the publisher validator that I'm not sure how to solve; I think it's best if you had a look at it yourself:

There are some changes to allow multiple publishers from #848 , as well as some changes to allow custom publisher ids for the unified form from this PR: 
image

src/server/routes/entity/process-unified-form.ts
```diff
@@ -0,0 +1,241 @@
...
+		return Promise.reject(new Error(err));
+	}
+}
+
+export function transformForm(body:Record<string, any>):Record<string, any> {
+	const modifiedForm = {};
+	for (const keyIndex in body) {
+		if (Object.prototype.hasOwnProperty.call(body, keyIndex)) {
+			const currentForm = body[keyIndex];
+			const transformedForm = transformFunctions[_.lowerFirst(currentForm.type)](currentForm);
```
MonkeyDo, line 66: The tests are failing because of this line; we should probably play it safe in any case and either catch any possible error or first store the transformfunction in a variable, then check if the funcion exists and is actually a function (isFunction lodash utility) and only then call it.

MonkeyDo (approved):
OK, I think this is ready to merge, and from there we will move one to the UI part to simplify testing.
We can open another PR down the line if we need to improve of fix anything.

PR #848, opened 2022-05-27
Feat(edition): Allow multiple publishers for a single edition

tr1ten:
### Problem

BB-655: Allow Publisher field to have multiple publishers

### DIFF
--- a/src/client/helpers/entity.tsx
+++ b/src/client/helpers/entity.tsx
@@ -203,12 +203,12 @@ export function getEditionPublishers(edition) {
 
 	if (hasPublishers) {
 		return edition.publisherSet.publishers.map(
-			(publisher) =>
+			(publisher, index) =>
 				(
 					<span key={publisher.bbid}>
 						<a href={`/publisher/${publisher.bbid}`}>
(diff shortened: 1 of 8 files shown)


src/client/helpers/entity.tsx
```diff
@@ -203,11 +203,14 @@ export function getEditionPublishers(edition) {
...
-				<a href={`/publisher/${publisher.bbid}`} key={publisher.bbid}>
-					{_get(publisher, 'defaultAlias.name', publisher.bbid)}
-				</a>
-			)
+			(publisher) =>
+				(
+					<span key={publisher.bbid}>
+						<a href={`/publisher/${publisher.bbid}`}>
+							{_get(publisher, 'defaultAlias.name', publisher.bbid)}
+						</a> &nbsp;
```
MonkeyDo, line 211: It would be good to separate the Publishers with comas, currently it's hard to know they are different entities:
image

MonkeyDo (changes requested):
Looking good !

MonkeyDo (approved):

PR #849, opened 2022-05-31
Improve test helpers for creating Publisher

tr1ten:
Improve test helper for publisher creation, using single PublisherType for each publisher.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           # We could also test on multiple Node versions if needed: https://github.com/actions/setup-node#matrix-testing
-          node-version: '12'
+          node-version: '16'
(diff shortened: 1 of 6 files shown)


MonkeyDo (commented):
I've made a change to fix the issue with the CI tests.
Now we see three failing tests due to the use of hardcoded publisher type name in the API browse routes:
https://github.com/metabrainz/bookbrainz-site/blob/master/test/src/api/routes/test-publisher.js#L261
https://github.com/metabrainz/bookbrainz-site/blob/master/test/src/api/routes/test-publisher.js#L270
https://github.com/metabrainz/bookbrainz-site/blob/master/test/src/api/routes/test-publisher.js#L325

Note as well that the same tests also check that publisherType is equal to 'publisher type 1'.
I'm not certain, but I think it might be as easy as passing a `publisherAttrib.label= "publisher type 1"` in the [optional attributes passed to createPublisher](https://github.com/metabrainz/bookbrainz-site/blob/master/test/src/api/routes/test-publisher.js#L182)

PR #852, opened 2022-06-27
Feat(language-select): Asynchronously load options in language select

tr1ten:
Replacing virtualized select for language field with async select

### DIFF
--- a/src/client/entity-editor/common/language-field.tsx
+++ b/src/client/entity-editor/common/language-field.tsx
@@ -19,9 +19,9 @@
 
 import * as React from 'react';
 import {Form, OverlayTrigger, Tooltip} from 'react-bootstrap';
+import AsyncSelect from 'react-select/async';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import ValidationLabel from './validation-label';
-import VirtualizedSelect from 'react-virtualized-select';
 import {convertMapToObject} from '../../helpers/utils';
 import createFilterOptions from 'react-select-fast-filter-options';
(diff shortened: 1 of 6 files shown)


MonkeyDo (changes requested):
I tried to build to deploy on test, I get an error from the CSS compilation: https://github.com/metabrainz/bookbrainz-site/runs/7072398833?check_suite_focus=true#step:6:525
Because of this import that needs to be removed: https://github.com/metabrainz/bookbrainz-site/blob/83c554fad43567c71b5d2165f03d682c4b525ef3/src/client/stylesheets/style.scss#L31

src/client/entity-editor/common/language-field.tsx
```diff
@@ -56,14 +57,16 @@ function LanguageField({
 	const label =
 		<ValidationLabel empty={empty} error={error}>Language</ValidationLabel>
 	;
-
+	const MAX_F2_OPTIONS = 20;
```
MonkeyDo, line 60: Why limit this to 20?

MonkeyDo (approved):
Great !
With the virtualization issues we've had, I have two thoughts: nice job , and good riddance !

src/client/entity-editor/common/language-field.tsx
tr1ten: As discussed showing top 20ish (frequency >=2) languages in language field should be enough

MonkeyDo: I think there are currently only 21 languages with frequency = 2  in the DB, so I thought maybe it was related.
I'd say let's not limit the number of f >= 2 languages and only limit f < 2

PR #857, opened 2022-06-28
Add useful columns to entity views

MonkeyDo:
This is tied to the CritiqueBrainz-BookBrainz integration project.
Using the views is particularly helpful in the CB codebase since  it is in python and they cannot use our JS ORM.
So we are adding some columns to the views for each entity (alias name and sort name, disambiguation content, entity type labels, etc.)

### DIFF
--- a/sql/schemas/bookbrainz.sql
+++ b/sql/schemas/bookbrainz.sql
@@ -560,7 +560,7 @@ ALTER TABLE bookbrainz.relationship ADD FOREIGN KEY (target_bbid) REFERENCES boo
 -- Relationship Attributes
 
 CREATE TABLE bookbrainz.relationship_attribute_set (
-	id SERIAL PRIMARY KEY 
+	id SERIAL PRIMARY KEY
 );
 
 ALTER TABLE bookbrainz.relationship ADD COLUMN attribute_set_id INTEGER;
@@ -585,7 +585,7 @@ CREATE TABLE bookbrainz.relationship_type__attribute_type (
 );
 
 CREATE TABLE bookbrainz.relationship_attribute (
-  id SERIAL PRIMARY KEY, 
+  id SERIAL PRIMARY KEY,
   attribute_type INT NOT NULL REFERENCES bookbrainz.relationship_attribute_type(id)
 );
 
@@ -598,7 +598,7 @@ CREATE TABLE bookbrainz.relationship_attribute_set__relationship_attribute (
 	set_id INTEGER REFERENCES bookbrainz.relationship_attribute_set (id),
 	attribute_id INTEGER REFERENCES bookbrainz.relationship_attribute (id),
   PRIMARY KEY (
-    set_id, 
+    set_id,
     attribute_id
   )
 );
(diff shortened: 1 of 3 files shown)


MonkeyDo:
@anshg1214 let me know if I missed any of the columns you required.
I know there are no author credits; fetching and processing those is taking a fair amount of time, and I don't want to add it to the views if there is another method. Let me know if that will work for you.
You can run the `sql/migrations/2022-06-28-views/up.sql` script against your dev database if you want to check the new columns.

anshg1214:
@MonkeyDo Thanks! The views covers everything we need in CB. For the author credits we'll fetch them directly from the database.

sql/schemas/bookbrainz.sql
```diff
@@ -833,57 +836,71 @@ CREATE VIEW bookbrainz.edition AS
...
+		als.default_alias_id, al."name", al.sort_name, wd.type_id, worktype.label as work_type, wd.alias_set_id, wd.identifier_set_id,
 		wd.relationship_set_id, e.type, wd.language_set_id
 	FROM bookbrainz.work_revision wr
 	LEFT JOIN bookbrainz.entity e ON e.bbid = wr.bbid
 	LEFT JOIN bookbrainz.work_header w ON w.bbid = e.bbid
 	LEFT JOIN bookbrainz.work_data wd ON wr.data_id = wd.id
 	LEFT JOIN bookbrainz.alias_set als ON wd.alias_set_id = als.id
+	LEFT JOIN bookbrainz.alias al ON al.id = als.default_alias_id
+	LEFT JOIN bookbrainz.disambiguation dis ON dis.id = wd.disambiguation_id
+	LEFT JOIN bookbrainz.publisher_type worktype ON worktype.id = wd.type_id
```
alastair, line 855: should be work_type

sql/schemas/bookbrainz.sql
```diff
@@ -833,57 +836,71 @@ CREATE VIEW bookbrainz.edition AS
...
 	SELECT
-		e.bbid, pcd.id AS data_id, pcr.id AS revision_id, (pcr.id = pc.master_revision_id) AS master, pcd.annotation_id, pcd.disambiguation_id,
-		als.default_alias_id, pcd.type_id, pcd.author_credit_id, pcd.alias_set_id, pcd.identifier_set_id,
+		e.bbid, pcd.id AS data_id, pcr.id AS revision_id, (pcr.id = pc.master_revision_id) AS master, pcd.annotation_id, dis.comment,
+		als.default_alias_id, al."name", al.sort_name, pcd.type_id, egtype.label as edition_group_type, pcd.author_credit_id, pcd.alias_set_id, pcd.identifier_set_id,
 		pcd.relationship_set_id, e.type
 	FROM bookbrainz.edition_group_revision pcr
 	LEFT JOIN bookbrainz.entity e ON e.bbid = pcr.bbid
 	LEFT JOIN bookbrainz.edition_group_header pc ON pc.bbid = e.bbid
 	LEFT JOIN bookbrainz.edition_group_data pcd ON pcr.data_id = pcd.id
```
alastair, line 882: maybe change the alias of these so that it makes sense compared to the table name

sql/schemas/bookbrainz.sql
```diff
@@ -810,21 +810,24 @@ ALTER TABLE bookbrainz.user_collection_collaborator ADD FOREIGN KEY (collaborato
...
+		cd.gender_id, cd.type_id, atype.label as author_type, cd.alias_set_id, cd.identifier_set_id, cd.relationship_set_id, e.type
 	FROM bookbrainz.author_revision cr
 	LEFT JOIN bookbrainz.entity e ON e.bbid = cr.bbid
 	LEFT JOIN bookbrainz.author_header c ON c.bbid = e.bbid
 	LEFT JOIN bookbrainz.author_data cd ON cr.data_id = cd.id
 	LEFT JOIN bookbrainz.alias_set als ON cd.alias_set_id = als.id
+	LEFT JOIN bookbrainz.alias al ON al.id = als.default_alias_id
+	LEFT JOIN bookbrainz.disambiguation dis ON dis.id = cd.disambiguation_id
+	LEFT JOIN bookbrainz.author_type atype ON atype.id = cd.type_id
 	WHERE e.type = 'Author';
```
alastair, line 825: this review returns one row for each revision of an entity. I think that we want to treat the view as a snapshot of the current state of the database, so it probably makes sense to add `AND master=true` to all of the where statements here (and perhaps that means it makes sense to remove `master` from the returned column list?

MonkeyDo: I looked into this more, and there are a few places in the website and ORM where we use these views to fetch older revision.
It can probably be done without too much problems, but will require a lot of carefulness and time to test everything (lest we break the whole website somehow).
Leaving this for another time.

PR #859, opened 2022-07-07
chore(sql): Update identifier types' display templates

MonkeyDo:
These can be used to reconstitute a URL to the resource on the front-end.
We need them added to the database because they will now be used by CritiqueBrainz and we don't want to duplicate the code in both codebases.

The database table was already set up to have a template, which we previously decided not to use.

We will have to get rid of the code in src/client/components/pages/entities/identifiers-links.js while making sure to preserve our [one special use-case](https://github.com/metabrainz/bookbrainz-site/blob/60ff054647e824f5cd2146ca132db22099ab92b9/src/client/components/pages/entities/identifiers-links.js#L65)

### DIFF
--- a/sql/migrations/2022-07-07-identifier-templates/down.sql
+++ b/sql/migrations/2022-07-07-identifier-templates/down.sql
@@ -1,3 +1,5 @@
+BEGIN TRANSACTION;
 UPDATE bookbrainz.identifier_type
 SET display_template = 'Placeholder Template'
-WHERE id IN (1,2,3,4,5,6,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34)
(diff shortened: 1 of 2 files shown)


sql/migrations/2022-07-07-identifier-templates/down.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE bookbrainz.identifier_type
```
alastair, line 1: do you have a guideline to always run migrations in a transaction, or only when there are multiple statements?

MonkeyDo: We don't really have an SQL guideline other than the in-progress MB one, but I think this is a sane suggestion regardless of the fact that there's only one statement.
Who knows who is going to copy this file's structure in the future?

PR #860, opened 2022-07-08
fix(language-select): Correctly format language options for alias row

tr1ten:
Fix language display issue in alias-row


MonkeyDo (approved):
Thank you for the quick fix !

PR #862, opened 2022-07-12
feat(editionGroup): Show CB reviews for Edition Group

anshg1214:
This PR adds CritiqueBrainz reviews for EditionGroup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "react-redux": "^7.2.6",
     "react-select": "^4.3.1",
     "react-select-fast-filter-options": "^0.2.3",
+    "react-simple-star-rating": "^4.0.5",
     "react-sortable-hoc": "^2.0.0",
     "react-sticky": "^6.0.1",
     "redis": "^3.1.2",
--- a/src/client/components/pages/entities/cb-review.js
+++ b/src/client/components/pages/entities/cb-review.js
@@ -1,16 +1,21 @@
 import * as bootstrap from 'react-bootstrap';
 import React from 'react';
 import {Rating} from 'react-simple-star-rating';
+import {getEntityLink} from '../../../../common/helpers/utils';
+import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
+import {faPlus} from '@fortawesome/free-solid-svg-icons';
 
 
-const {Col, Row} = bootstrap;
+const {Button, Col, Row} = bootstrap;
 
+
+const REVIEW_CONTENT_PREVIEW_LENGTH = 75;
 
 function ReviewCard({reviewData}) {
 	const publishedDate = new Date(reviewData.published_on).toDateString();
 	let reviewText = reviewData.text;
-	if (reviewText.length > 75) {
-		reviewText = `${reviewText.substring(0, 75)}...`;
+	if (reviewText.length > REVIEW_CONTENT_PREVIEW_LENGTH) {
+		reviewText = `${reviewText.substring(0, REVIEW_CONTENT_PREVIEW_LENGTH)}...`;
 	}
 	const reviewLink = `https://critiquebrainz.org/review/${reviewData.id}`;
 	return (
@@ -39,14 +44,15 @@ function ReviewCard({reviewData}) {
 	);
 }
 
-function EntityReviews({entity}) {
-	const {reviews} = entity.reviews;
+function EntityReviews({entityReviews, entityType, entityBBID}) {
 	let reviewContent;
 	const mapEntityType = {
-		EditionGroup: 'bb_edition_group'
+		EditionGroup: 'edition-group'
 	};
-	const entityType = mapEntityType[entity.type];
-	if (reviews.length) {
+	entityType = mapEntityType[entityType];
+	const entityLink = `https://critiquebrainz.org/${entityType}/${entityBBID}`;
+	if (entityReviews?.length) {
+		const {reviews} = entityReviews;
 		reviewContent = (
 			<React.Fragment>
 				{
@@ -57,8 +63,23 @@ function EntityReviews({entity}) {
 						/>
 					))
 				}
-				<a href={`https://critiquebrainz.org/${entityType}/${entity.bbid}`}>View all reviews &gt;</a>
+				<a href={entityLink}>View all reviews &gt;</a>
 			</React.Fragment>
+		);
+	}
+	else {
+		reviewContent = (
+			<div>
+				<h4>No reviews yet.</h4>
+				<Button
+					className="margin-top-d15"
+					href={entityLink}
+					variant="success"
+				>
+					<FontAwesomeIcon icon={faPlus}/>
+					{'  Add a review'}
+				</Button>
+			</div>
 		);
 	}
 	return (
--- a/src/client/components/pages/entities/cb-review.js
+++ b/src/client/components/pages/entities/cb-review.js
@@ -81,11 +81,7 @@ class EntityReviews extends React.Component {
 	}
 
 	async handleClick() {
-		const data = await request.get('/external-service/critiquebrainz/reviews')
-			.query({
-				entityBBID: this.entityBBID,
-				entityType: this.entityType
-			});
+		const data = await request.get(`/${this.entityType}/${this.entityBBID}/reviews`);
 
 		this.setState({
 			reviews: data.body.reviews,
--- a/src/client/components/pages/entities/cb-review.js
+++ b/src/client/components/pages/entities/cb-review.js
@@ -1,0 +1,70 @@
+import * as bootstrap from 'react-bootstrap';
+const { Col, Row } = bootstrap;
+import React from 'react';
+import { Rating } from "react-simple-star-rating";
+
+
+function ReviewCard({ reviewData }) {
+    const publishedDate = new Date(reviewData.published_on).toDateString();
+    let reviewText = reviewData.text;
+    if (reviewText.length > 75) {
+        reviewText = reviewText.substring(0, 75) + '...';
+    }
+    const reviewLink = "https://critiquebrainz.org/review/" + reviewData.id;
+    return (
+        <div className='border bg-light p-2 mb-2' >
+            <div>
+                <Rating
+                    className="rating-stars"
+                    ratingValue={0}
+                    initialValue={reviewData.rating}
+                    size={15}
+                    stars={5}
+                    readonly
+                    allowHalfIcon
+                    fillColor="#46433A"
+                    allowHover={false}
+                />
+                <small className='float-right'>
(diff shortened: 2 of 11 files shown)


src/client/components/pages/entities/edition-group.js
```diff
@@ -105,12 +125,20 @@ function EditionGroupDisplayPage({entity, identifierTypes, user}) {
...
+						<EntityLinks
+							entity={entity}
+							identifierTypes={identifierTypes}
+							urlPrefix={urlPrefix}
+						/>
+						<EntityRelatedCollections collections={entity.collections}/>
+
+					</Col>
+					<Col lg={4}>
+						<EntityReviews entity={entity} />
```
MonkeyDo, line 139: We can directly pass the props we need rather than the whole entity object to avoid unnecessary rerenders.
Not sure it will have a big impact here but it's best practice:
```suggestion
						<EntityReviews entity={entity.reviews} entityType={entity.type} />
```

Or even, after refactoring the code as I suggest in another comment:
```suggestion
						<EntityReviews entity={entity.reviews} cbEntityType={getReviewEntityNameFromType (entity.type)} />
```

src/server/helpers/critiquebrainz.ts
```diff
@@ -0,0 +1,18 @@
...
+    console.log(entityType);
+	const res = await request
+        .get("https://beta.critiquebrainz.org/ws/1/review")
+        .query({
+            entity_id: bbid,
+            entity_type: entityType,
+            limit: 10,
+            offset: 0,
+        });
+    return res.body;
```
MonkeyDo, line 17: This will need some error handling.
You'll need to do a few things: 
1. wrap this block (lines 9-17) in a try-catch block
2. I think (the docs aren't very clear for async/await syntax) that you might need to only return the body if `res.ok` is truthy 
3. if `res.ok` is falsy you can throw an error; not sure how you get the error message though
4. In the catch block you can log any error and return…judging from [this line](https://github.com/metabrainz/bookbrainz-site/pull/862/files#diff-1e08379bd6206ad9c3911bf9d48d1977e284ffc53ae218928448a2f197bdd8e9R47) an empty array?


It is possible that superagent throws errors for successful requests with an error http status, in which case steps 2 and 3 aren't needed.

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,70 @@
...
+    )
+}
+
+function EntityReviews({ entity }) {
+    const reviews = entity.reviews.reviews;
+    let reviewContent;
+    const mapEntityType = {
+		EditionGroup: "bb_edition_group",
+	};
+    let entityType = mapEntityType[entity.type];
```
MonkeyDo, line 46: ~~This code is duplicated; it would be good to make it a helper (we won't win a prize for the name, but what about `getReviewEntityNameFromType` or `getCritiqueBrainzEntityType`?) in `scr/common/…` that can be used both by the server and client side.~~

Edit: With my comment below about kebabCase I don't think this will be necessary

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,70 @@
...
+                        <ReviewCard
+                            reviewData={review}
+                            key={review.id}
+                        />
+                    ))
+                }
+                <a href={`https://critiquebrainz.org/${entityType}/${entity.bbid}`}>View all reviews &gt;</a>
+            </React.Fragment>
+        )
+    }
```
MonkeyDo, line 61: I think you could add an else block and setting reviewContent accordingly for empty reviews.
I'm thinking a short message and a call to action, along the lines of "No reviews yet" with a  "Write a review" button that will later open a modal.
Fro now it can be a link to CB.

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,70 @@
...
+            <React.Fragment>
+                {
+                    reviews.slice(0, 3).map((review) => (
+                        <ReviewCard
+                            reviewData={review}
+                            key={review.id}
+                        />
+                    ))
+                }
+                <a href={`https://critiquebrainz.org/${entityType}/${entity.bbid}`}>View all reviews &gt;</a>
```
MonkeyDo, line 58: This isn't right at the moment with entityType `bb_edition_group` defined on line 44. Should be `edition-group`.
In which case if we just want to transform the entity names with a dash, I would recommend using lodash's `kebabCase` method like so: https://github.com/metabrainz/bookbrainz-site/blob/83c554fad43567c71b5d2165f03d682c4b525ef3/src/common/helpers/utils.ts#L169

package.json
```diff
@@ -72,6 +72,7 @@
     "react-redux": "^7.2.6",
     "react-select": "^4.3.1",
     "react-select-fast-filter-options": "^0.2.3",
+    "react-simple-star-rating": "^4.0.5",
```
MonkeyDo, line 75: You should also commit the change in `yarn.lock`  file.
If you did not install with yarn (`yarn add react-simple-star-rating`)  you might have to run the command again to generate the changes to the lockfile.

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,70 @@
...
+const { Col, Row } = bootstrap;
+import React from 'react';
+import { Rating } from "react-simple-star-rating";
+
+
+function ReviewCard({ reviewData }) {
+    const publishedDate = new Date(reviewData.published_on).toDateString();
+    let reviewText = reviewData.text;
+    if (reviewText.length > 75) {
+        reviewText = reviewText.substring(0, 75) + '...';
```
MonkeyDo, line 11: Let's define this number as a constant REVIEW_CONTENT_PREVIEW_LENGTH = 75 and use it below 

```suggestion
    if (reviewText.length > REVIEW_CONTENT_PREVIEW_LENGTH) {
        reviewText = reviewText.substring(0, REVIEW_CONTENT_PREVIEW_LENGTH) + '...';
```

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,70 @@
...
+            <div>
+                {reviewText}
+                <a className='float-right' href={reviewLink}>View &gt;</a>
+            </div>
+        </div>
+    )
+}
+
+function EntityReviews({ entity }) {
+    const reviews = entity.reviews.reviews;
```
MonkeyDo, line 41: I think this can be improved:
1. if entity.reviews is undefined, or not an object, this line will throw an error
2. I commented in another comment but we can directly pass the reviews array and entity type as props rather than the whole entity object.

MonkeyDo (changes requested):
Great start @anshg1214 !

It's really exciting to see it come together !

There are a few formatting  (spaces instead of tabs) issues that will need to be fixed. 
You can run this script to show you formatting errors:
`yarn lint` or `npm run lint`
You can also auto-fix the simple issues with `yarn lint --fix` /  `npm run lint --fix`

src/client/components/pages/entities/cb-review.js
anshg1214: We would have to keep this, because the links in CB are not always the kebabCase links returned by the defined function in utils. Like for the entity `Work` , we don't have the link critiquebrainz.org/work/\<entity-id> for it.

src/server/helpers/critiquebrainz.ts
```diff
@@ -0,0 +1,49 @@
...
+				// eslint-disable-next-line camelcase
+				entity_type: cbEntityType,
+				limit: 10,
+				offset: 0
+			});
+		return {reviews: res.body, successfullyFetched: true};
+	}
+	catch (err) {
+		// eslint-disable-next-line no-console
+		console.error(err);
```
MonkeyDo, line 46: This should use the `log` library instead like done [here](https://github.com/metabrainz/bookbrainz-site/blob/0e6807c9b6151dc166e201a3aa2d6425db2902b9/src/server/helpers/handler.js#L36) for example
```suggestion
		log.error(err);
```

src/server/routes/externalService.js
```diff
@@ -0,0 +1,34 @@
...
+
+import * as cbHelper from '../helpers/critiquebrainz';
+
+
+import express from 'express';
+
+
+const router = express.Router();
+
+router.get('/critiquebrainz/reviews', async (req, res) => {
```
MonkeyDo, line 28: I'm not convinced by this URL structure.
I think compared to other routes, it would make more sense to have `/$entityType/$entityBBID/reviews`, which does mean adding that sub-route to each entity router.
And in that case we probably don't need to send the entity type in the query params.

What do you think? What was your thought process for this endpoint?

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,178 @@
...
+
+const {Button} = bootstrap;
+
+
+const REVIEW_CONTENT_PREVIEW_LENGTH = 75;
+
+function ReviewCard({reviewData}) {
+	const publishedDate = new Date(reviewData.published_on).toDateString();
+	let reviewText = reviewData.text;
+	if (reviewText.length > REVIEW_CONTENT_PREVIEW_LENGTH) {
```
MonkeyDo, line 37: This throws an error if there is no review or no text:
`Unhandled rejection TypeError: Cannot read properties of null (reading 'length')`

This should do the trick:
```suggestion
	if (reviewText?.length > REVIEW_CONTENT_PREVIEW_LENGTH) {
```

src/client/components/pages/entities/cb-review.js
```diff
@@ -0,0 +1,178 @@
...
+import request from 'superagent';
+
+
+const {Button} = bootstrap;
+
+
+const REVIEW_CONTENT_PREVIEW_LENGTH = 75;
+
+function ReviewCard({reviewData}) {
+	const publishedDate = new Date(reviewData.published_on).toDateString();
```
MonkeyDo, line 35: I would be conservative here and return undefined if reviewData is falsy.
You never know…

MonkeyDo (commented):
Nice! this is going well!

src/server/helpers/critiquebrainz.ts
anshg1214: Done

src/server/routes/externalService.js
anshg1214: Yes, I also think this URL structure is much cleaner and makes more sense. But we don't need to add a sub-route to each entity router. One main router can handle reviews of every entity type.

MonkeyDo (approved):
Awesome, thanks ! :)
Nothing more to add. Looking forward to seeing the rest of the feature and the average rating once it's been added to the CB API.

As we previously discussed, i'm merging this into the `critiquebrainz-integration` feature branch.

PR #865, opened 2022-07-19
Add CritiqueBrainz as Oauth Service

anshg1214:
This PR:
1. Connects CritiqueBrainz as an external service to allow users rate and write reviews.
2. Adds Modal for posting reviews.
3. Allows Users to post reviews for Edition Group, Author, Work and Series.

Pending tasks: 
Update config files.


src/server/helpers/critiquebrainz.ts
```diff
@@ -16,10 +16,124 @@
...
+			.type('form')
+			.send({
+				client_id: cbConfig.clientID,
+				client_secret: cbConfig.clientSecret,
+				code,
+				grant_type: 'authorization_code',
+				redirect_uri: cbConfig.redirectURL
+			});
+
+		const token = await data.body;
```
MonkeyDo, line 83: Reading the superagent docs, I think data.body is a JSON object rather than a promise, so you wouldn't need tha `await` here:
https://visionmedia.github.io/superagent/#parsing-response-bodies

Same goes for refreshAccessToken

src/server/routes/externalService.js
```diff
@@ -0,0 +1,146 @@
...
+
+
+router.get('/critiquebrainz/callback', auth.isAuthenticated, async (req, res) => {
+	const {orm} = req.app.locals;
+	const {code} = req.query;
+	const editorId = req.user.id;
+	if (!code) {
+		res.send('No code provided');
+	}
+	const token = await cbHelper.fetchAccessToken(code, editorId, orm);
```
MonkeyDo, line 80: Not sure if this needs to be wrapped in a try-catch block;
A unit test where you manually fail the request on purpose (using a mock) will help us know what happens in that case.

src/server/routes/externalService.js
```diff
@@ -0,0 +1,146 @@
...
+	const {orm} = req.app.locals;
+	let token = await orm.func.externalServiceOauth.getOauthToken(
+		editorId,
+		'critiquebrainz',
+		orm
+	);
+	if (!token?.length) {
+		res.send('User has not connected to CB');
+	}
+	token = token[0];
```
MonkeyDo, line 100: This ties back to my comment on the ORM PR: are we expecting that the getOauthToken method returns an array while the others return a single entry?

src/server/routes/externalService.js
```diff
@@ -0,0 +1,146 @@
...
+		editorId,
+		'critiquebrainz',
+		orm
+	);
+	if (token?.length) {
+		await orm.func.externalServiceOauth.deleteOauthToken(
+			editorId,
+			'critiquebrainz',
+			orm
+		);
```
MonkeyDo, line 127: A comment would be helpful here to explain why we delete the token if one already exists. 
I don't understand it myself currently :)

src/server/routes/externalService.js
```diff
@@ -0,0 +1,146 @@
...
+	const redirectUrl = cbHelper.getAuthURL();
+
+	return res.send(redirectUrl);
+});
+
+
+router.post('/critiquebrainz/disconnect', auth.isAuthenticated, async (req, res) => {
+	const editorId = req.user.id;
+	const {orm} = req.app.locals;
+	await orm.func.externalServiceOauth.deleteOauthToken(
```
MonkeyDo, line 138: Similarly here a test to see what happens if the request fails would tell us if we need to wrap this in a try-catch block

MonkeyDo (commented):
This is looking quite nice at first glance, nicely done :)

I haven't run it locally yet but have a few preliminary comments below:

src/server/routes/externalService.js
anshg1214: I've corrected the ORM PR. We expect it to be a single entity

src/server/routes/externalService.js
anshg1214: Since fetchAccessToken is already wrapped in a try-catch block, we don't need to add it here.

src/server/helpers/critiquebrainz.ts
```diff
@@ -16,10 +16,124 @@
...
+
+export async function addNewUser(
+	editorId: number,
+	token: Record<string, any>,
+	orm: Record<string, any>
+): Promise<any> {
+	const expires = Math.floor(new Date().getTime() / 1000.0) + token.tokenExpires;
+
+	try {
+		const newUser = await orm.func.externalServiceOauth.saveOauthToken(
```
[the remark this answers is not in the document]
anshg1214: Here, `saveOauthToken` saves the token, and also updates it in case of conflict. So I chose to do it by raw SQL queries.

src/client/components/pages/entities/cb-review.js
```diff
@@ -78,6 +78,7 @@ class EntityReviews extends React.Component {
 		};
```
MonkeyDo, line 78: On line 63, I'm not sure we need the `pb-3` class; it's making the padding at the top and bottom uneven

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+	readonly maxTextLength = 100000;
+
+	private CBBaseUrl = 'https://critiquebrainz.org';
+
+	private MBBaseUrl = 'https://metabrainz.org';
+
+	// gets all iso-639-1 languages and codes for dropdown
+	private allLanguagesKeyValue = Object.entries(iso.getNames('en'));
+
+	private CBInfoButton = (
```
MonkeyDo, line 92: On small screens the tooltip is very long and cropped to make it unreadable:
image

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+			try {
+				let result: any;
+				const response = await request.post(`/${entityType}/${entityBBID}/reviews`)
+					.set('Content-Type', 'application/json')
+					.send({
+						accessToken: this.accessToken,
+						review
+					});
+
+				if (response.status >= 200 && response.status < 300) {
```
MonkeyDo, line 266: Fetch already does that neatly for us :)
```suggestion
				if (response.ok) {
```

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+					this.setState({
+						reviewID: result?.reviewID
+					});
+					this.resetCBReviewForm();
+				}
+				else {
+					this.setState({
+						alert: {
+							message: result?.message,
+							title: 'Error',
```
MonkeyDo, line 287:
```suggestion
							title: 'Error submitting your review',
```

src/server/routes/externalService.js
```diff
@@ -0,0 +1,181 @@
...
+		return next(new BadRequestError('Response type must be code'));
+	}
+	const token = await cbHelper.fetchAccessToken(code, editorId, orm);
+
+	let alertType = 'success';
+	let alertDetails = 'You have successfully linked your account with CritiqueBrainz.';
+	if (!token) {
+		alertType = 'danger';
+		alertDetails = 'Failed to connect to CritiqueBrainz';
+	}
```
MonkeyDo, line 97: We already have an alert system in place, you should be able to add an alert to `res.locals.alert` like it is done here for example and remove the code to display these alerts from the ExternalServices component:
https://github.com/metabrainz/bookbrainz-site/blob/e001f4b05729528fb7b414f458bccd9457b457b6/src/server/app.js#L148

src/server/routes/reviews.js
```diff
@@ -29,4 +30,44 @@ router.get('/:entityType/:bbid/reviews', async (req, res) => {
...
+	const {accessToken} = req.body;
+	const {review} = req.body;
+
+	let response = await cbHelper.submitReviewToCB(
+		accessToken,
+		review
+	);
+
+	// If the token has expired, we try to refresh it and then submit the review again.
+	if (response.message === 'The provided authorization token is invalid, expired, revoked, or was issued to another client.') {
```
MonkeyDo, line 46: Is there not a specific error code we could use here instead? This seems a bit flimsy (would break if the error message ever changes)

MonkeyDo (changes requested):
It's awesome to see it come together so well ! :) 

As far as I tested it's working quite well, with a few details and improvements below.
 A couple of more general thoughts:

- There should probably be a link to the external services page in the menu, under the profile options, as well as a tab in the editor page. This can be done in another PR though.
- How would you feel about using the logo's orange as the colour for the rating stars? 
image

src/server/routes/reviews.js
anshg1214: True. I'll replace it with a specific error code.

src/client/components/pages/entities/author.js
```diff
@@ -64,6 +67,20 @@ function AuthorAttributes({author}) {
 					<dl>
 						<dt>Sort Name</dt>
 						<dd>{sortNameOfDefaultAlias}</dd>
+						<dt>Ratings</dt>
+						<dd>
+							<Rating
```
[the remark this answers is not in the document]
anshg1214: Yes this information is available in CB. So it can be returned in the API along with the average ratings.

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+						<h4
+							className="modal-title"
+							id="CBReviewModalLabel"
+							style={{padding: 'auto'}}
+						>
+							<img
+								alt="CritiqueBrainz Logo"
+								className="cb-img-responsive"
+								height="30"
+								src="/images/critiquebrainz-logo.svg"
```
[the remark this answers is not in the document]
anshg1214: The image got ignored by git. I've now added it  https://github.com/metabrainz/bookbrainz-site/blob/f89a85c01337877fe811a5baf4f5c8a838bc6bf6/.gitignore#L17

config/config.json.example
```diff
@@ -9,6 +9,11 @@
 		"clientSecret": null,
 		"callbackURL": "http://localhost:9099/cb"
 	},
+    "critiquebrainz": {
```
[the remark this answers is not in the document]
anshg1214: Sure! I'll add it there also

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+		) {
+			let nonZeroRating: number;
+			if (rating !== 0) {
+				nonZeroRating = rating;
+			}
+
+			const review = {
+				entityBBID,
+				entityType,
+				language,
```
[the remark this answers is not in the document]
anshg1214: Fixed :)

src/client/components/pages/entities/cb-review.js
```diff
@@ -118,8 +119,8 @@ class EntityReviews extends React.Component {
 					<h4>No reviews yet.</h4>
 					<Button
 						className="margin-top-d15"
-						href={entityLink}
 						variant="success"
+						onClick={this.handleModalToggle}
```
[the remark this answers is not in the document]
anshg1214: Fixed

src/client/stylesheets/style.scss
```diff
@@ -726,4 +726,66 @@ ul {
 }
 .series-editor-select {
 	margin-top: 1.7em;
+}
+
+.external-service-option {
```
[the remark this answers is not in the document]
anshg1214: These styles look good. I've added them

src/client/components/pages/entities/cbReviewModal.tsx
anshg1214: Fixed

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+					.send({
+						accessToken: this.accessToken,
+						review
+					});
+
+				if (response.status >= 200 && response.status < 300) {
+					result = response.body;
+				}
+
+				if (result?.reviewID) {
```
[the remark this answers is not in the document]
anshg1214: Yes, it sounds good. I've stored the reference of EntityReviews component in the entities page as `reviewsRef`. When the user submits a review successfully, a function `handleUpdateReviews` which is passed as props to CBReviewModal is executed which updates the reviews.

src/client/components/pages/externalService.js
```diff
@@ -0,0 +1,155 @@
...
+
+const {Alert} = bootstrap;
+class ExternalServices extends React.Component {
+	constructor(props) {
+		super(props);
+		this.alertDetails = props.alertDetails;
+		this.alertType = props.alertType;
+		this.state = {
+			cbPermission: props.cbPermission
+		};
```
MonkeyDo, line 33: The alert info should use the component's state rather than being stored as a variable. Then the click handler should update the state accordingly.
```suggestion
		this.state = {
			cbPermission: props.cbPermission,
			alertDetails: props.alertDetails,
			alertType: props.alertType
		};
```

anshg1214: Done

src/client/components/pages/entities/author.js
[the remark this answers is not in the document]
MonkeyDo: I don't want to hold up this PR, so let's do that in a separate PR.

src/client/components/pages/entities/cbReviewModal.tsx
```diff
@@ -0,0 +1,549 @@
...
+				(target as HTMLInputElement).checked :
+				target.value;
+		const {name} = target;
+
+		this.setState({
+			[name]: value
+		} as any);
+	};
+
+	handleTextInputChange = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
```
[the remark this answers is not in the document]
MonkeyDo: After discussing this a bit more, that would probably be over-optimizing something that doesn't need to be.

src/server/routes/externalService.js
MonkeyDo: After further discussion, the existing alert system is not entirely appropriate, as it only supports alerts sent as props on page load.
In this feature we also need to use alerts to show dynamic changes, which will require refactoring the alert system as a higher-order-component and/or using a React Context so that we can create alerts from any place in the front-end codebase.

src/server/templates/target.js
```diff
@@ -55,6 +55,7 @@ export default ({
 		'BookBrainz – The Open Book Database'}</title>
 			<link rel='stylesheet' href='/stylesheets/bundle.css' />
 			<link rel='stylesheet' href='/stylesheets/style.css' />
+			<link rel='stylesheet' href='/stylesheets/external-services.css' />
```
MonkeyDo, line 58: So I wasn't very clear in my instructions about how to add a new SCSS file, my bad !

Instead of processing it separately and adding it in target.js, the file needs to be imported in styles.scss.
It will be compiled along with the rest in the final style.css file.
See for example how we import other files here: https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/stylesheets/style.scss#L30

You'll need to add `@import "./external-services.scss";` and  then remove mentions of the new css file from this file and from webpack.clisent.js

MonkeyDo (changes requested):
Looking very good! Thanks for the changes.

Final details, and any other improvements can be done in a separate PR.

src/server/templates/target.js
anshg1214: This method makes more sense. I've updated it.

src/server/routes/reviews.js
```diff
@@ -29,4 +30,44 @@ router.get('/:entityType/:bbid/reviews', async (req, res) => {
...
+				token.refresh_token,
+				orm
+			);
+
+			response = await cbHelper.submitReviewToCB(
+				newAccessToken.access_token,
+				review
+			);
+		}
+		catch (error) {
```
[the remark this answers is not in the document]
anshg1214: Yes, you are right. The second call to `submitReviewToCB` does not require to be wrapped inside try-catch block. I have updated it.

src/client/components/pages/entities/work.js
```diff
@@ -90,7 +133,10 @@ function WorkDisplayPage({entity, identifierTypes, user}) {
 				</Col>
 				<Col lg={10}>
 					<EntityTitle entity={entity}/>
```
MonkeyDo, line 135: This EntityTitle is missing the prop `handleModalToggle={handleModalToggle}`

MonkeyDo (changes requested):
Last small changes request but everything I tested is working very well !

We talked about showing the total number of reviews (especially next to the average rating).
Did you want to do this in this PR or a separate one?
Here's what I propose for that:
image
FYI I'm applying the classes `small` and `text-muted` for styling the text.

src/client/components/pages/entities/work.js
anshg1214: Added

src/client/stylesheets/external-services.scss
```diff
@@ -0,0 +1,66 @@
...
+
+			width: 40px;
+			height: 40px;
+			line-height: 40px;
+
+			border: 2px solid #5aa854;
+			background: white;
+			box-shadow: 3px 1px 5px -2px #053b47;
+		}
+	}
```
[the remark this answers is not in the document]
anshg1214: I've added the styles.

src/client/components/pages/entities/cb-review.js
```diff
@@ -118,8 +117,8 @@ class EntityReviews extends React.Component {
...
```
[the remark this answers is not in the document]
anshg1214: I'll fix this. They are maps for two different things. The map in the server maps BB Entity Name to CB Entity Name whereas the map in Client maps BB Entity Name to CB Urls

src/client/components/pages/entities/author.js
```diff
@@ -40,7 +40,7 @@ const {deletedEntityMessage, extractAttribute, getTypeAttribute, getEntityUrl,
 	ENTITY_TYPE_ICONS, getSortNameOfDefaultAlias, transformISODateForDisplay} = entityHelper;
 const {Button, Col, Row} = bootstrap;
 
-function AuthorAttributes({author, averageRating}) {
+function AuthorAttributes({author, averageRating, reviewsCount}) {
```
MonkeyDo, line 43: A question about this: If I followed correctly, the `reviews` object should already be in the `author` object; why are we passing the ratings props separately?

src/client/components/pages/entities/author.js
```diff
@@ -81,6 +81,11 @@ function AuthorAttributes({author, averageRating}) {
 								stars={5}
 							/>
 						</dd>
+						{reviewsCount ?
+							<dd className="small text-muted">
+								{reviewsCount} Review
```
MonkeyDo, line 86: We can add a conditional 's' when there are multiple reviews, and using the same capitalization as in the cb-review component:
```suggestion
								{reviewsCount} review{reviewsCount > 1 ? 's' : ''}
```

Since the same code is repeated for each entity, I'm thinking we should have a separate review component where we can isolate this logic, and reuse that component for all entities.

src/client/components/pages/entities/author.js
```diff
@@ -81,6 +81,11 @@ function AuthorAttributes({author, averageRating}) {
 								stars={5}
 							/>
 						</dd>
+						{reviewsCount ?
```
MonkeyDo, line 84: I think it would be good to show "No reviews" or "0 reviews" when reviewsCount=0.
Currently we are not rendering anything i reviewsCount=0.

src/client/components/pages/entities/cb-review.js
```diff
@@ -142,7 +146,14 @@ class EntityReviews extends React.Component {
 		}
 		return (
 			<Row className="flex-column">
-				<h2>Reviews</h2>
+				<h2>
+                    Reviews
+					{this.reviewsCount ?
+						<span className="small text-muted">
+						    {` ${this.reviewsCount} review`}
```
MonkeyDo, line 153: Same comment as above with the conditional 's' after 'review'

src/client/components/pages/entities/author.js
```diff
@@ -64,6 +67,25 @@ function AuthorAttributes({author}) {
...
+								className="rating-stars"
+								fillColor="#46433A"
+								initialValue={averageRating}
+								ratingValue={0}
+								size={20}
+								stars={5}
+							/>
+						</dd>
+						{reviewsCount ?
+							<dd className="small text-muted">
```
MonkeyDo, line 85: You can also add the utility css class `mt-n2` (= "margin top negative 2") as well to have the text a bit closer to the stars:
```suggestion
							<dd className="mt-n2 small text-muted">
```

Before:
image

After: 
image

src/client/components/pages/entities/author.js
anshg1214: Done

src/client/components/pages/entities/author.js
anshg1214: That's a good idea. I've made the changes

src/client/components/pages/entities/author.js
anshg1214: This looks much neat. Done 👍

src/client/components/pages/entities/author.js
anshg1214: Correct. The ratings data is inside the `author` object. I've made the changes required.

MonkeyDo (approved):
Thank you so much, great work !

I'm excited to see the community's reaction ! :)

PR #867, opened 2022-07-26
Fix(AC): Handle AC modal state for editionGroup entity

tr1ten:
### Problem

Edit button on AC section doesn't work for edition-group entity form.

### Solution

Add AC modal state in editionGroupSection.


MonkeyDo (approved):
Thanks for fixing this !

PR #869, opened 2022-08-07
BB-682: Use plain links in call-to-action buttons

tr1ten:
Relevant ticket [BB-682](https://tickets.metabrainz.org/projects/BB/issues/BB-682): Make Create entity buttons real links


MonkeyDo (approved):
Thank you :)

PR #871, opened 2022-08-11
Fix(AC): AC field should show error when empty

tr1ten:
### Problem

Since Author Credit field is necessary in entity form, it should show error when author-credit-editor is empty.

### DIFF
--- a/src/client/entity-editor/validators/common.ts
+++ b/src/client/entity-editor/validators/common.ts
@@ -193,7 +193,7 @@ export function validateAuthorCreditRow(row: any): boolean {
 export const validateAuthorCreditSection = _.partialRight(
 	// Requires at least one Author Credit row
 	validateMultiple, _.partialRight.placeholder,
-	validateAuthorCreditRow, null, false
+	validateAuthorCreditRow, null, true
 );
 // In the merge editor we use the authorCredit directly instead of the authorCreditEditor state
 export function validateAuthorCreditSectionMerge(authorCredit:AuthorCredit) :boolean {
--- a/src/client/components/author-credit-display.js
+++ b/src/client/components/author-credit-display.js
@@ -22,15 +22,17 @@ import {map as _map} from 'lodash';
 
 
 function AuthorCreditDisplay({names}) {
-	const authorBBID = name.authorBBID ?? name.author?.id;
-	const nameElements = _map(names, (name) => (
-		<span key={`author-credit-${authorBBID}`}>
-			<a href={`/author/${authorBBID}`}>
-				{name.name}
-			</a>
-			{name.joinPhrase}
-		</span>
-	));
+	const nameElements = _map(names, (name) => {
+		const authorBBID = name.authorBBID ?? name.author?.id;
+		return (
+			<span key={`author-credit-${authorBBID}`}>
+				<a href={`/author/${authorBBID}`}>
+					{name.name}
+				</a>
+				{name.joinPhrase}
+			</span>
+		);
+	});
 
 	return (
 		<span>
--- a/src/client/entity-editor/author-credit-editor/author-credit-section.tsx
+++ b/src/client/entity-editor/author-credit-editor/author-credit-section.tsx
@@ -85,7 +85,7 @@ function AuthorCreditSection({
 		</Button>);
 
 	const label = (
-		<ValidationLabel empty={authorCreditPreview.length === 0} error={!isValid}>
+		<ValidationLabel empty={false} error={!isValid}>
 			Author Credit
 		</ValidationLabel>
 	);


src/client/entity-editor/submission-section/actions.ts
```diff
@@ -119,7 +119,11 @@ export function submit(
 	submissionUrl: string
 ): SubmitResult {
 	return (dispatch, getState) => {
-		const rootState = getState();
+		let rootState = getState();
+		dispatch(setSubmitted(true));
+		if (!rootState.getIn(['authorCreditEditor', 'n0', 'author'], null)) {
+			rootState = rootState.set('authorCreditEditor', Map({}));
+		}
```
MonkeyDo, line 126: I'm not sure this is the way to go. Instead of an empty object representing an optional AC, I would prefer if there was a specific boolean passed along with the form data that can be used to ascertain if the AC should be optional (basically the same thing as what we do visually on the page).

That way for example if we one day have the ability to create entities via an API we can control this aspect more declaratively rather than rely on the ambiguous empty object vs. null

test/src/client/entity-editor/validators/test-edition.js
```diff
@@ -441,37 +441,37 @@ function describeValidateForm() {
...
 		const result = validateForm(
 			{
 				...validForm,
-				aliasEditor: INVALID_ALIASES
+				authorCreditEditor: {}
 			},
 			IDENTIFIER_TYPES
 		);
-		expect(result).to.be.false;
+		expect(result).to.be.true;
```
MonkeyDo, line 452: Following my comment above about a specific boolean for optional ACs, we want to be able to test separately that:
- empty AC with `authorCreditDisabled = false` should fail
- empty AC with `authorCreditDisabled = true` should pass
- valid AC with `authorCreditDisabled = true` … ? What should we do here? Should validation fail? Or should we accept it?

MonkeyDo (commented):
Thanks for working on making this better !
Let me know if you have any question about the comments below

test/src/client/entity-editor/validators/test-edition.js
tr1ten: No, since AC field should be disable at that point. passing test will cause unnecessary confusion.

src/client/entity-editor/validators/common.ts
```diff
@@ -184,7 +184,8 @@ export function validateSubmissionSection(
 	);
 }
 
-export function validateAuthorCreditRow(row: any): boolean {
+export function validateAuthorCreditRow(row: any, isOptional = false): boolean {
+	if (!getIn(row, ['author', 'id'], null) && !get(row, 'name', null) && isOptional) { return true; }
```
MonkeyDo, line 188: I'm not sure I understand why we check this in the AC row.
Shouldn't the check be done in validateAuthorCreditSection instead?

src/client/entity-editor/validators/common.ts
```diff
@@ -193,7 +194,7 @@ export function validateAuthorCreditRow(row: any): boolean {
 export const validateAuthorCreditSection = _.partialRight(
 	// Requires at least one Author Credit row
 	validateMultiple, _.partialRight.placeholder,
-	validateAuthorCreditRow, null, false
+	validateAuthorCreditRow, _.partialRight.placeholder, _.partialRight.placeholder
```
MonkeyDo, line 197: Could you explain the double placeholder to me?
I don't understand what's happening

src/client/entity-editor/validators/edition.ts
```diff
@@ -133,11 +133,14 @@ export function validateForm(
...
 	let validAuthorCredit;
+	const authorCreditEnable = isIterable(formData) ? formData.getIn(['buttonBar', 'authorCreditEnable'], true) :
+		get(formData, 'buttonBar.authorCreditEnable', true);
 	if (isMerge) {
 		validAuthorCredit = validateAuthorCreditSectionMerge(get(formData, 'authorCredit', {}));
 	}
 	else {
-		validAuthorCredit = validateAuthorCreditSection(get(formData, 'authorCreditEditor', {}));
+		validAuthorCredit = (authorCreditEnable || size(get(formData, 'authorCreditEditor', {})) === 0) &&
+		 validateAuthorCreditSection(get(formData, 'authorCreditEditor', {}), !authorCreditEnable, authorCreditEnable);
```
MonkeyDo, line 143: Following my request regarding the double placeholder, could you please also explain why we pass `!authorCreditEnable` and `authorCreditEnable`?

MonkeyDo (commented):
This is looking pretty good

I tried the PR locally, but it looks like creation an Edition with name section filled, EG selected and AC disabled is not considered valid (submit button is disabled).
Can you please also try this?

Secondly, we should disable the AC edit button when AC is disabled. Currently with AC disabled the AC field is disabled but I can still click edit to open the modal.

src/client/entity-editor/validators/common.ts
tr1ten: This check is mainly for first row of AC, since in inline editor first row always exist but might not contain any author. which should be valid if optional checkbox is ticked.

src/client/entity-editor/validators/common.ts
tr1ten: This should help - 
second placeholder is for isOptional paramter of validateAuthorCreditRow to check if row can contain null author.
third placeholder is for requiredOneOrMore paramter of validatMultiple which should be used by server to validate optional AC.

src/client/entity-editor/validators/edition.ts
tr1ten: It is something like, AC can either be empty or contain null values for author (first row) in case of optional checkbox ticked.
though rather than using the partials we can go with separate function which should handle all these edge cases between client and server validation of AC.

MonkeyDo: I think that might make it easier to follow the code, and perhaps we can out find why I was hitting that validation bug I mentioned.

PR #872, opened 2022-08-19
Feat: Improve Unified creation form

tr1ten:
This is a follow up PR for unified creation form, mainly for minor UI improvements & bug patches.


MonkeyDo (approved):
Thank you for the valuable improvements !

I'm going to merge this PR as well as #874 so that we can continue moving forward and testing thoroughly. Currently it's tricky to know which PR solves what issue, if something is solved in one PR but not the other, etc.

I want to deploy the feature branch with both PRs merged so that I can continue thorough testing.

PR #873, opened 2022-08-20
Fix(achievements): count consecutive days for timed achievements

tr1ten:
### Problem

[BB-680](https://tickets.metabrainz.org/projects/BB/issues/BB-680): Fun runner and Marathoner achievements are broken

### Solution

Currently, we just return the number of distinct (by date) revisions within specified interval but what we actually want is number of consecutive revisions starting from today upto to given interval. this PR does exactly that. 

_note: we might be able to use some sql magic here but i'm not sure how we would do that._

### Areas of Impact

src/server/helpers/achievement.js

### DIFF
--- a/src/server/helpers/achievement.js
+++ b/src/server/helpers/achievement.js
@@ -401,7 +401,19 @@ export async function getEditsInDays(orm, editorId, days) {
 		and created_at > (SELECT CURRENT_DATE - INTERVAL '${days} days');`;
 
 	const out = await bookshelf.knex.raw(rawSql);
-	return out.rowCount;
+	const dateList = out.rows.map(row => row.created_at);
+	let countDays = 0;
+	let lastDate = new Date();
+	const oneDayMs = 24 * 60 * 60 * 1000;
+	// count number of consecutive days starting from today
+	for (let i = dateList.length - 1; i >= 0; i--) {
+		if (lastDate.getTime() - dateList[i].getTime() > oneDayMs) {
+			break;
+		}
+		countDays++;
+		lastDate = dateList[i];
+	}
+	return countDays;
 }
 
 async function processFunRunner(orm, editorId) {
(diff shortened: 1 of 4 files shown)


src/server/helpers/achievement.js
```diff
@@ -401,7 +401,19 @@ export async function getEditsInDays(orm, editorId, days) {
 		and created_at > (SELECT CURRENT_DATE - INTERVAL '${days} days');`;
 
 	const out = await bookshelf.knex.raw(rawSql);
-	return out.rowCount;
+	const dateList = out.rows.map(row => row.created_at);
+	let countDays = 0;
+	let lastDate = new Date();
+	const oneDayMs = 24 * 60 * 60 * 1000;
```
MonkeyDo, line 407: I can suggest a better way to compare dates (it gets very tricky; using 86400000 milliseconds as "a day" doesn't work for all cases.

We use the `date-fns` library in other parts of the codebase to remove the headache of dealing with time.

The `differenceInCalendarDays` utility should be what you are looking for https://date-fns.org/v2.29.2/docs/differenceInCalendarDays

When you map your rows to dateList, if created_at is a string rather than a Date object you'll need to transform it with `parseISO` from date-fns.

src/server/helpers/achievement.js
```diff
@@ -401,7 +401,19 @@ export async function getEditsInDays(orm, editorId, days) {
 		and created_at > (SELECT CURRENT_DATE - INTERVAL '${days} days');`;
```
MonkeyDo, line 401: I think we should rename this method `getConsecutiveDaysWithEdits` and modify the JSDocs definitions to specify consecutive days and avoid possible confusion

MonkeyDo (commented):
Thanks for looking at this :)

MonkeyDo:
Thanks !

PR #874, opened 2022-08-21
Feat(UF): Submit modal entities directly to server

tr1ten:
This PR changes the current behavior of unified form by submitting newly created entities directly to server. this reduces the duplication of code require for maintaining temporary ids as well as the complexity of redux state. 
My initial thought for late submission (submitting bulk entities) was that it will allow us to easily manipulate the entity attributes and undo the edits but after some discussion with my mentor i realize this might not worth the added complexity.


MonkeyDo (commented):
Thanks a lot for going back and simplifying the code !
It's looking a  lot cleaner, with a lot less special-casing and workarounds.
I know it can feel like "one step forward, two steps back" but in reality I think it's great that you got the project working first before coming back to clean up, refactor and simplify the code.

src/client/entity-editor/submission-section/actions.ts
```diff
@@ -117,81 +117,70 @@ function postSubmission(url: string, data: Map<string, any>): Promise<void> {
 function transformFormData(data:Record<string, any>):Record<string, any> {
 	const newData = {};
 	const nextId = 0;
-	// add new publishers
-	_.forEach(data.Publishers, (publisher, pid) => {
```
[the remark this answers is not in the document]
tr1ten: I don't think there's any harm keeping them, also they help in preview in summary section. though we don't use them in post data since they are already been created and we don't modify them after creation.

src/client/entity-editor/submission-section/actions.ts
```diff
@@ -271,3 +259,33 @@ export function submit(
...
+		const JSONState = rootState.toJS();
+		const entity = {...JSONState, type: entityType};
+		const postData = {
+			0: entity
+		};
+		try {
+			const response = await request.post(submissionUrl).send(postData);
+			const mainEntity = response.body[0];
+			const entityObject = {...initialState,
+				__isNew__: true,
```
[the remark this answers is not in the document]
tr1ten: Yeah, needed only for knowing which entity is being created (edition, eg) and which are being edited (works..).
since entity editor using different routes for both operation, it doesn't need one.

src/client/entity-editor/submission-section/actions.ts
```diff
@@ -117,81 +117,56 @@ function postSubmission(url: string, data: Map<string, any>): Promise<void> {
...
 		});
-		if (series.__isNew__ || _.size(series.seriesSection.seriesItems) > 0) {
-			if (!series.__isNew__) {
-				series.__isNew__ = false;
-				series.submissionSection = {
-					note: _.get(data, ['submissionSection', 'note'])
-				};
-			}
+		// if new items have been added to series, then add series to the post data
+		if (_.size(series.seriesSection.seriesItems) > 0) {
```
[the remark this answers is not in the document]
tr1ten: My bad, forgot about adding old series items. though I will filtering series items so only new items remains so if zero that means nothing has been added to series items.

src/server/routes/entity/process-unified-form.ts
```diff
@@ -242,116 +173,39 @@ export function handleCreateMultipleEntities(
...
+	}
+	// create all entities
+	async function getAchievementPromise() {
+		await Object.keys(body).reduce((promise, entityKey) => promise.then(() => processEntity(entityKey)), Promise.resolve());
+		return processedAchievements;
+	}
 	return handler.sendPromiseResult(
 		res,
-		achievementPromise,
+		getAchievementPromise(),
```
MonkeyDo, line 208: This should probably not be called here, instead passing the promise and letting `sendPromiseResult` resolve the promise safely.
```suggestion
		getAchievementPromise,
```

MonkeyDo (commented):
Last detail, but I think this is ready to merge. Will require more testing, but I want to merge ongoing PRs to the main feature branch with a view to deploy said branch to test.

src/server/routes/entity/process-unified-form.ts
tr1ten: But function does return promise that resolves to entity object, right?

MonkeyDo: Maybe I'm misreading the code, but in short by calling the function here you're executing the promise (without it being in a try-catch block; any error will be uncaught and crash the server).
The idea with the `handler.sendPromiseResult` helper is that it will take care of `await`-ing the promise in a try-catch block and figure out what to do on success or fail.
An async function *is* essentially a promise, but does not necessarily *return* a promise. In this case it returns the `processedAchievements` array.

src/server/routes/entity/process-unified-form.ts
```diff
@@ -242,116 +173,39 @@ export function handleCreateMultipleEntities(
...
-			log.error(err);
-			throw err;
-		}
-	});
-	const achievementPromise = entityEditPromise.then(
-		(entitiesJSON:Record<string, any>) => {
-			const entitiesAchievementsPromise = [];
-			for (const entityJSON of Object.values(entitiesJSON)) {
-				entitiesAchievementsPromise.push(processAchievement(orm, editorJSON.id, entityJSON));
+	const {body}: {body: Record<string, any>} = req;
```
MonkeyDo, line 177: OK, I took a bit of time to rewrite some of the code from promises to async/await, I always get confused when I mix the two styles :)

You'll notice I separate the entity processing and achievement processing a bit, and I am passing an async function (which ~= a Promise) to sendPromiseResult rather than executing the function.
SendPromiseResult will take care of `await`-ing the promise for us in a try-catch block.

Here's what I came up with for the whole function:

```
export function handleCreateMultipleEntities(
	req: PassportRequest,
	res: $Response
) {
	const {orm}: {orm?: any} = req.app.locals;
	const editorJSON = req.user;
	let currentEntity: {
		__isNew__: boolean | undefined,
		aliasSet: {id: number} | null | undefined,
		annotation: {id: number} | null | undefined,
		bbid: string,
		disambiguation: {id: number} | null | undefined,
		identifierSet: {id: number} | null | undefined,
		type: EntityTypeString
	} | null | undefined;


	const {body}: {body: Record<string, any>} = req;
	async function processEntity(entityKey:string) {
		const entityForm = body[entityKey];
		const entityType = _.upperFirst(entityForm.type);
		const deriveProps = _.pick(entityForm, additionalEntityProps[_.camelCase(entityType)]);
		const isNew = _.get(entityForm, '__isNew__', true);
		currentEntity = null;
		if (!isNew) {
			currentEntity = await orm.func.entity.getEntity(orm, entityType, entityForm.id, getEntityRelations(entityType as EntityTypeString));
			if (!currentEntity) {
				throw new FormSubmissionError('Entity with given id not found');
			}
		}
		const {bookshelf} = orm;
		const entityJSON = await bookshelf.transaction((transacting) =>
			processSingleEntity(entityForm, isNew ? null : currentEntity, null, entityType, orm, editorJSON, deriveProps, false, transacting));

		return entityJSON;
	}
	// create all entities
	async function processAllEntities() {
		// For each entity in the request body
		const processedEntities = await Object.keys(body).reduce((promise, entityKey) =>
			promise
				// submit the entity
				.then(() => processEntity(entityKey))
				// process potential achievements for this edit
				.then(entityJSON => processAchievement(orm, editorJSON.id, entityJSON))
		// Empty promise for initial value sent to .reduce
		, Promise.resolve());
		return processedEntities;
	}
		res,
		processAllEntities,
		_bulkIndexEntities
	);
}
```

Full disclaimer, I haven't tried to run the code, so I'll leave that for you to try, and I'll hop on to debugging that failing test ;)

MonkeyDo (commented):
Promises are hard…

src/server/routes/entity/process-unified-form.ts
MonkeyDo: We discussed on IRC, I was totally wrong and I apparently can't read code :p
sendPromiseResult awaits the promise but does not execute it

MonkeyDo (approved):
As discussed on IRC, the main goal of this PR is achieved.

While testing, we've uncovered some issues with the series section in particular, which we want to resolve in another PR.
Consequently the series section has been disabled and should be considered not fully working for this PR.

PR #876, opened 2022-09-03
Fix(BB-538): Allow concurrent relationship edits

tr1ten:
### Problem

Reference ticket [BB-538](https://tickets.metabrainz.org/browse/BB-538) 
Since we use old relationship items to generate the difference array of allRemovedItems & allAddedItems, this cause issue where we don't have updated rel items. 

### Solution

We can just use two new attributes on rel to track removed and added items, this will ensure difference array is independent of old rel items.
Since Orm function need some changes to support this workflow, i will be creating similar PR there.
Edit: [PR](https://github.com/metabrainz/bookbrainz-data-js/pull/289)
### Tasks
- [x] Test showing this behavior
- [x]  ORM PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-fontawesome": "^0.1.11",
     "array-move": "^3.0.1",
-    "bookbrainz-data": "2.14.1",
+    "bookbrainz-data": "3.0.0",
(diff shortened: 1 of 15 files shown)


MonkeyDo:
I'm thoroughly testing the PR locally, seems to be working in most cases.
I made a couple of changes to handle special cases where we add relationships manually on the server side (pre-filling rels with query parameters, and when merging entities)

PR #883, opened 2022-09-23
feat: Modify CB review license text

anshg1214:
This PR modifies the license text, which talks about a choice between two licenses, even though we don't offer an option in this modal.


MonkeyDo:
Thank you! :rocket:

PR #884, opened 2022-09-25
BB-681: refresh search page on same search query

tr1ten:
### Problem

[BB-681](https://tickets.metabrainz.org/browse/BB-681): In Search page, Clicking on "search" should run search again

### Solution

There's currently no way to trigger search manually without causing rendering issues, that is search will trigger only when url params change which is not ideal for all situations so one quick way to resolve this issue is to refresh the whole page when user hit submit with same query. params.

### Areas of Impact

search-page component

### DIFF
--- a/src/client/components/pages/search.tsx
+++ b/src/client/components/pages/search.tsx
@@ -94,6 +94,10 @@ class SearchPage extends React.Component<Props, State> {
 	 * @param {string} type - Entity type selected from dropdown
 	 */
 	handleSearch = (query: string, type: string) => {
+		// if no change in query or type, refresh the page to get new results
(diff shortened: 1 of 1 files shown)


MonkeyDo (changes requested):
I think reloading the page is not ideal UX.

I suggest another option: since the search API call is handled in the PagerElement, we can store a `ref` to that component and call the appropriate method.
So it would be something like 
```
if (query === this.state.query && type === this.state.type) {
	this.pagerElementRef.triggerSearch();
}
```

This would in theory trigger the search again without the need to refresh the page.
What do you think?

MonkeyDo:
Closed in favor of #1003

PR #889, opened 2022-10-17
fix: CritiqueBrainz star rating issues

MonkeyDo:
Having some issues with the package version on install.
Updating the library to the latest version, and renaming the file to .tsx to benefit from Typescript integration and use the right properties for the rating component.


MonkeyDo:
Never mind me, that was not at all the issue I was facing. Please see #890

PR #902, opened 2022-11-21
Fix: Deleting entity crashes the site

tr1ten:
This PR fixes two behavior:-

1. CBRevisionModal causes the site to crash on the deleted entity.
2. Revision Parent doesn't update on deleting an entity.


MonkeyDo (approved):
Thank you !

PR #930, opened 2023-02-12
fix([BB-698]): Line break at the character level in annotations

the-good-boy:
### Problem

This PR addresses [BB-698](https://tickets.metabrainz.org/browse/BB-698). Text in the annotations is breaking after any character (line-break: anywhere) rather than at the word level.

### Solution

Overriding the `word-break: break-all` property which is inherited due to the `pre` element.

### DIFF
--- a/src/client/stylesheets/style.scss
+++ b/src/client/stylesheets/style.scss
@@ -453,6 +453,7 @@ hr.wide {
 
 .annotation-content {
 	white-space: pre-wrap;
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Thank you ! :rocket:

kellnerd:
I think this PR has to be renamed in order to automatically mark https://tickets.metabrainz.org/browse/BB-698 as resolved.

PR #931, opened 2023-02-12
BB-692: Website server doesn't check revision id format before querying PostgreSQL

the-good-boy:
### Problem

When a revision page is requested with a crafted id, for example https://bookbrainz.org/revision/whynot, a PostgreSQL query is made.

It mainly adds a small unneeded load to the database server.

### Solution

Adding a middleware function which checks whether the parameter id is numeric or not.

### DIFF
--- a/src/server/helpers/middleware.ts
+++ b/src/server/helpers/middleware.ts
@@ -192,6 +192,17 @@ export function loadEntityRelationships(req: $Request, res: $Response, next: Nex
 		})
 		.catch(next);
 }
+
+export function checkValidRevisionId(req: $Request, res: $Response, next: NextFunction, id: string) {
+	function isNumeric(value) {
+		return (/^-?\d+$/).test(value);
+	}
+	if (!isNumeric(id)) {
+		return next(new error.BadRequestError(`Invalid revision id: ${req.params.id}`, req));
+	}
+	return next();
+}
+
 export async function redirectedBbid(req: $Request, res: $Response, next: NextFunction, bbid: string) {
 	if (!commonUtils.isValidBBID(bbid)) {
 		return next(new error.BadRequestError(`Invalid bbid: ${req.params.bbid}`, req));
--- a/src/server/helpers/middleware.ts
+++ b/src/server/helpers/middleware.ts
@@ -194,10 +194,8 @@ export function loadEntityRelationships(req: $Request, res: $Response, next: Nex
 }
 
 export function checkValidRevisionId(req: $Request, res: $Response, next: NextFunction, id: string) {
-	function isNumeric(value) {
-		return (/^-?\d+$/).test(value);
-	}
-	if (!isNumeric(id)) {
+	const idToNumber = _.toNumber(id);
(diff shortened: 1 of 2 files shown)


src/server/helpers/middleware.ts
```diff
@@ -192,6 +192,17 @@ export function loadEntityRelationships(req: $Request, res: $Response, next: Nex
 		})
 		.catch(next);
 }
+
+export function checkValidRevisionId(req: $Request, res: $Response, next: NextFunction, id: string) {
+	function isNumeric(value) {
+		return (/^-?\d+$/).test(value);
+	}
```
MonkeyDo, line 199: We can use the lodash function [isNumber](https://lodash.com/docs/4.17.15#isNumber) instead of defining our own

MonkeyDo (commented):
Looks good, thanks !
Small comment about using a library we use for this type of thing

src/server/helpers/middleware.ts
the-good-boy: We can't use that function for a string input. I found no clean way of doing it with lodash(atleast not cleaner than this). 

I thought of doing something like `_.isNaN(parseFloat(id))` , but then it will run a query in case someone uses urls like `https://bookbrainz.org/revision/33.6`. Can you think of some better way to handle this?

MonkeyDo: Thanks, I hadn't seen that, I read the lodash docs too fast :p

I had a go with toInteger, which I think could be helpful for our case.
If the input is not a number (string, null, undefined, etc.) it will return `0`, which we can consider as a signal for raising an error (there is no revision #0)
If the input is a floating point number it will be converted to an integer safely:

image

What do you think?

the-good-boy: I think it is probably not a good idea to show `https://bookbrainz.org/revision/122882` for URLs like `https://bookbrainz.org/revision/122882.8573`. Instead of telling the user that the revision id format is wrong, it will redirect to another revision id (which may or may not be the expected revision). Thoughts?

MonkeyDo: I'm a bit torn.
My main worry is about the readability of the code. For example, why is there a provision for an optional dash - at the beginning of the id?
I assume it's because this regexp comes from stackoverflow or something like that and is supposed to support negative numbers, which would make sense.
That being said it illustrates my point that it's harder to read and understand what a regexp does, and I'd rather go with the toInteger version that is pretty clear.

As for the decimal point numbers issue, you're right we can probably improve it.
How about simply using lodash's `toNumber` in combination with `isInteger`?
```
const idToNumber = _.toNumber(id);
if(!_.isInteger(idToNumber)){
…throw error
```

the-good-boy: You're right. The optional dash shouldn't have been there. I also now realize that we can get away without using the regex. Thanks for the suggestions!

MonkeyDo (approved):
Looking great, thanks !

And thanks also for the discussion on the implementation details, we managed to find a good way to do it !

MonkeyDo:
@the-good-boy Could I also ask you to create another PR adding a test for this case please?

the-good-boy:
@MonkeyDo will do

PR #932, opened 2023-02-20
chore: Update redis package to v4

MonkeyDo:
We use 'legacyMode' in order to play well with the connect-redis package (which links ExpressJS sessions to Redis). This also means we don't need to change the structure of the `createClient` args object. See  https://github.com/tj/connect-redis/pull/337#issuecomment-981754350

Also of interest, the redis package changelog: https://github.com/redis/node-redis/blob/redis%404.0.0/packages/client/CHANGELOG.md

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "react-sortable-hoc": "^2.0.0",
     "react-sticky": "^6.0.1",
     "react-tooltip": "^4.2.21",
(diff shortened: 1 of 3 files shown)


leftmostcat (approved):
Seems sensible, particularly given the constraints.

MonkeyDo:
Thanks for the sanity check @leftmostcat !

PR #933, opened 2023-02-21
BB-703: Series of authors are not displayed correctly

siddhart1o1:
### Problem
Series of authors are not displayed correctly 
Gender always shows a question mark — even if is set.
Type is empty (and presumably shouldn't be there at all).
Disambiguationshow as "(undefined)".

Solution
Made changes in the Author table Component, for ```gender```, ```authorType``` and ```disambiguation``` fields


src/client/components/pages/entities/author-table.js
```diff
@@ -30,8 +30,8 @@ function AuthorTableRow({author, showAddedAtColumn, showCheckboxes, selectedEnti
 	const name = getEntityLabel(author);
 	const disambiguation = getEntityDisambiguation(author);
 	const number = author.number || '?';
-	const authorType = author.authorType ? author.authorType.label : '?';
-	const gender = author.gender ? author.gender.name : '?';
+	const authorType = author.authorType || author.authorType?.label || author?.type || '?';
```
MonkeyDo, line 33: author.type here refers to the entity type rather than the author type (so it will always be "Author").
I know it's not very clear and some early choices of column names were not the smartest :p

Also, I would suggest this order:
```suggestion
	const authorType = author.authorType?.label || author.authorType || '?';
```

src/client/components/pages/entities/author-table.js
```diff
@@ -30,8 +30,8 @@ function AuthorTableRow({author, showAddedAtColumn, showCheckboxes, selectedEnti
...
+	const gender = author.genderId === 1 ? "Male" : author.genderId ? "Female" : "?";
```
MonkeyDo, line 34: As I mentioned on IRC, the right way to do this is to use the [`loadGenders` middleware](https://github.com/metabrainz/bookbrainz-site/blob/816f687b34e2653fbf27eda1488f99b195c73cc0/src/server/helpers/middleware.ts#L69) on the server side (in the [series get endpoint](https://github.com/metabrainz/bookbrainz-site/blob/0e6807c9b6/src/server/routes/entity/series.js#L138)) to load  the genders from the database (which also attaches them to the response 'locals' props).
Then you can pass the genders to the page props (see [example here](https://github.com/metabrainz/bookbrainz-site/blob/0e6807c9b6151dc166e201a3aa2d6425db2902b9/src/server/routes/register.js#L68-L70)), and use that to get the name associated to the `author.genderId`:

something along the lines of `genderOptions.find((option) => option.id === author.genderId).name`

MonkeyDo (changes requested):
Thanks for opening this PR !

Don't hesitate to ask if you have question, I realize a lot of the things in the codebase are not obvious when starting on this project :)

src/server/routes/entity/entity.tsx
```diff
@@ -150,9 +150,9 @@ export function displayEntity(req: PassportRequest, res: $Response) {
 		if (EntityComponent) {
 			const props = generateProps(req, res, {
 				alert,
-				identifierTypes
+				identifierTypes,
+				genderOptions: res.locals.genders
```
MonkeyDo, line 154: My linter is complaining that the keys are not in alphabetical order. Not that I care that much, but…
```suggestion
				genderOptions: res.locals.genders,
				identifierTypes
```

MonkeyDo (changes requested):
Looking good !
Thanks for the improvements, that's working quite nicely !

Last little bits of cleanup and we're ready to merge !

src/client/helpers/entity.tsx
```diff
@@ -227,10 +227,9 @@ export function authorCreditToString(authorCredit) {
...
 export function getEntityDisambiguation(entity) {
-  if (entity.disambiguation) {
-	if(entity.disambiguation.comment){
+	if(entity.disambiguation?.comment){
 		return <small>{` (${entity.disambiguation.comment})`}</small>;
-	}else{
+	} else if(entity.disambiguation){
 		return <small>{` (${entity.disambiguation})`} </small>;
 	}
   }
```
MonkeyDo: There is now a superfluous closing curly bracket } that will make this file fail.

MonkeyDo (changes requested):

siddhart1o1 (commented):
requested  changes

MonkeyDo (approved):
:rocket: Thank you !

PR #936, opened 2023-02-27
BB-673: Rewrite middleware file using async/await syntax

RohanSasne:
### Problem

[BB-673](https://tickets.metabrainz.org/browse/BB-673):
Across the codebase we have asynchronous code written using promises, which is hard to read and to maintain and error-prone.

### Solution

### Areas of Impact

The changes were reflected in  `src/server/helpers/middleware.ts`


src/server/helpers/middleware.ts
```diff
@@ -291,7 +290,7 @@ export async function validateCollectionParams(req, res, next) {
 		return next(new error.BadRequestError('Invalid collection name: Empty string not allowed', req));
 	}
 	const entityTypes = _.keys(commonUtils.getEntityModels(orm));
-	if (!entityTypes.includes(_.upperFirst(_.camelCase(entityType)))) {
+	if (!entityTypes.includes(.upperFirst(.camelCase(entityType)))) {
```
kellnerd, line 293: Unrelated change, removing the reference to the lodash library should break the code here.

src/server/helpers/middleware.ts
```diff
@@ -32,25 +32,24 @@ interface $Request extends Request {
...
+			const model = orm[modelName];
+			const results = await model.fetchAll({withRelated: [...relations]});
+			const resultsSerial = results.toJSON();
+			res.locals[propName] =
+				sortFunc ? resultsSerial.sort(sortFunc) : resultsSerial;
+			next();
+			return null;
+		};
+	} catch (error) {
+		throw error;
```
kellnerd, line 48: This doesn't make sense, why should you catch an error just to immediately throw it again?
As far as I can see this would still crash the server, the equivalent of the old code should be (untested):
```suggestion
		next(error);
```

kellnerd (changes requested):

RohanSasne:
Hey @kellnerd , I realised the errors in the code, made the required changes, please have a look once

src/server/helpers/middleware.ts
```diff
@@ -32,25 +32,24 @@ interface $Request extends Request {
...
+			const results = await model.fetchAll({withRelated: [...relations]});
+			const resultsSerial = results.toJSON();
+			res.locals[propName] =
+				sortFunc ? resultsSerial.sort(sortFunc) : resultsSerial;
+			next();
+			return null;
+		};
+	} catch (error) {
+		next(error);
+	}
```
kellnerd, line 49: Oops, didn't notice that earlier, but the catch block has to be inside the returned middleware function of course.
Probably because I've reviewed the diff on GitHub, where the display isn't as comfortable as in a local editor.
```suggestion
	return async function loaderFunc(req: $Request, res: $Response, next: NextFunction) {
		try {
			const {orm}: any = req.app.locals;
			const model = orm[modelName];
			const results = await model.fetchAll({withRelated: [...relations]});
			const resultsSerial = results.toJSON();
			res.locals[propName] =
				sortFunc ? resultsSerial.sort(sortFunc) : resultsSerial;
			next();
			return null;
		} catch (error) {
			next(error);
		}
	};
```

kellnerd (changes requested):
@RohanSasne: I'm just a BB community member who is watching this repository, not a developer who is authorized to merge your PR. So you don't have to see my comments as final or authoritative. I haven't even tried to run the code from this PR, so I'm just mentioning a few things which I noticed :smiley:

P.S. Have you already joined the #bookbrainz IRC channel where the community and developers chat? See https://wiki.musicbrainz.org/Development/Summer_of_Code/Getting_started and https://wiki.musicbrainz.org/Communication/IRC.

src/server/helpers/middleware.ts
```diff
@@ -131,67 +130,67 @@ export async function loadWorkTableAuthors(req: $Request, res: $Response, next:
...
+					'relationships.source',
+					'relationships.target',
+					'relationships.type.attributeTypes',
+					'relationships.attributeSet.relationshipAttributes.value',
+					'relationships.attributeSet.relationshipAttributes.type'
+				]
+			});
+
+		await addRelationships(entity, relationshipSet, orm);
+
```
MonkeyDo, line 187: I'm getting another ESLint error ([callback-return](https://github.com/mysticatea/eslint-plugin-node/blob/v11.1.0/docs/rules/callback-return.md)), which we can solve easily by moving this `next();` a coupl eof lines, right after the end of the catch block.

The same applies to the `loaderFunc` function on line 36

MonkeyDo (changes requested):
Thanks, looking good !
A couple of details to resolve below, but functionally this looks great already

RohanSasne:
Hey @MonkeyDo , thanks a lot for pointing out the changes, I have made the necessary changes, also all the checks were passed, can you please review this PR once :)

MonkeyDo (approved):

PR #939, opened 2023-02-27
Enforce Unix end-of-line convention.

pbryan:
### Problem
Windows systems cloning repo would insert CRs to end of lines. This causes shell scripts executing in WSL to fail.

### Solution
Enforce Unix end-of-line convention; IDEs can handle this quite well.

### Areas of Impact
Git client will avoid adding CRs, and will strip CRs from text files on commit.


MonkeyDo (approved):
Thank you for looking into and testing this !

PR #940, opened 2023-02-27
BB-712: Change alias delimiter to middle dot 

kellnerd:
### Problem

[BB-712](https://tickets.metabrainz.org/browse/BB-712): The display of multiple aliases looks confusing [in certain cases](https://beta.bookbrainz.org/work/03ab2c2e-ff8a-4e10-bb07-f0afb6189dc8):

before

### Solution

* [Render aliases as an inline list](https://github.com/metabrainz/bookbrainz-site/commit/e712e6079b6b79833cea1ae8f0911f18e08f8541): This makes it clearer where an alias begins and ends if the alias contains the text which is used as list separator. This change alone causes no visible difference.
* Use a character which is less likely to be used within an alias than a comma (middle dot, as agreed on IRC)
* Make the dot a bit more visible by increasing its font weight

after

### Areas of Impact

The title section of all entity display pages.

### DIFF
--- a/src/client/stylesheets/style.scss
+++ b/src/client/stylesheets/style.scss
@@ -383,6 +383,20 @@ a.contact-text:visited {
 	height: 10px;
 }
 
+/* Style inline alias lists the same way as the previously used h4 elements */
+ul.inline-aliases {
+	@extend h4;
+	padding: 0;
+	li {
+		display: inline;
+	}
+	li::after {
+		content: ", ";
+	}
+	li:last-child:after {
+		content: "";
+	}
+}
 
 /* Text wrapping in list-group-items (for long revision notes, for example) */
 /* ============= */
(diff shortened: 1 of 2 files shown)


src/client/stylesheets/style.scss
```diff
@@ -383,6 +383,21 @@ a.contact-text:visited {
...
+	li {
+		display: inline;
+	}
+	li::after {
+		content: " · ";
+		font-weight: bolder;
+	}
+	li:last-child:after {
+		content: "";
+	}
```
MonkeyDo, line 399: I think we might be able to refine the selector to remove the override you currently have for the last child:

`li:not(:last-child)::after`

MonkeyDo (commented):
Looking good, thank you !

Small css optimization suggestion

src/client/stylesheets/style.scss
kellnerd: Good point! I also think this should work, but I will test it later nevertheless.

MonkeyDo (approved):
Thank you !

PR #943, opened 2023-03-02
BB-530: Show Wikipedia extracts for entities

kellnerd:
### Problem

[BB-530](https://tickets.metabrainz.org/browse/BB-530): Showing the first paragraph from the Wikipedia page of an entity makes e.g. author pages more informative without requiring our editors to write a descriptive annotation for them.

### Solution

* Client asynchronously loads the extract from a new server route `/wikidata/:id/wikipedia-extract`
* Client renders the extract as a new component (content and URL)
* Server queries the Wikidata API to get all available Wikipedia articles
* Server loads the extract in the first available preferred language (currently hardcoded to English) from the Wikipedia API

screenshot

### TODO

- [x] Handle other entity types (insert the new component, detect Wikidata identifier type for them)
- [x] Determine preferred languages (from Accept-Language request header and alias languages)
- [ ] Handle entities with multiple Wikidata identifiers (e.g. [!Mediengruppe Bitnik](https://bookbrainz.org/author/37e2bdfe-28b1-456f-8723-eecd80957937)) (irrelevant edge case?)
- [ ] Collapse extract if it is too long (refactor annotation component) (will be done later)
- [x] Cache API responses (~share cache with MBS?~ use BB's own Redis instance)
- [x] Already include cached extracts in the initial page properties
- [x] Sanitize HTML from the API before injection
- [x] Follow https://www.mediawiki.org/wiki/API:Etiquette

### Areas of Impact

* Author display pages have a new optional section for the Wikipedia extract before the annotation
* New server route to deliver Wikipedia extracts
* https://github.com/metabrainz/bookbrainz-data-js/pull/304


kellnerd:
Support for all entity types is completed now!

After stumbling across a1f26415068c6c6a4506b16dc512d7cf2f826bb9, I finally decided to improve the type definitions for entity objects, independent from the Bookshelf models because of their limitations.
I started with the things that I needed for this PR in https://github.com/kellnerd/bookbrainz-data-js/tree/entity-typings, if you want to watch the progress I can also open a draft PR for that branch.

P.S. While I was extracting the Wikidata identifier type IDs for all entity types from the latest dump, I saw that some of the descriptions still use the old entity type names "Creator" and "Publication".

src/server/app.js
```diff
@@ -89,7 +89,12 @@ else {
 }
 app.use(express.static(path.join(rootDir, 'static')));
 
-/* Set up sessions, using Redis in production and default in-memory for testing environment*/
+// Set up Redis, which is used for sessions and to cache requests to external APIs
+// eslint-disable-next-line no-console
+redisClient.connect().catch(redisError => { console.error('Redis error:', redisError); });
```
MonkeyDo, line 94: Previously, the `connect` call was in the block below, making sure we don't need redis when running tests.
If I'm not mistaken, now it will try to connect in all cases, which might just mean we get an extra error when running tests.
Probably not a big deal in itself, but since we're not expecting to have an instance of redis in that case, it would probably be better to keep the connection in the `if (process.env.NODE_ENV !== 'test')` block

kellnerd: Agreed, I had been thinking that it might be causing issues if the cache is not always available, but I will test how the Wikipedia helpers behave when `redisClient` is not connected and adapt their code if necessary.
And I think I had also mixed up 'test' and 'development' when things were not working as expected yesterday and I wanted to exclude one potential issue...

kellnerd: By the way, is there any specific reason to use `console.log` instead of `debug` [here](https://github.com/metabrainz/bookbrainz-site/commit/e949ae64c4cc90385eb6592ef95dfb4f9e815e77#diff-c9e676a021d3cb1d0e5e7ae832276cbfebab683b325852042a9b0aa9b9e8d11cR72)?

MonkeyDo: > if the cache is not always available

*A* cache should always be available. When we don't set `sessionOptions.store`, the cache will be in memory instead (which is why we should remove the redis connection).

> any specific reason to use console.log

None at all, that's just unclean code from me !
I assume I copied the example connection string from the library examples and didn't correct it to use `debug`

kellnerd: > A cache should always be available

I was more worried about the API cache not being available than the session cache, but I've wrapped the Redis commands with a condition in 6538335e9f7af3fe1dd0b5c4665572ae68be04f4.

> I assume I copied the example connection string from the library examples and didn't correct it to use debug

Logging has now been standardized in 5d873671eb54030e7071bde465b73507bf5eba2a, after I've squashed my previous Redis refactoring commits into 239977633d3d9d331f3d617e0fbec89033ec39b1. I've also reused the same Redis setup code for the BB API app and extracted the whole duplicated session setup code into a new module in 6c4f3b96a67d585940aeebe334ca8a0516327b87.

docker-compose.yml
```diff
@@ -9,6 +9,8 @@ services:
       context: ./
       dockerfile: Dockerfile
       target: bookbrainz-dev
+      args:
+        - GIT_COMMIT_SHA=${GIT_COMMIT_SHA}
```
kellnerd, line 13: Maybe this change is useless, you can already pass this argument directly to the docker-compose command.
Originally I had also included the first example command from the commit message in `develop.sh`, but I have decided to drop this again, because this would lead to the entire image being built from scratch each time `develop.sh`  is run for a different commit.

kellnerd: The more I think about this, the more I'm determined to drop the commit with this change...
It also results in a `WARN[0000] The "GIT_COMMIT_SHA" variable is not set. Defaulting to a blank string.` Docker warning because I intentionally want to have no fallback at this level in order to be able to handle the different fallback cases separately:
1) Display the repository link in the footer as "unknown revision" but ...
1) ... append "master" to the GH tree base URL to still have a valid link.
1) Fail to extract a version number and fallback to "latest".
    Thinking about this one again, it might make more sense to fallback to "dev" here, as it is used in the user agent and could be any checked out commit. What do you think?

src/common/helpers/cache.ts
```diff
@@ -0,0 +1,30 @@
...
+const redisPort = _get(config, 'session.redis.port', 6379);
+const redisClient = createClient({
+	url: `redis://${redisHost}:${redisPort}`
+});
+
+export async function getCachedJSON<T>(cacheKey: string) {
+	if (redisClient.isReady) {
+		const cachedValue = await redisClient.get(cacheKey);
+		if (cachedValue) {
+			return JSON.parse(cachedValue) as T;
```
MonkeyDo, line 17: I always wrap `JSON.parse` calls in a try…catch block, I've been bitten by malformed JSON too many times !

src/common/helpers/cache.ts
```diff
@@ -0,0 +1,30 @@
...
+		}
+	}
+	return null;
+}
+
+export function cacheJSON(cacheKey: string, value: any, options: {expireTime: number}) {
+	if (redisClient.isReady) {
+		return redisClient.set(cacheKey, JSON.stringify(value), {EX: options.expireTime});
```
MonkeyDo, line 25: Similarly, probably a good idea to wrap the `stringify` call in a try…catch block first before setting int in cache.

src/client/components/pages/entities/wikipedia-extract.tsx
```diff
@@ -0,0 +1,95 @@
...
+		if (!this.state.extract) {
+			const wikidataId = getWikidataId(this.props.entity);
+			if (!wikidataId) {
+				return;
+			}
+
+			const aliasLanguages = getAliasLanguageCodes(this.props.entity);
+			const preferredLanguages = uniq(['en', ...aliasLanguages]);
+
+			getWikipediaExtractForWikidata(wikidataId, preferredLanguages).then((result) => {
```
MonkeyDo, line 65: This should also have a `.catch(…` clause, in case the fetch request or the subsequent json parsing fail

src/client/components/pages/entities/wikipedia-extract.tsx
```diff
@@ -0,0 +1,95 @@
...
+	render() {
+		const {extract, article} = this.state;
+		const licenseUrl = 'https://creativecommons.org/licenses/by-sa/3.0/';
+
+		return extract ? (
+			<Row className="wikipedia-extract">
+				<Col>
+					<h2>Wikipedia</h2>
+					{/* eslint-disable-next-line react/no-danger */}
+					<div dangerouslySetInnerHTML={{__html: DOMPurify.sanitize(extract)}}/>
```
MonkeyDo, line 82: Yes, thank you for sanitizing :) 💯

src/server/info.ts
```diff
@@ -0,0 +1,15 @@
+
+export const repositoryName = 'bookbrainz-site';
+
+export const repositoryUrl = 'https://github.com/metabrainz/bookbrainz-site/';
```
MonkeyDo, line 4: Not sure it's particularly necessary, but any reason you didn't use this?
```suggestion
export const repositoryUrl = `https://github.com/metabrainz/${repositoryName}/`;
```

src/server/helpers/i18n.ts
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Parses an Accept-Language header to obtain the language codes, optional subtags and weights.
+ * @param {string} acceptLanguage - Accept-Language header value.
+ * @returns {Array} Parsed languages, sorted by weight in descending order.
+ */
+export function parseAcceptLanguage(acceptLanguage: string): AcceptedLanguage[] {
+	return acceptLanguage
+		.split(',')
+		.map((value) => {
+			const match = value.match(/(?<tag>[a-zA-Z]{2,3})(?:-(?<subtag>[\w-]+))?(?:;q=(?<weight>[01](?:\.[0-9]+)?))?/);
```
MonkeyDo, line 37: * Chef's kiss *

src/server/helpers/wikimedia.ts
```diff
@@ -0,0 +1,166 @@
...
+
+	const apiUrl = new URL('https://www.wikidata.org/w/api.php');
+	apiUrl.search = new URLSearchParams({
+		action: 'wbgetentities',
+		format: 'json',
+		ids: wikidataId,
+		props: 'sitelinks'
+	}).toString();
+
+	const response = await request.get(apiUrl.href)
```
MonkeyDo, line 87: This can fail; like my other comments, I'd like to see it wrapped in a try…catch block.

src/server/helpers/wikimedia.ts
```diff
@@ -0,0 +1,166 @@
...
+		format: 'json',
+		formatversion: '2',
+		prop: 'extracts',
+		// eslint-disable-next-line sort-keys -- `exintro` only allowed with `prop: 'extracts'`
+		exintro: '1',
+		redirects: '1',
+		titles: article.title
+	}).toString();
+
+	const response = await request.get(apiUrl.href)
```
MonkeyDo, line 158: Same here, the request can fail and take the site down with it (arguably in edge-cases, but still…)

MonkeyDo (changes requested):
Great job! 🥇 

I lack the level of familiarity with the Wikimedia APIs that would help me ensure I don't miss any gotchas, but I did some extensive testing locally and everything is working fine, and the code also looks good to me.

I'll deploy this to test.BB this week to test it in live conditions.

MonkeyDo:
I've just published the new version (`4.0.0`) of the bookbrainz-data package as discussed: https://github.com/metabrainz/bookbrainz-data-js/releases/tag/v4.0.0

docker-compose.yml
kellnerd: There are two cases where I want the version extraction to fail deliberately:
1. GIT_COMMIT_SHA is not set inside the container. This happens in development unless it is explicitly set as Compose argument or as a local environment variable (with this patch).
2. GIT_COMMIT_SHA contains a real commit hash and not a version tag which can be parsed. This could happen depending on how the production image is built.

In both of these cases we agreed that the user agent should fall back to "dev" in absence of a version number.
But I explicitly did not want to set the fallback in docker-compose already as that would also cause the GIT_COMMIT_SHA fallbacks for the footer (https://github.com/metabrainz/bookbrainz-site/pull/943/commits/871ff902e5488cc4837192b26c4ce01a0d856b2a) to be ignored and lead to a broken link https://github.com/metabrainz/bookbrainz-site/tree/dev.

As I already mentioned above, the standard way to set GIT_COMMIT_SHA is via the Compose args, which is done when building a production image.
So my remaining question is whether I should drop the possibility to also set this via a local environment variable (https://github.com/metabrainz/bookbrainz-site/pull/943/commits/3a046a2003302af259120e9d0f8f6950f62f43f6)? I would say yes per my comments above.

src/server/info.ts
kellnerd: I had considered doing this when I moved the constant into this module but then I thought that it only obfuscates the URL without actually simplifying anything.
As long as the constants are declared on consecutive lines, I would rather keep them as is.

src/common/helpers/cache.ts
kellnerd: While I usually let errors propagate until they reach the middleware layer, you're right that it makes more sense to handle them directly in the case of the cache.
In 0421c23 I will log an error and then fail deliberately, i.e. treat this case as a cache miss.

src/common/helpers/cache.ts
kellnerd: I have wrapped the whole line in a catch block because that's simpler and in this case we don't want to write to the cache anyway, see also 0421c23.

src/client/components/pages/entities/wikipedia-extract.tsx
kellnerd: Good catch, I hadn't really thought about this scenario because an exception doesn't really cause any harm here (because it happens only inside a browser when we're already done).

Did you have any specific error handling in mind?
I'm asking because I had to disable ESLint in order to be able to log the error to `console` in 73e0201 (the `log` module is only for Node.js) and I don't think we want to do more for a non-required component that fails to load (and that should only happen rarely).

src/server/helpers/wikimedia.ts
kellnerd: In theory I could already catch these errors inside the Wikimedia helper functions but at this stage it is difficult to decide how they should be treated. There's even one case where I considered it to be better to throw an error myself:
https://github.com/metabrainz/bookbrainz-site/blob/b0759410469e0f5cd5085cfc7a5612d9b618dd80/src/server/helpers/wikimedia.ts#L92-L94
So my intention was to let these errors propagate until they reach the middleware layer where they are actionable and catch them there.
I had already done this for the JSON route but forgot to handle this in the (newer) pre-loader middleware as well, which I did now in 307876a.

src/server/helpers/wikimedia.ts
kellnerd: Also handled by the same catch blocks at router level as per my previous comment.

kellnerd:
Now that I've also converted the React component class into a functional component, we should be ready for the final review. From my side, only one question remains from my comments above.

Also, I have intentionally not bumped the bookbrainz-data package to v4 yet because that will probably cause conflicts with the master branch's lock file. Do you want me to merge master into my feature branch another time right before we merge it into master? Or should we avoid the double merge and directly commit the version bump to master after this PR has been merged?

src/client/components/pages/entities/wikipedia-extract.tsx
MonkeyDo: Yes, console.error or console.log is fine (that is, as an exception rather than the rule).
We don't currently have a better reporting system in place.

src/server/helpers/wikimedia.ts
MonkeyDo: Great, that's totally fine now.

src/server/info.ts
MonkeyDo: Little added value indeed.

docker-compose.yml
kellnerd: Commit has been dropped, I've force-pushed the branch.

kellnerd:
Merging master and updating `bookbrainz-data` has been successful, the site is still running :grin:
There were a few yarn warnings, but they look unrelated to the current changes in yarn.lock (which are related to our Bookshelf/Knex update):
```
warning " > react-select-fast-filter-options@0.2.3" has incorrect peer dependency "react-select@^1.0.0-beta14".
warning "swagger-jsdoc > swagger-parser > @apidevtools/swagger-parser@10.0.2" has unmet peer dependency "openapi-types@>=7".
warning " > bootstrap@4.6.1" has unmet peer dependency "jquery@1.9.1 - 3".
warning " > bootstrap@4.6.1" has unmet peer dependency "popper.js@^1.16.1".
```

MonkeyDo (approved):
Absolutely great job on this feature, thank you very much !

Sorry it took two months to get it merged, but you did solve some big issues on the way, for which I am also very grateful !

PR #944, opened 2023-03-03
Update connect-redis 

MonkeyDo:
I borked the redis connection in #932 and needed to revisit redis.
In the meantime, connect-redis v4 came out which removes the need for legacyMode in node-redis.
This PR updates the required bits and pieces to update the whole chain.

Also see https://github.com/redis/node-redis/blob/redis%404.0.0/docs/v3-to-v4.md and https://github.com/tj/connect-redis/releases/tag/v7.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "chartjs-adapter-date-fns": "^1.0.0",
     "classnames": "^2.3.2",
     "compression": "^1.7.4",
-    "connect-redis": "^6.0.0",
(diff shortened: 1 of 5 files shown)


MonkeyDo:
I need to merge this urgently because we have new contributors everyday at the moment and the master branch is broken.
Will deploy to test.bookbrainz.org next week to confirm in production and ensure there is no other work needed.

PR #946, opened 2023-03-06
fix(call-to-action): properly encode query parameters once

kellnerd:
### Problem

Double encoding the seeding query parameters (e.g. http://localhost:9099/author/create?name=Rhythm%2520%2526%2520Beauty%253A%2520The%2520Art%2520of%2520Percussion) is entirely unnecessary.

### Solution

Properly encode the query parameter once (e.g. http://localhost:9099/author/create?name=Rhythm+%26+Beauty%3A+The+Art+of+Percussion).
Now we can also drop the `decodeUrlQueryParams` middleware (which was undoing the second encoding pass), Express.js is already doing one decoding pass for us automatically.

### Areas of Impact

"Are we missing an entry?" section of the search results page.

### DIFF
--- a/src/client/components/pages/parts/call-to-action.js
+++ b/src/client/components/pages/parts/call-to-action.js
@@ -34,11 +34,10 @@ const {Button, ButtonGroup} = bootstrap;
  * the 'CallToAction' component.
  */
 function CallToAction(props) {
-	// double encoding to prevent browser from automatically decoding it
-	const nameQueryParameter = props.query ? `?name=${encodeURIComponent(encodeURIComponent(props.query))}` : '';
+	const seedingParameters = new URLSearchParams({name: props.query});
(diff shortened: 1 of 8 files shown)


MonkeyDo (approved):
Perfect, thanks a lot !

MonkeyDo:
Making a quick note for posterity that the double encoding was introduced in #819

kellnerd:
Oh yes, I only mentioned the source of the issue on IRC but forgot to mention it here.
Good catch!

Also (while we are already at clarifying this PR), removing one of the `encodeURIComponent` calls would've been sufficient technically, but I decided to use the cleaner way with `URLSearchParams` instead of manually assembling the query string.
And the case that the query is empty should not happen usually, so I decided to skip that special treatment since seeding an empty name causes no harm.

PR #947, opened 2023-03-06
fix(display): always use Col as container inside Row flex-boxes 

kellnerd:
### Problem

Previously the list of collections was displayed inline with the heading:
before

### Solution

Wrap heading and list into a `<div>` container which serves as flex item of the `<Row>` flex-box:
after

I'm wondering whether I should have used a `<Col>` instead of a `<div>` which seems to be the semantically correct way. For now I have decided against it because the "Relationships" and "Identifiers" sections directly above are also using a `<div>` which results in [less unused space](https://user-images.githubusercontent.com/52860029/223233846-8ccd9bc2-6a0e-4777-a05c-7a826547e4d5.png) to the left than a `<Col>` (which is used by "Annotation" and the "Works" section of an Edition).
But I think we should make this consistent in the end?

Another related question: Should we really show the headings for empty sections? I would either leave them away or show at least a placeholder text.
Maybe even handle that on a case by case basis as an empty relationship section is a call to action while an empty collection section is usually not.

### Areas of Impact

"Related Collections" section  of all entity display pages.

### DIFF
--- a/src/client/components/pages/entities/links.js
+++ b/src/client/components/pages/entities/links.js
@@ -27,7 +27,7 @@ import React from 'react';
 
 
 const {filterOutRelationshipTypeById} = entityHelper;
-const {Row} = bootstrap;
+const {Row, Col} = bootstrap;
 
 function EntityLinks({entity, identifierTypes, urlPrefix}) {
 	// relationshipTypeId = 10 refers the relation (<Work> is contained by <Edition>)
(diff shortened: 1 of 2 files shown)


MonkeyDo:
I think it makes sense to use a `<Col>`  for every row that is not currently using one.
If we don't like the padding we can revisit the css style of said columns at a later date.

Do you want to do that in this PR?

As for empty sections, placeholder text would be a good improvement. It usually ends up being the right place to add a call to action as you mention, and also means we do show all the sections, even empty ones (which means more homogenous pages)

kellnerd:
> Do you want to do that in this PR?

Yeah, I could do that later once I'm back in my development setup. I would also reword the PR's title to cover the new, broader topic.

> As for empty sections, placeholder text would be a good improvement.

So, would you do this for all sections? I think it makes sense for Relationships and Identifiers, but I would hide Related Collections, Annotation and the coming Wikipedia section if empty.

MonkeyDo:
Thank you !

> So, would you do this for all sections? I think it makes sense for Relationships and Identifiers, but I would hide Related Collections, Annotation and the coming Wikipedia section if empty.

Makes sense for Annotation and Wikipedia, but we could still decide to add a call to action to create a new collection with the current entity. Not suggesting to do that in this PR, but it would make sense to me to show it.

kellnerd:
Oh, I forgot to mention that I'm already done with using `Col` elements inside every `Row`. I have added these in the same file which declares the Row elements because it seemed weird for a component (which might be embedded elsewhere) to return a `Col` element instead of a `div`.

The placeholder text for empty sections should be done in a separate PR, I've added BB-721 for it.

MonkeyDo (approved):
Thank you !

PR #948, opened 2023-03-08
BB-563: Added banner to give test.bb.org a distinguishing feature

riyaku11:
### Problem
Test or beta site requires a distinguishing banner to differentiate it from Bookbrainz site

### Solution
Added a banner to distinguish test or beta site

### DIFF
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -150,17 +150,17 @@ app.use((req, res, next) => {
 	res.locals.repositoryUrl = repositoryUrl;
 	res.locals.alerts = [];
 	req.signUpDisabled = false;
-	const msg = '';
 	if (process.env.DEPLOY_ENV === 'test' || process.env.DEPLOY_ENV === 'beta') {
+		const msg = '';
 		if (process.env.DEPLOY_ENV === 'beta') {
-		msg = 'You are on the beta website, which uses the main database but with a newer version of the code to test new features.'
-	}
-	else {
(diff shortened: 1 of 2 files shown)


src/server/app.js
```diff
@@ -151,6 +151,13 @@ app.use((req, res, next) => {
 	res.locals.alerts = [];
 	req.signUpDisabled = false;
 
+	if (process.env.DEPLOY_ENV === 'test' || process.env.DEPLOY_ENV === 'beta') {
+		res.locals.alerts.push({
+			level: 'danger',
```
MonkeyDo, line 156: I think this would be a bit less harsh than a big red alert :D
```suggestion
			level: 'info',
```

src/server/app.js
```diff
@@ -151,6 +151,13 @@ app.use((req, res, next) => {
...
+			message: `You are on a ${process.env.DEPLOY_ENV} site`
```
MonkeyDo, line 157: Here it would be useful to spell out the difference between beta and test. something like:

> You are on the beta website, which uses the main database but with a newer version of the code to test new features.

and

> You are on the test website; all changes made here are not synced with the main database and will be overwritten periodically. Bugs and issues with the interface are also more likely.

MonkeyDo (changes requested):
Thanks for looking into this. A couple of suggestions below:

MonkeyDo (approved):

MonkeyDo:
Thank you @riyaku11  !

PR #949, opened 2023-03-09
BB-674: Rewrite revision file using async/await syntax

RohanSasne:
### Problem

[BB-673](https://tickets.metabrainz.org/browse/BB-674):
Across the codebase we have asynchronous code written using promises, which is hard to read and to maintain and error-prone.
### Solution

### Areas of Impact

The changes were reflected in `src/server/routes/revision.js`

### DIFF
--- a/src/server/routes/revision.js
+++ b/src/server/routes/revision.js
@@ -176,19 +176,19 @@ function formatEditionGroupChange(change) {
 	return [];
 }
 
-function diffRevisionsWithParents(orm, entityRevisions, entityType) {
+async function diffRevisionsWithParents(orm, entityRevisions, entityType) {
 	// entityRevisions - collection of *entityType*_revisions matching id
-	return Promise.all(entityRevisions.map(
+	const promises = entityRevisions.map(
 		async (revision) => {
 			const dataId = revision.get('dataId');
 			const revisionEntity = revision.related('entity');
 			const entityBBID = revisionEntity.get('bbid');
 			const entity = await orm.func.entity.getEntity(orm, entityType, entityBBID);
 			const isEntityDeleted = !entity.dataId;
-			return revision.parent()
-				.then(
-					(parent) => {
-						let isNew = false;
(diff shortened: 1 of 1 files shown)


src/server/routes/revision.js
```diff
@@ -178,52 +178,53 @@ function formatEditionGroupChange(change) {
...
+					revision
+				});
+			}
+			// If calling .parent() is rejected (no parent rev), we still want to go ahead without the parent
+			catch {
+				return makePromiseFromObject({
+					changes: revision.diff(null),
+					entity: revisionEntity,
+					entityAlias: dataId ?
+						await revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias', 'aliasSet.aliases']}) :
```
MonkeyDo, line 215: Same as above

src/server/routes/revision.js
```diff
@@ -178,52 +178,53 @@ function formatEditionGroupChange(change) {
...
+				let isNew = false;
+				const isDeletion = !dataId;
+				if (!parent) {
+					isNew = Boolean(dataId);
+				}
+				return makePromiseFromObject({
+					changes: revision.diff(parent),
+					entity: revisionEntity,
+					entityAlias: dataId ?
+						await revision.related('data').fetch({require: false, withRelated: ['aliasSet.defaultAlias', 'aliasSet.aliases']}) :
```
MonkeyDo, line 199: Here I believe we don't want to `await` the promise, as that is what `makePromiseFromObject` is used for.
It returns a Promise resolving to an object with each value resolved (if the value was a promise).

MonkeyDo (commented):
Thank you for continuing on this train ! 🚂

RohanSasne:
Thought over it and yes you're right, it was a error on my side it shouldn't await, fixed it :) thanks for pointing out !

MonkeyDo (approved):
Thank you !

PR #950, opened 2023-03-09
BB-725: refactor achievement component

meziyum:
### Problem [BB-725](https://tickets.metabrainz.org/browse/BB-725)

1. Simplicity and readability
2. Performance
3. Easier to test

### Solution
I have refactored the Achievement component from a class component to a functional component.  I have also optimized the image rendering logic by using a ternary operator instead of an if-else statement. The refactored component has improved code readability and maintainability.

### Areas of Impact
User Profile and Achievement Pages

image
image


src/client/components/forms/parts/achievement.js
```diff
@@ -56,64 +56,54 @@ const maxAchievementProgress = {
...
-				</Container>
-			</Card>
-		);
-	}
+function Achievement(props) {
+	const {achievement, counter, unlocked} = props;
+	const imgElement = unlocked ? (
+		<DragAndDropImage
+			achievementId={achievement.id}
+			achievementName={achievement.name}
```
MonkeyDo, line 65: For clarity and because we use these props everywhere in this component, what about destructuring the achievement id, name, etc. like you do for the props?

MonkeyDo (commented):
Looks sensible, thanks for this improvement!

src/client/components/forms/parts/achievement.js
meziyum: I did think about it initially but decide to wait for your review to get idea of how you want it to be done. I will immediately work on it.

MonkeyDo (approved):
Great job :)

I'll squash the commits as I merge this PR, because the commit history is a bit messy, but no worries about that.

PR #951, opened 2023-03-09
chore: update host and user of the test DB

kellnerd:
Changed in the config files but not in the test DB script, see 814d07b3161908e82127b9ef53c04751dd20c77f.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,7 @@ jobs:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
           POSTGRES_DB: bookbrainz_test
+          POSTGRES_HOST: localhost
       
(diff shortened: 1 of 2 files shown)


MonkeyDo (approved):
Yes, I arrived at the same conclusion, but ran out of time to test. Thanks !

kellnerd:
Only the CI is apparently not happy with the changed host...

kellnerd:
I have zero experience with GitHub actions, but since your attempt also failed, I would try to change the host in line 63 instead?

MonkeyDo:
> I have zero experience with GitHub actions, but since your attempt also failed, I would try to change the host in line 63 instead?
> 
> _Edit_: Yep, that did the trick.

Oh god. That's what I get for multi-tasking :p
Thanks for the assist!

PR #957, opened 2023-03-13
BB-709: Add Author Credits column in publisher editions listing

the-good-boy:
### Problem
It would be quite informative to see the author(s) of an edition in the list of editions for a given publisher. This PR addresses [BB-709](https://tickets.metabrainz.org/browse/BB-709)

### Solution
Added a column to display the author credits using the already existing AuthorCreditDisplay component.

### DIFF
--- a/src/client/components/pages/entities/publisher.js
+++ b/src/client/components/pages/entities/publisher.js
@@ -101,7 +101,7 @@ function PublisherDisplayPage({entity, identifierTypes, user}) {
 			<EntityAnnotation entity={entity}/>
 			{!entity.deleted &&
 			<React.Fragment>
-				<EditionTable editions={entity.editions} entity={entity} showAuthorCreditsColumn/>
+				<EditionTable showAuthorCreditsColumn editions={entity.editions} entity={entity}/>
 				<EntityLinks
 					entity={entity}
 					identifierTypes={identifierTypes}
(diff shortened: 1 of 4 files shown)


src/client/components/pages/entities/publisher.js
```diff
@@ -101,7 +101,7 @@ function PublisherDisplayPage({entity, identifierTypes, user}) {
 			<EntityAnnotation entity={entity}/>
 			{!entity.deleted &&
 			<React.Fragment>
-				<EditionTable editions={entity.editions} entity={entity}/>
+				<EditionTable editions={entity.editions} entity={entity} showAuthorCreditsColumn={entity.showAuthorCreditsColumn}/>
```
MonkeyDo, line 104: I don't think we need to set showAuthorCreditsColumn in the entity at all on the server side (which means we can remove [this](https://github.com/metabrainz/bookbrainz-site/pull/957/files#diff-6e372470c7eb4abe59f52ad00d5de4cb951ea90de60d72fd8b16fdf9e8d2b852R211)); instead we can simply use this boolean prop directly right here.
```suggestion
				<EditionTable editions={entity.editions} entity={entity} showAuthorCreditsColumn/>
```

MonkeyDo (changes requested):
Thanks for this improvement !

src/client/components/pages/entities/publisher.js
the-good-boy: You're right! Thanks for the suggestion :)

PR #962, opened 2023-03-14
BB-718 - CritiqueBrainz auth endpoint is called server-side.

Legend101Zz:
### Problem
<!-- What are you trying to solve?
(BB-718) CritiqueBrainz auth endpoint is called server-side.

### Solution
Initially, the CritiqueBrainz auth endpoint ( called by the `getAccessToken`) was server side rendered as it was called using `componentDidMount` method .  The method was changed to `componentDidUpdate` and `showModal` key was used to initialise it's trigger . `hasComponentUpdated` localStorage item was created to only hit the `getAccessToken` function once .

### Areas of Impact

The changes were reflected in `src/client/components/pages/entities/cbReviewModal.tsx`


MonkeyDo (changes requested):
Hi! It looks like you IDE is set up to automatically format on save, but not set up to follow our particular set of rules (maybe it uses prettier by default?).
Consequently, you'll see in the [PR modified files](https://github.com/metabrainz/bookbrainz-site/pull/962/files) that there are hundreds of changes unrelated to what you are trying to do with this PR: quotation marks, formatting, surprisingly even some transformation of async/await syntax to promise.then(…), etc.

You'll also see that this creates a lot of ESLint errors.


This also makes it very very hard for me to review this PR, as I don't know what I should be looking at. I can't currently see what change you have made to resolve the ticket.

I'm also unclear why the src/client/components/pages/contribute.js file is committed in here, considering the ticket is specifically about the CBReviewModal component.

Please review those issues an either force-push to this PR or open a new PR if that's easier.

Legend101Zz:
> Hi! It looks like you IDE is set up to automatically format on save, but not set up to follow our particular set of rules (maybe it uses prettier by default?). Consequently, you'll see in the [PR modified files](https://github.com/metabrainz/bookbrainz-site/pull/962/files) that there are hundreds of changes unrelated to what you are trying to do with this PR: quotation marks, formatting, surprisingly even some transformation of async/await syntax to promise.then(…), etc.
> 
> You'll also see that this creates a lot of ESLint errors.
> 
> This also makes it very very hard for me to review this PR, as I don't know what I should be looking at. I can't currently see what change you have made to resolve the ticket.
> 
> I'm also unclear why the src/client/components/pages/contribute.js file is committed in here, considering the ticket is specifically about the CBReviewModal component.
> 
> Please review those issues an either force-push to this PR or open a new PR if that's easier.

@MonkeyDo I'll do the necessary changes, apologies for that , would keep that in mind for future , cheers !!!

MonkeyDo:
I'm closing the PR due to inactivity, and considering the hundreds of unrelated formatting changes.
More pointedly, the actual implementation does not seem like a good way to fix the issue.

PR #964, opened 2023-03-14
BB-690: added unset functionality for dropdowns

stoichy:
### Problem
 BB-690

### Solution

Reset functionality for _Select_ elements  in the +Add sections, like "Type", "Format" or "Status"

### Areas of Impact

Modified 6 files that handle 6 different pages in the +Add category. Added an attribute for _Select_ elements that enables the possibility to unset an option.

### DIFF
--- a/src/client/components/forms/registration-details.js
+++ b/src/client/components/forms/registration-details.js
@@ -146,6 +146,7 @@ class RegistrationForm extends React.Component {
 										defaultValue={this.props.genders.filter((option) => option.id === initialGender)}
 										getOptionLabel={this.getOptionLabel}
 										instanceId="gender"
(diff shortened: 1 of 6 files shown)


stoichy:
PTAL. 
Found one more case in the registration form, where gender is optional. In some other places (ex. Identifier editor), there was no need for an unset option, because the fields were mandatory, not optional.

meziyum (approved):

meziyum (approved):
I personally went through all 20 occurrences of files which import react-select and this PR adds unset functionality to all the optional dropdowns wherever needed. Great job 😄

meziyum (approved):

PR #966, opened 2023-03-14
BB-726: refactor drag-and-drop component

meziyum:
### Problem [BB-726](https://tickets.metabrainz.org/browse/BB-726)

1. Simplicity and readability
2. Performance
3. Easier to test

### Solution
The refactored code replaces the class component with a function component. Specifically, it uses the useState and useCallback hooks to manage state and event handlers respectively. The handleClick, handleDragOver, and handleDrop methods in the class component are now replaced with callback functions that are defined using the useCallback hook.

The refactored code also removes the constructor method and instead initializes the state using the useState hook. The addChild method is now replaced with a new function defined within the component scope, and the state is updated using the setAchievement function returned by the useState hook.

Overall, the refactored code simplifies the class component by using React hooks, reducing the amount of boilerplate code, and making the code easier to read and understand.

### Areas of Impact
Achievement Pages


MonkeyDo (commented):
Thank you for this improvement !

While we are here refactoring, could you please make the file type safe with typescript?
You will need to change the file extension to `.tsx` and make the code comply with TypeScript types. I'm happy to help if you are unfamiliar with TS.

PR #967, opened 2023-03-15
BB-727: refactor drag-and-drop-image component

meziyum:
### Problem [BB-727](https://tickets.metabrainz.org/browse/BB-727)

1. Simplicity and readability
2. Performance
3. Easier to test

### Solution
The refactored code replaces the class component with a functional component that uses the useCallback hook. The useCallback hook memoizes the handleDragStart function, which ensures that it does not get redefined on every render unless any of its dependencies change.

The DragAndDropImage function takes in four props, achievementId, achievementName, height, and src. It then defines the handleDragStart function that creates a data object from the props and sets it as the text data in the DragAndDrop event.

The function returns an image element that has the draggable attribute set to true, the height set to the value of the height prop, the src set to the value of the src prop, and the onDragStart event listener set to the memoized handleDragStart function.

Finally, the function sets the displayName and propTypes properties on the DragAndDropImage function for easier debugging and type checking, respectively.

Overall, the refactored code is simpler, more concise, and more performant than the original code.

### Areas of Impact
Achievement Pages


MonkeyDo (commented):
Thanks also for this one !
Similarly to #966 please also make this file a typescript file and make it strong ! 💪 

Also see the ESLint warning at the beginning of the file.

MonkeyDo (approved):
Clean, thank you ! :rocket:

PR #979, opened 2023-03-28
BB-730: Rewrite test files using async/await syntax

faraz16iqbal:
Problem
[BB-673](https://tickets.metabrainz.org/browse/BB-673):
Across the codebase, we have asynchronous code written using promises, which is hard to read and to maintain and error-prone.

Solution

Areas of Impact
The changes were reflected in test/common.js

### DIFF
--- a/data/test-data.js
+++ b/data/test-data.js
@@ -475,7 +475,7 @@ export function createSprinter() {
 		);
 }
 
-export function sprinterHelper(numRevisions) {
+export async function sprinterHelper(numRevisions) {
 	const promiseList = [];
 	for (let i = 0; i < numRevisions; i++) {
 		promiseList.push(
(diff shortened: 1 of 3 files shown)


MonkeyDo:
Sorry it took so long to answer @faraz16iqbal !
Yes, the modifications look great.
Should I merge the PR now or do you want to modify other files in the same PR?

faraz16iqbal:
@MonkeyDo thank you for the review! I think I'll use this PR to convert all the files in one go.

faraz16iqbal:
@MonkeyDo I mentioned there were some files that used the following syntax: 
`if (days === funRunnerDays) {
 editPromise = Promise.resolve(threshold);
			}`

Do these also need to be changed?

MonkeyDo:
> 

Looking at that file with that code I don't think it's necessary to rewrite those as an async function.
ESLint will tell you an async function needs an await expression, and we don't have anything to await in that in this particular case.
I think you can leave those as-is.

faraz16iqbal:
@MonkeyDo I have converted all tests files to follow the async-await syntax. Please review whenever you can!

MonkeyDo (approved):
And thank you for this one as well ! :)

PR #987, opened 2023-05-08
BB-714: Fix search results page on Safari

faraz16iqbal:
[BB-714](https://tickets.metabrainz.org/browse/BB-714)
### Problem
The width of tables on the Webkit-based browser are broken.

The width of the columns is defined using Bootstrap CSS classes, which is not working on Webkit.

### Solution
Setting the column sizes using the width attribute in percentages solves the issue.

### Areas of Impact
All tables rendered on the Bookbrainz website that use the `col-md` class name to specify the width.

### DIFF
--- a/src/client/components/pages/entities/series-table.js
+++ b/src/client/components/pages/entities/series-table.js
@@ -79,7 +79,7 @@ function SeriesTable({series, showAddedAtColumn, showCheckboxes, selectedEntitie
 				<Table striped>
 					<thead>
 						<tr>
-							<th style={{width: '40%'}}>Name</th>
+							<th width="40%">Name</th>
 							<th>Series Type</th>
 							<th>Ordering Type</th>
 							{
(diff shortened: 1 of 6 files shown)


faraz16iqbal:
@MonkeyDo there were some other tables that were using the `col-md` class name to specify the width. I have updated all of them under this PR

faraz16iqbal:
> Thanks, this looks a whole lot better already!
> 
> I did a bit of testing after putting the PR up on test.bookbrainz, and found the profile page looking like this:
> 
> image
> Not sure what's happening there but I'll leave it with you to investigate. If you need help testing with Safari I can help out (otherwise there's also BrowserStack that I sometimes use for browser testing)

I think its because of the _dt_ and _dh_ tags not supporting the `width` attribute. I think reverting this will solve the problem

faraz16iqbal:
@MonkeyDo this should be good to go now!

MonkeyDo (approved):
Thank you !

PR #990, opened 2023-05-22
BB-741: Add https:// prefix without using a lookbehind expression

kellnerd:
### Problem

[BB-741](https://tickets.metabrainz.org/browse/BB-741): Regular expression error in Safari

RegExp lookbehind expressions were unsupported by Safari until version 16.4 from March 2023.

### Solution

Add missing https:// prefixes to plain text URLs without using a lookbehind expression.
This should have been the only occurrence of such an expression in our codebase.

### Areas of Impact

All pages which display annotations or revision notes, but apparently it broke the whole site's client JS in Safari (which I couldn't verify without having an installation of Safari).


MonkeyDo (approved):
Tested on test.bookbrainz.org, solves the Safari issue and does improve the way we display the link.
Nice work, and so quick! Thank you!

PR #992, opened 2023-05-23
fix(core): handle undefined Accept-Language header

kellnerd:
### Problem

The HTTP Accept-Language header can also be undefined which causes a type error.

### Solution

Treat this case the same way as an empty (`''`) or wildcard (`'*'`) header which both return an empty array of language codes.

### Areas of Impact

Server routes which are loading Wikipedia extracts.


MonkeyDo (approved):
Thanks for the lightning quick response :)

PR #996, opened 2023-06-13
feat(Admin System): Admin Panel Search Page and Edit Privileges

the-good-boy:
### Overview:
This PR adds the functionality which allows an Administrator to search for users and then change their privileges.

### Components Added:

- **Privileges Dropdown:** A privileges dropdown has been added in the NavBar, which will eventually contain options corresponding to the special privileges that a user has. Currently only the Admin Panel can be accessed using this dropdown. Ideally, this Admin Panel can be accessed only by users who have the Administrator priv. We'll later use middlewares to fix these.

- **Admin Panel Search Page:** As the name implies, it allows us to search for users and makes use of two sub-components named: `AdminPanelSearchField` and `AdminPanelSearchResults`. The functionality of this is largely inspired from the [Search Page](https://bookbrainz.org/search?q=&size=20&from=0), which we already have.

- **Privs Edit Modal:** A modal which allows us to change privileges.
- **Privilege Badges:** This component displays the titles of the privileges in the form of badges.

### Icons:
I have taken some liberty and added some icons to differentiate users depending on the sets of privileges they have. They are used in the `PrivEditModal` and `AdminPanelSearchResults` components and brief outline of what they represent is as follows:
- **[White shield:](https://github.com/metabrainz/bookbrainz-site/pull/996/files#diff-2841b9ee1c3e343f5d9461bf49f59123cb363f82ef91331403e7e2e605c4dbae)** This is used to indicate a user with no privileges.
- **[Grey shield:](https://github.com/metabrainz/bookbrainz-site/pull/996/files#diff-f53166921b3b32f37af522f177776488af279d8f1121f2147f64d15e9194d2c6)** This is used to indicate a user with the Entity-Editor privilege (i.e., a normal editor without any special privilege)
- **[Orange shield:](https://github.com/metabrainz/bookbrainz-site/pull/996/files#diff-3c622667fa4bf8cacf9bfe88c2c1cd50b270627db30a5c99c88ec8e09bf7578d)** This is used to indicate a user with some special privileges, but not the admin priv.
- **[Orange shield with a tick:](https://github.com/metabrainz/bookbrainz-site/pull/996/files#diff-39cc26903a69ea6ed494e3e3b2251b30f86757933681c021b4eac7d1e290363c)** This is used to indicate that the user has the Admin priv.


src/common/helpers/privileges-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+	const variants = {
+		'Entity Editor': 'secondary',
+		'Identifier Type Editor': 'warning',
+		'Relationship Type Editor': 'info',
+		'Reindex Search Engine': 'success',
+		Administrator: 'danger'
+	};
+
+	return variants[title];
+}
```
anshg1214, line 65: You can create a `PRIVILEGE` object with every detail about the privilege type like `title`, `variant` mapped to it's bit. Whenever you'd need to add a new privilege, or modify some information, you can directly update the object content instead of modifying the functions.

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -0,0 +1,149 @@
...
+					'Content-Type': 'application/json; charset=utf-8'
+				},
+				method: 'POST'
+			});
+			if (!response.ok) {
+				const {error} = await response.json();
+				throw new Error(error ?? response.statusText);
+			}
+			window.location.reload();
+		}
```
anshg1214, line 68: It'll be better if you can avoid reloading the window, by getting the updated values from the API and updating the states.

src/common/helpers/privileges-utils.ts
anshg1214, line 65: You can create a `PRIVILEGE` object with every detail about the privilege type like `title`, `variant` mapped to it's bit. Whenever you'd need to add a new privilege, or modify some information, you can directly update the object content instead of modifying the functions.

src/server/routes/editor.js
```diff
@@ -174,6 +174,26 @@ router.post('/edit/handler', auth.isAuthenticatedForHandler, (req, res) => {
...
+				'No change to Privileges', req
+			));
+		}
+		const modifiedEditor = await editor.save({privs: req.body.newPrivs});
+		return res.status(200).send();
+	}
+	catch (err) {
+		return next(err);
+	}
+});
```
anshg1214, line 195: You should add a check, if the user who has sent the request has the privilege to perform this operaiton.

anshg1214 (changes requested):
I also found a couple of linting errors, try fixing those too.

src/server/routes/editor.js
the-good-boy: We'll eventually have to add middlewares for privilege-based access on almost all the routes. I think that will cover this.

src/client/components/pages/parts/privs-edit-modal.js
the-good-boy: Agreed. Updating the states definitely seems to be a better way. I have made some changes. I am not getting updated values from the API because each 'Save Privilege' action makes only one change to the state: it changes the `privs` variable of the `targetUser`, so I have directly updated the value on the state object with the new object. What do you think?

anshg1214: This solution works, but you should fetch the entire data from the API again so that in case any other use has changed the privileges or a new user gets added, it gets reflected on your end as well. So this function `updatePrivsOnResultsList` should pull a new copy of the data.

src/common/helpers/privileges-utils.ts
anshg1214: Also what do you think about this? @the-good-boy

src/client/components/pages/parts/privs-edit-modal.js
the-good-boy: Hmm, good point. I have made the changes. Now, I am using the `triggerSearch` function inside the Pager component to refetch the data using the same query whenever we are updating the privileges. For achieving this task, I am passing a trigger as a prop which helps us run this `triggerSearch` function(since this lies in a child Pager component) from the parent AdminPanelSearch component. What do you think of this approach?

src/common/helpers/privileges-utils.ts
the-good-boy: I have made some relevant changes for this.

the-good-boy:
> 



> I also found a couple of linting errors, try fixing those too.

About this, can you point me to a line where a linting error is arising? I have tried running linters on my local machine and all seems good.

src/common/helpers/privileges-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+		title: 'Identifier Type Editor',
+		badgeVariant: 'warning'
+	},
+	2: {
+		title: 'Relationship Type Editor',
+		badgeVariant: 'info'
+	},
+	3: {
+		title: 'Reindex Search Engine',
+		badgeVariant: 'succes'
```
anshg1214, line 17:
```suggestion
		badgeVariant: 'success'
```
A typo here

anshg1214 (changes requested):
Just a small change. Otherwise good to go

anshg1214:
> > 
> 
> > I also found a couple of linting errors, try fixing those too.
> 
> About this, can you point me to a line where a linting error is arising? I have tried running linters on my local machine and all seems good.

It seems that they got fixed. Also there are a few linting warnings, if you can fix them as well.

the-good-boy:
> 



> > > 
> > 
> > 
> > > I also found a couple of linting errors, try fixing those too.
> > 
> > 
> > About this, can you point me to a line where a linting error is arising? I have tried running linters on my local machine and all seems good.
> 
> It seems that they got fixed. Also there are a few linting warnings, if you can fix them as well.

I was running `npm run lint-errors`. Hence, I didn't see the warnings. Thanks for pointing this out. I've now fixed the warnings.

anshg1214 (approved):
🎉

src/client/components/pages/admin-panel-search.tsx
```diff
@@ -0,0 +1,186 @@
...
+						<PagerElement
+							from={this.props.from}
+							nextEnabled={this.props.nextEnabled}
+							paginationUrl={this.paginationUrl}
+							querySearchParams={querySearchParams}
+							results={results}
+							searchParamsChangeCallback={this.searchParamsChangeCallback}
+							searchResultsCallback={this.searchResultsCallback}
+							size={this.props.resultsPerPage}
+							updateResultsTrigger={updateResultsTrigger}
```
MonkeyDo, line 168: I have another mechanism to propose for this, that looks less like a workaround.
Instead of using this `updateResultsTrigger` to trigger a method in `componentDidUpdate`, you can save a reference to the Pager component in this AdminPanelSearchPage class (using the ref prop like so `ref={this.pagerRef}`), and then in `updateResultsOnPrivsChange` we can use that ref to call a public method of the Pager class directly: `this.pagerRef?.current?.triggerSearch()` <- this method in Pager default to using the current state values from `from` and `size`.

Note the syntax above uses `?.` too make sure there is a stored reference before calling the method, otherwise you can get some nasty "method XXX does not exist on undefined" errors.

You'll need instantiate that ref variable in the AdminPanelSearchPage constructor (`this.pagerRef = React.createRef();`)

src/client/components/pages/parts/admin-panel-search-results.tsx
```diff
@@ -0,0 +1,160 @@
...
+								adminId={this.props.user.id}
+								handleCloseModal={this.onCloseModal}
+								show={this.state.showModal}
+								targetUser={this.state.selectedUser}
+								updateResultsOnPrivsChange={this.props.updateResultsOnPrivsChange}
+							/>
+						</div>
+					)
+				}
+				<h3 style={{color: '#754e37'}}>
```
MonkeyDo, line 136: CSS classes instead of inlined styles, please :)
It's easier to reason about the CSS and keep it all synchronized when it is all in one place instead of scattered across the codebase.

src/client/components/pages/parts/admin-panel-search-results.tsx
```diff
@@ -0,0 +1,160 @@
...
+					Search Results
+				</h3>
+				<hr className="thin"/>
+				<Table
+					responsive
+					className={tableCssClasses}
+				>
+					<thead>
+						<tr>
+							<th className="col-md-5">Name</th>
```
MonkeyDo, line 146: Using bootstrap class names here does not work on Safari (we just fixed that same issue across the codebase in #987, it's fresh on my mind :) ).

You can use the `width` property (native to the `<th>` element) to the same effect. The value should be `width="42%"` which is what col-md-5 does, although I have found that the following values work well:

image

src/client/components/pages/parts/pager.js
```diff
@@ -60,6 +60,10 @@ class PagerElement extends React.Component {
 			// eslint-disable-next-line react/no-did-update-set-state
 			this.setState({from: 0}, this.triggerSearch);
 		}
+
+		if (prevProps.updateResultsTrigger !== this.props.updateResultsTrigger) {
+			this.triggerSearch(this.state.from, this.state.size);
+		}
```
MonkeyDo, line 66: With the `ref` method described above, this won't be needed.

src/client/components/pages/parts/pager.js
```diff
@@ -200,7 +204,8 @@ PagerElement.propTypes = {
 	results: PropTypes.array,
 	searchParamsChangeCallback: PropTypes.func,
 	searchResultsCallback: PropTypes.func.isRequired,
-	size: PropTypes.number
+	size: PropTypes.number,
+	updateResultsTrigger: PropTypes.number.isRequired
```
MonkeyDo, line 208: Same comment as above.

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -0,0 +1,153 @@
...
+			}
+			this.props.updateResultsOnPrivsChange();
+			this.props.handleCloseModal();
+		}
+		catch (err) {
+			throw new Error(err);
+		}
+	}
+
+	/* eslint-disable no-bitwise */
```
MonkeyDo, line 77: You know, there's rarely a good use for disabling linting rules, but if I ever need a good example this will be the perfect one ! :D

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -0,0 +1,153 @@
...
+		return (
+			<Modal
+				scrollable
+				show={this.props.show}
+				onHide={this.props.handleCloseModal}
+			>
+				<Modal.Header closeButton>
+					<Modal.Title>
+						<img height="20" src={getPrivilegeShieldIcon(this.state.privs)}/> {' '}
+						<a href={link}>
```
MonkeyDo, line 118: I wonder if this should be a link to the  user's profile. It might be clearer to have this be just text, and have a separate "visit profile" link in the modal body (perhaps opening in a new tab).
What do you think?

src/client/containers/layout.js
```diff
@@ -123,12 +123,33 @@ class Layout extends React.Component {
 			</span>
 		);
 
+		const privilegesDropdownTitle = (
+			<span>
+				<FontAwesomeIcon icon={faShieldHalved}/>
+				{' Privileges'}
```
MonkeyDo, line 129: This doesn't need to be quoted and in a JSX block, it can just be the text content

src/client/containers/layout.js
```diff
@@ -123,12 +123,33 @@ class Layout extends React.Component {
...
+					id="privs-dropdown"
+					open={this.state.menuOpen}
+					title={privilegesDropdownTitle}
+					onMouseDown={this.handleMouseDown}
+					onSelect={this.handleDropdownClick}
+					onToggle={this.handleDropdownToggle}
+				>
+					<NavDropdown.Item href="/admin-panel">
+						<FontAwesomeIcon icon={faUserGear}/>
+						{' Admin Panel'}
```
MonkeyDo, line 150: Same as above, this can be just the text

MonkeyDo (commented):
Great work so far ! You've got a pretty clean interface, working smoothly, and the code is nice and readable. Well done !
I've deployed the PR to test.bookbrainz.org for testing.

I'm very late to the party, but I have some code suggestions below. None of them are super pressing, but I'd love to see them implemented in a follow-up PR.

Looking forward to seeing the continuation !

src/client/components/pages/parts/admin-panel-search-results.tsx
```diff
@@ -0,0 +1,160 @@
...
+					<td>
+						<PrivilegeBadges privs={result.privs}/>
+					</td>
+					<td>
+						<Button
+							variant="link"
+							onClick={() => this.openPrivsEditModal(result)}
+						>
+							<img height="18" src={getPrivilegeShieldIcon(result.privs)}/> {' '}
+							<FontAwesomeIcon icon={faPencilAlt}/>
```
MonkeyDo, line 110: I wonder about this; it seems to me that the shield icon only repeats the information we already have in the privileges column.

I think we only need an "Edit" button for now (in the future maybe "Block"/"Delete" buttons as well), for which I'd rather have a text button that the very small icon which is harder to click on.

What do you think?
Here it is with text, and some button classes (btn-sm for size and btn-success for style)
image

the-good-boy: The shield icons were never really for any kind of additional information. They were just an aesthetic choice, one which I dont mind rescinding. One other reason I clubbed them with the pencil icon was so I can increase the hitbox of the click.

I did give some thoughts on how the edit button should look like and here's why I ditched this simple Edit button. The main reason is because I think these buttons, when iterated in a table look really out of place to me, mostly because these types of buttons have mostly been used throughout the site as singular action initiators(i made that up), basically you find them mostly in isolation (as a 'Submit' button in a page, or a 'Create a collection' button, etc). I find that they look a bit odd when iterated in a table:
Screenshot 2023-06-27 at 16-44-39 Search Results – BookBrainz

This also reminds me that I've been in a similar dilemma before, in this old [PR](https://github.com/metabrainz/bookbrainz-site/pull/830) where I used a question mark as a clickable link. I do agree that it does not work very good as a call to action, but now you know why I went for it instead of a button.

I do realise that this can all be a very subjective discussion, and it is quite possible that I may be reading too much into this. I am okay with changing it to an 'Edit' button as suggested for now, but I think we should go through some more ideas on what kind of buttons we'll use in a table.

MonkeyDo: > The shield icons were never really for any kind of additional information.
I think they look neat, but in this case they should be used in another place if they repeat info from the previous column. Maybe they would be a good indicator in front of user names on the website? Possibly only showing admins and privileged users, and not showing the empty shield icon in front of every user's name.

For this table, if you were to add it, I would add it in front of the user name instead of the user icon (which currently doesn't add any info). Might be helpful when quickly reading through the table.


I get your thought process; the repeated buttons are perhaps not ideal, but IMO they are more *usable* that just the icon (as you mention the hitbox is small) even if they look better. The issue will compound once we have more actions.
I think it'll end up looking fine with multiple buttons, each having a different color will help prevent misclicks (when compared to standalone icons).
If you want to jazz them up, I have two suggestion:
1. we have outline style buttons which could reduce the repeating color block issue (i.e. `btn-outline-success` class etc. for -info, -danger, -warning etc.)
2. you can add the icon in the button too

src/client/components/pages/admin-panel-search.tsx
the-good-boy: Thanks for this suggestion. I have made a [PR](https://github.com/metabrainz/bookbrainz-site/pull/998) with these changes.

src/client/components/pages/parts/admin-panel-search-results.tsx
the-good-boy: Indeed. I have created a class named `search-results-heading` in this [PR](https://github.com/metabrainz/bookbrainz-site/pull/998). Since this `AdminPanelSearchResults` component is largely inspired from the `SearchResults` component, I have made this change in that file as well. It was a small enough change to not warrant another PR, and so I took the liberty of including this change too.

src/client/components/pages/parts/privs-edit-modal.js
the-good-boy: For now, I have removed the link to the user's profile in this follow up [PR](https://github.com/metabrainz/bookbrainz-site/pull/998). I will, however, be sure to take care of this "Visit Profile" functionality in the next iteration of the PrivsModal.

PR #997, opened 2023-06-22
fix(display): prevent two consecutive spaces in JSX

kellnerd:
3feb7142977b652015456838ae6d7178e7068330 inserted a non-breaking space at the end of the line which is not automatically stripped, but treated as text. Because the next line also starts with text, an additional space is inserted to separate "words" (which was unwanted in this case).


MonkeyDo (approved):
Thanks :)

PR #998, opened 2023-06-27
feat(Admin System): Add edit buttons instead of icons and use refs instead of updateResultsTrigger

the-good-boy:
This PR is a continuation of previous [PR](https://github.com/metabrainz/bookbrainz-site/pull/996).


src/client/components/pages/admin-panel-search.tsx
```diff
@@ -70,15 +69,17 @@ class AdminPanelSearchPage extends React.Component<Props, State> {
...
+			results: props.initialResults
 		};
 
 		this.paginationUrl = './admin-panel/search';
+		this.pagerRef = React.createRef();
 	}
 
 	paginationUrl: string;
 
+	pagerRef: any;
```
MonkeyDo, line 81: The refs can be a bit tricky to access (ref.current can potentially be undefined) so it's best to type these strongly to avoid any runtime errors. This should do it:
```
    this.pagerRef = React.createRef<PagerElement>();
    [...]
	pagerRef: React.RefObject<PagerElement>;
```

src/client/components/pages/admin-panel-search.tsx
```diff
@@ -125,10 +126,7 @@ class AdminPanelSearchPage extends React.Component<Props, State> {
 	 * triggerSearch function
 	 */
 	updateResultsOnPrivsChange = () => {
-		const {updateResultsTrigger: trigger} = this.state;
-		this.setState({
-			updateResultsTrigger: 1 - trigger
-		});
+		this.pagerRef?.current?.triggerSearch();
```
MonkeyDo, line 129: Much cleaner IMO 👍

src/client/components/pages/parts/admin-panel-search-results.tsx
```diff
@@ -143,9 +140,9 @@ class AdminPanelSearchResults extends React.Component<AdminPanelSearchResultsPro
 				>
 					<thead>
 						<tr>
-							<th className="col-md-5">Name</th>
-							<th className="col-md-5">Privileges</th>
-							<th className="col-md-5"/>
+							<th width="42%">Name</th>
+							<th width="42%">Privileges</th>
+							<th width="42%"/>
```
MonkeyDo, line 145: I haven't checked the final output, but using my quick maths power I deduct we have 126% total.
I guess the first two columns take 42% each and the last one gets what's left (16%)?
In which case you can probably just leave out the width for the last column header.

MonkeyDo (commented):
Thanks for the follow up PR !

src/client/components/pages/parts/admin-panel-search-results.tsx
the-good-boy: haha! my bad. I thought the 42% means that it will take up a maximum of 42% width of the total container depending upon the contents.

MonkeyDo: Do check the result, I haven't run the code so this is just a hunch

the-good-boy: at first glance, it looked alright. But I think it is proportionally scaling down the width percentages depending upon the content in each column, if the sum is greater than 100. If the sum is 100, then it keeps the proportion constant.

src/client/components/pages/admin-panel-search.tsx
the-good-boy: Indeed

PR #999, opened 2023-07-03
feat(Admin System): Add middleware for securing routes according to privs

the-good-boy:
This PR adds a middleware to facilitate privilege-based access to certain routes.

### DIFF
--- a/src/server/helpers/auth.js
+++ b/src/server/helpers/auth.js
@@ -172,3 +172,24 @@ export function isAuthenticatedForCollectionView(req, res, next) {
 		'You do not have permission to view this collection', req
 	);
 }
+
+export function isAuthorized(flag) {
+	return async (req, res, next) => {
+		try {
+			const {Editor} = req.app.locals.orm;
+			const latestPrivs = await Editor.query({where: {id: req.user.id}})
+				.fetch({require: true})
+				.then(editor => editor.get('privs'));
+			/* eslint-disable no-bitwise */
+			if (latestPrivs & flag) {
+				return next();
+			}
+			throw new error.NotAuthorizedError(
+				'You do not have the privilege to access this route', req
+			);
+		}
+		catch (err) {
+			return next(err);
+		}
+	};
+}
--- a/src/server/helpers/auth.js
+++ b/src/server/helpers/auth.js
@@ -177,11 +177,10 @@ export function isAuthorized(flag) {
 	return async (req, res, next) => {
 		try {
 			const {Editor} = req.app.locals.orm;
-			const latestPrivs = await Editor.query({where: {id: req.user.id}})
-				.fetch({require: true})
-				.then(editor => editor.get('privs'));
+			const editor = await Editor.query({where: {id: req.user.id}})
+				.fetch({require: true});
 			/* eslint-disable no-bitwise */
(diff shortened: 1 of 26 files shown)


src/server/helpers/auth.js
```diff
@@ -172,3 +172,24 @@ export function isAuthenticatedForCollectionView(req, res, next) {
...
+		try {
+			const {Editor} = req.app.locals.orm;
+			const latestPrivs = await Editor.query({where: {id: req.user.id}})
+				.fetch({require: true})
+				.then(editor => editor.get('privs'));
+			/* eslint-disable no-bitwise */
+			if (latestPrivs & flag) {
+				return next();
+			}
+			throw new error.NotAuthorizedError(
```
MonkeyDo, line 187: Something to keep in mind is that this error, combined with our error handler down the line, will return an HTML page to the client (the error page).

This means that if you get a 403 when changing privs in the admin panel, you currently don't get any feedback that it failed.
The error handler there expects JSON, not an HTML page:
image

To replicate:
- make a second account
- give account2 admin privs
- log in with account2 and got to admin panel
- search for a user and open the privs edit modal
- with account1, remove admin privs from account2
- with account2, change privs and submit
- you'll see nothing happens apart from errors in the console


Obviously this is an edge case, but that might mean that we need to return different types of errors depending on whether we have a page route or a programmatic endpoint, or at least have a look at https://github.com/metabrainz/bookbrainz-site/blob/administration-system/src/client/components/pages/parts/privs-edit-modal.js#L66-L67 to ensure we can catch that error too

MonkeyDo (commented):
Looking really good @the-good-boy !
Everything working as expected. I tried hard to break it, but only found some minor issues with error handling.

I deployed the PR to test.BB for testing in situ, works a treat :) 
Nice and reactive with the fetching privs for each query !

src/server/helpers/auth.js
```diff
@@ -172,3 +172,24 @@ export function isAuthenticatedForCollectionView(req, res, next) {
...
 	);
 }
+
+export function isAuthorized(flag) {
+	return async (req, res, next) => {
+		try {
+			const {Editor} = req.app.locals.orm;
+			const latestPrivs = await Editor.query({where: {id: req.user.id}})
+				.fetch({require: true})
+				.then(editor => editor.get('privs'));
```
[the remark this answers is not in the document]
the-good-boy: I agree. Thanks!

src/server/helpers/auth.js
the-good-boy: Well caught! I have made some changes so that we don't expect a JSON in case of a 403. Do you think this solution is good enough?

MonkeyDo: Yes, that works

the-good-boy: Hmm... Definitely it will be better to have a different way of dealing with errors for a programmatic endpoint.

MonkeyDo: Like so most likely: https://github.com/metabrainz/bookbrainz-site/blob/39e80222ca84bc71f059c61c95d06d05d1de7f98/src/api/helpers/entityLoader.ts#L60

MonkeyDo (approved):
Good to go !

PR #1000, opened 2023-07-06
feat(Admin System): Improve type validation with TypeScript

the-good-boy:
This PR is a followup to the previous [PR](https://github.com/metabrainz/bookbrainz-site/pull/999), and it specifically addresses the points raised [here](https://github.com/metabrainz/bookbrainz-site/pull/999#discussion_r1252176292).

### DIFF
--- a/src/server/routes/adminPanel.tsx
+++ b/src/server/routes/adminPanel.tsx
@@ -30,16 +30,17 @@ import {PrivilegeType} from '../../common/helpers/privileges-utils';
 import React from 'react';
 import ReactDOMServer from 'react-dom/server';
 import express from 'express';
+import {parseQuery} from '../helpers/utils';
 import target from '../templates/target';
 
 
-type PassportRequest = express.Request & {
-	user: any,
-	session: any,
-	query: any
-};
-
 const {ADMIN} = PrivilegeType;
+
+type SearchResultsT = {
+	initialResults: any,
+	query: string,
+	total?: number
+};
 
 const router = express.Router();
 
@@ -47,14 +48,14 @@ const router = express.Router();
  * Generates React markup for the search page that is rendered by the user's
  * browser.
  */
-router.get('/', auth.isAuthenticated, auth.isAuthorized(ADMIN), async (req: PassportRequest, res, next) => {
+router.get('/', auth.isAuthenticated, auth.isAuthorized(ADMIN), async (req, res, next) => {
 	const {orm} = req.app.locals;
-	const query = req.query.q ?? '';
+	const query = parseQuery(req.url).get('q') ?? '';
 	const type = 'editor';
-	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
-	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
+	const size = req.query.size ? parseInt(parseQuery(req.url).get('size'), 10) : 20;
+	const from = req.query.from ? parseInt(parseQuery(req.url).get('from'), 10) : 0;
 	try {
-		let searchResults: any = {
+		let searchResults: SearchResultsT = {
 			initialResults: [],
 			query,
 			total: 0
--- a/src/server/routes/adminPanel.tsx
+++ b/src/server/routes/adminPanel.tsx
@@ -26,14 +26,20 @@ import {snakeCase as _snakeCase, isNil} from 'lodash';
 import {escapeProps, generateProps} from '../helpers/props';
 import AdminPanelSearchPage from '../../client/components/pages/admin-panel-search';
 import Layout from '../../client/containers/layout';
-import {PrivilegeTypes} from '../../common/helpers/privileges-utils';
+import {PrivilegeType} from '../../common/helpers/privileges-utils';
 import React from 'react';
 import ReactDOMServer from 'react-dom/server';
 import express from 'express';
 import target from '../templates/target';
 
 
-const ADMIN = PrivilegeTypes.ADMIN_PRIV.value;
+type PassportRequest = express.Request & {
+	user: any,
+	session: any,
+	query: any
+};
+
+const {ADMIN} = PrivilegeType;
 
 const router = express.Router();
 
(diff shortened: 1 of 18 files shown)


src/server/routes/adminPanel.tsx
```diff
@@ -26,29 +26,35 @@ import {snakeCase as _snakeCase, isNil} from 'lodash';
...
 import express from 'express';
 import target from '../templates/target';
 
 
-const ADMIN = PrivilegeTypes.ADMIN_PRIV.value;
+type PassportRequest = express.Request & {
+	user: any,
+	session: any,
+	query: any
+};
```
kellnerd, line 40: I'm not a fan of having to redeclare this type in every module which uses custom request properties. This will only get worse once we upgrade more routes from JS to TS.
I know that there is already existing code which does this, but even having this declaration in one place and importing it to overwrite the standard parameter type of so many request handlers is suboptimal.

Fortunately, there is a better way to declare these properties in one place which requires no change at all in all of these route handler modules:
```ts
// Extend the express Request type via declaration merging
// Precise type declarations for `user` and `session` would be even better, of course.
declare module 'express-serve-static-core' {
	interface Request {
		user: any;
		session: any;
	}
}
```
It does not really matter in which module this is declared, but ideally it should reside wherever these properties are attached to the standard request. This would probably be in `auth.ts` and `session.js` (then `session.ts`), i.e. in two steps of declaration merging.

Overwriting the `query` property with `any` should not be done in any case, by the way. It may mute the type error because the default query parser has an ugly return type of `string | ParsedQs | string[] | ParsedQs[]` for each query parameter while we usually expect a string, but this is not safe.
When I was working on the Wikipedia PR, I had a similar problem and used a workaround to avoid having to use express' query parser:

https://github.com/metabrainz/bookbrainz-site/blob/1029d42f3f710cc3e1aeba525ead1b443d131d40/src/server/routes/wikimedia.ts#L25-L36

Maybe we should move that helper method into our utilities and make it the new standard? :grin:

src/server/routes/editor.tsx
```diff
@@ -73,7 +79,7 @@ router.get('/edit', auth.isAuthenticated, async (req, res, next) => {
 	// Parallel fetch the three required models. Only "editorModelPromise" has
 	// "require: true", so only that can throw a NotFoundError, which is why the
 	// other two model fetch operations have no error handling.
-	const [editorModel, titleUnlockModel, genderModel] = await Promise.all([
+	const [editorModel, titleUnlockModel, genderModel]: any = await Promise.all([
```
kellnerd, line 82: The types for these models should probably be declared in `bookbrainz-data`, which is something I will probably continue to work on for my GSoC project soon. If you want, you can also have a look yourself and help to add type definitions for a few more models here: https://github.com/metabrainz/bookbrainz-data-js/tree/master/src/types

src/server/routes/adminPanel.tsx
```diff
@@ -26,29 +26,35 @@ import {snakeCase as _snakeCase, isNil} from 'lodash';
...
-router.get('/', auth.isAuthenticated, auth.isAuthorized(ADMIN), async (req, res, next) => {
+router.get('/', auth.isAuthenticated, auth.isAuthorized(ADMIN), async (req: PassportRequest, res, next) => {
 	const {orm} = req.app.locals;
 	const query = req.query.q ?? '';
 	const type = 'editor';
 	const size = req.query.size ? parseInt(req.query.size, 10) : 20;
 	const from = req.query.from ? parseInt(req.query.from, 10) : 0;
 	try {
-		let searchResults = {
+		let searchResults: any = {
```
kellnerd, line 57: Using `any` is almost never a good idea, it's always better to start with a minimal object type definition which can be improved later.

src/server/routes/entity/author.ts
```diff
@@ -305,7 +309,7 @@ export function authorToFormState(author) {
 		}
 	));
 
-	const optionalSections = {};
+	const optionalSections: any = {};
```
kellnerd, line 312: See my previous comment :wink:

kellnerd (changes requested):
I know that adding types to existing JS code often is a pain, so I've added a few comments based on my prior experiences in this area :grin:

src/server/routes/adminPanel.tsx
the-good-boy: Thanks for these suggestions. I've made some changes.

src/server/routes/editor.tsx
the-good-boy: I agree that it would be better if we can define them in `bookbrainz-data`. I'm marking this as a stretch-goal for now.

src/server/routes/reviews.ts
```diff
@@ -45,7 +45,7 @@ router.post('/:entityType/:bbid/reviews', auth.isAuthenticated, auth.isAuthorize
 		review
 	);
 
-	let newAccessToken = '';
+	let newAccessToken: any = '';
```
MonkeyDo, line 48: Should this be `string` rather than `any`?

MonkeyDo (commented):
Good improvements, nice work !

anshg1214:
When transitioning to TypeScript, it is necessary to install type definitions for packages that do not come with built-in TypeScript support. So we would have to install type definitions for `passport`, `log`, and all similar packages.

src/server/routes/reviews.ts
the-good-boy: It shouldn't be string, because it returns a JSON object and we're using one of its values below: `newAccessToken.access_token`.

I suppose we can do something like this:
`let newAccessToken: Record<string, any> = {};`

What do you think?

kellnerd: Either that or not assigning an initial value at all:
```suggestion
	let newAccessToken: any;
```
If the token fetching logic throws, we don't even reach the code block where we try to access `newAccessToken.access_token`, so it does not matter whether `newAccessToken` is an object.

the-good-boy: I just realised that we already have a type defined for this on `bookbrainz-data`. I think this works.

PR #1001, opened 2023-07-18
feat(Admin System): Admin logs page

the-good-boy:
This PR adds an Admin logs page.

I have also added a field for adding a note/reason in the privs-edit-modal.

### DIFF
--- a/src/client/components/footer.js
+++ b/src/client/components/footer.js
@@ -64,7 +64,22 @@ function Footer(props) {
 								Admin logs
 							</a>
 						</small>
-						&nbsp;—&nbsp;
+					</Col>
+				</Row>
+				<Row>
+					<Col xs={4}/>
+					<Col className="text-center" xs={4}>
+						<small>
+							Alpha Software —{' '}
+							<a href={`${repositoryUrl}tree/${siteRevision || 'master'}`}>
+								{siteRevision || 'unknown revision'}
+							</a> —&nbsp;
+							<a href="https://tickets.metabrainz.org/projects/BB/issues/">
+								Report a Bug
+							</a>
+						</small>
+					</Col>
+					<Col className="text-right" xs={4}>
 						<small>
 							<a href="/privacy">
 								Privacy & Terms
@@ -72,17 +87,6 @@ function Footer(props) {
 						</small>
 					</Col>
 				</Row>
-				<div className="text-center">
-					<small>
-						Alpha Software —{' '}
-						<a href={`${repositoryUrl}tree/${siteRevision || 'master'}`}>
-							{siteRevision || 'unknown revision'}
-						</a> —&nbsp;
-						<a href="https://tickets.metabrainz.org/projects/BB/issues/">
-							Report a Bug
-						</a>
-					</small>
-				</div>
 			</Container>
 		</footer>
 	);
--- a/src/client/components/footer.js
+++ b/src/client/components/footer.js
@@ -60,6 +60,12 @@ function Footer(props) {
 					</Col>
 					<Col className="text-right" xs={4}>
 						<small>
+							<a href="/admin-logs">
+								Admin logs
+							</a>
+						</small>
+						&nbsp;—&nbsp;
+						<small>
 							<a href="/privacy">
 								Privacy & Terms
 							</a>
--- a/src/client/components/pages/adminLogs.js
+++ b/src/client/components/pages/adminLogs.js
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2023 Shivam Awasthi
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import AdminLogsTable from './parts/admin-logs-table';
+import PagerElement from './parts/pager';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+class AdminLogsPage extends React.Component {
+	constructor(props) {
+		super(props);
+		this.state = {
+			results: this.props.results
+		};
+
+		// React does not autobind non-React class methods
+		this.searchResultsCallback = this.searchResultsCallback.bind(this);
+		this.paginationUrl = './admin-logs/admin-logs';
+	}
+
+	searchResultsCallback(newResults) {
+		this.setState({results: newResults});
+	}
+
+	render() {
+		return (
+			<div id="pageWithPagination">
+				<AdminLogsTable
+					results={this.state.results}
+				/>
+				<PagerElement
+					from={this.props.from}
+					nextEnabled={this.props.nextEnabled}
+					paginationUrl={this.paginationUrl}
+					results={this.state.results}
+					searchResultsCallback={this.searchResultsCallback}
+					size={this.props.size}
+				/>
+			</div>
+		);
+	}
+}
+
+
+AdminLogsPage.displayName = 'AdminLogsPage';
+AdminLogsPage.propTypes = {
+	from: PropTypes.number,
+	nextEnabled: PropTypes.bool.isRequired,
+	results: PropTypes.array,
+	size: PropTypes.number
+};
+AdminLogsPage.defaultProps = {
+	from: 0,
+	results: [],
+	size: 20
+};
+
+export default AdminLogsPage;
--- a/src/client/components/pages/parts/admin-log-statement.js
+++ b/src/client/components/pages/parts/admin-log-statement.js
@@ -24,7 +24,7 @@ import {constructAdminLogStatement} from '../../../helpers/adminLogs';
 function AdminLogStatement({logData}) {
 	const {note} = logData;
 	return (
-		<div>
+		<>
 			{constructAdminLogStatement(logData)}
 			{
 				note.length ?
@@ -32,7 +32,7 @@ function AdminLogStatement({logData}) {
 						<strong>Note/Reason:&nbsp;</strong>{note}
 					</div> : null
 			}
-		</div>
+		</>
 	);
 }
 
--- a/src/client/components/pages/parts/admin-log-statement.js
+++ b/src/client/components/pages/parts/admin-log-statement.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright (C) 2023 Shivam Awasthi
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import PropTypes from 'prop-types';
+import React from 'react';
+import {constructAdminLogStatement} from '../../../helpers/adminLogs';
+
+
+function AdminLogStatement({logData}) {
+	const {note} = logData;
+	return (
+		<div>
+			{constructAdminLogStatement(logData)}
+			{
+				note.length ?
+					<div className="small">
+						<strong>Note/Reason:&nbsp;</strong>{note}
+					</div> : null
+			}
+		</div>
+	);
+}
+
+AdminLogStatement.propTypes = {
+	logData: PropTypes.object.isRequired
+};
+
+AdminLogStatement.displayName = 'AdminLogStatement';
+
+export default AdminLogStatement;
--- a/src/client/components/pages/parts/admin-logs-table.js
+++ b/src/client/components/pages/parts/admin-logs-table.js
@@ -16,15 +16,12 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import * as bootstrap from 'react-bootstrap';
-import * as utilsHelper from '../../../helpers/utils';
 import AdminLogStatement from './admin-log-statement';
 import PropTypes from 'prop-types';
 import React from 'react';
+import {Table} from 'react-bootstrap';
+import {formatDate} from '../../../helpers/utils';
 
-
-const {Table} = bootstrap;
-const {formatDate} = utilsHelper;
 
 function AdminLogsTable(props) {
 	const {results, tableHeading} = props;
--- a/src/client/components/pages/parts/admin-logs-table.js
+++ b/src/client/components/pages/parts/admin-logs-table.js
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2023 Shivam Awasthi
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import * as utilsHelper from '../../../helpers/utils';
+import AdminLogStatement from './admin-log-statement';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+const {Table} = bootstrap;
+const {formatDate} = utilsHelper;
+
+function AdminLogsTable(props) {
+	const {results, tableHeading} = props;
+	return (
+		<div>
+			<div>
+				<h1 className="text-center">{tableHeading}</h1>
+			</div>
+			<hr className="thin"/>
+			{
+				results.length > 0 ?
+					<Table
+						borderless
+						responsive
+						striped
+					>
+						<thead>
+							<tr>
+								<th width="85%">Action</th>
+								<th width="15%">Date</th>
+							</tr>
+						</thead>
+
+						<tbody>
+							{
+								results.map((logData) => (
+									<tr key={logData.id}>
+										<td>
+											<AdminLogStatement logData={logData}/>
+										</td>
+										<td>
+											{formatDate(new Date(logData.time), true)}
+										</td>
+									</tr>
+								))
+							}
+						</tbody>
+					</Table> :
+
+					<div>
+						<h4> No logs to show</h4>
+						<hr className="wide"/>
+					</div>
+			}
+		</div>
+
+	);
+}
+
+AdminLogsTable.propTypes = {
+	results: PropTypes.array.isRequired,
+	tableHeading: PropTypes.node
+};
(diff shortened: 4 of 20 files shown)


src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,130 @@
...
+		grantStatement += ' privilege';
+		if (privsAdded.length > 1) {
+			grantStatement += 's';
+		}
+	}
+	const andStatement = privsAdded.length && privsRemoved.length ? ' and' : '';
+
+	let removedStatement = '';
+	if (privsRemoved.length) {
+		removedStatement = ' removed ';
```
MonkeyDo, line 81: I'd suggest the same method as above using native array  functions instead of the for loop.

src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,130 @@
...
+		if (privsRemoved.length > 1) {
+			removedStatement += 's';
+		}
+	}
+
+	const preposition = privsRemoved.length ? ' from ' : ' to ';
+
+	const finalStatement = grantStatement + andStatement + removedStatement + preposition;
+	/* eslint-disable react/no-danger */
+	// We are disabling this rule because we are already sanitizing the html here
```
MonkeyDo, line 105: :+1:  thanks for the comment

src/client/components/pages/adminLogs.js
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+import AdminLogsTable from './parts/admin-logs-table';
+import PagerElement from './parts/pager';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+class AdminLogsPage extends React.Component {
+	constructor(props) {
```
MonkeyDo, line 26: How would you feel about making this component a functional component instead of a class?
It seems very simple to convert and we might as well try to stay up to date with the newer syntax when we can.
Let me know if you need a hand.

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -106,6 +115,19 @@ class PrivsEditModal extends React.Component {
 			/>
 		));
 
+		const noteField = this.state.submittable &&
+			<Form.Group>
+				<Form.Label>
+					Note/Reason:
+				</Form.Label>
+				<Form.Control
```
MonkeyDo, line 123: We probably need this input to be required.
I don't think there is currently validation on the server-side for an empty note, but we should make it as hard as possible to take admin actions without leaving some reason for it.
So possibly we should throw an error in the server route if the note is empty, as well as enforcing the front-end input?

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -93,6 +96,12 @@ class PrivsEditModal extends React.Component {
 		}
 	}
 
+	handleNoteChange(event) {
+		this.setState({
+			note: event.target.value
```
MonkeyDo, line 101: Maybe adding `submittable: Boolean(event.target.value?.length)` here would be a good safeguard

src/client/components/footer.js
```diff
@@ -59,6 +59,12 @@ function Footer(props) {
 						</small>
 					</Col>
 					<Col className="text-right" xs={4}>
+						<small>
+							<a href="/admin-logs">
```
MonkeyDo, line 63: Yes please.
I would also like to see this link duplicated in the "privileges" dropdown in the menu bar for easy access (considering this will eventually only be visible to a limited amount of users and is quite a utilitarian menu item, I think the duplication is fine in this case.

MonkeyDo (changes requested):
Looking good!
Currently up on the test server: https://test.bookbrainz.org/admin-logs

MonkeyDo:
Forgot to add a comment: please make as many as possible of these new files into typescript files (.ts for javascript only files and .tsx for the ones that contain JSX syntax). You seem to be getting a good handle on typing with typescript, so let's continue pursuing that goal.

src/common/helpers/privileges-utils.ts
```diff
@@ -30,6 +30,10 @@ export enum PrivilegeType {
 	ENTITY_EDITOR = 1
 }
 
+export enum AdminActionType {
+	CHANGE_PRIV = 'Change Privileges'
```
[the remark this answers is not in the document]
the-good-boy: We've defined this type like [this](https://github.com/metabrainz/bookbrainz-site/blob/e7a36db220f0130fa102156345a3bc24a4b84539/sql/schemas/bookbrainz.sql#L23C1-L25C3). So I think it should remain capitalized here.

src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,130 @@
...
+				grantStatement += ' and ';
+				grantStatement += `<strong>${privsAdded[i]}</strong>`;
+			}
+			else if (i !== 0) {
+				grantStatement += ', ';
+				grantStatement += `<strong>${privsAdded[i]}</strong>`;
+			}
+			else {
+				grantStatement += `<strong>${privsAdded[i]}</strong>`;
+			}
```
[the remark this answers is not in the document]
the-good-boy: Nice. Thanks for this suggestion.

the-good-boy:
> Forgot to add a comment: please make as many as possible of these new files into typescript files (.ts for javascript only files and .tsx for the ones that contain JSX syntax). You seem to be getting a good handle on typing with typescript, so let's continue pursuing that goal.

Noted. I will make sure to make all the new files into typescript from now on.

src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,130 @@
...
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as React from 'react';
+import {AdminActionType, PrivilegeTypes, getPrivilegeTitleFromBit} from '../../common/helpers/privileges-utils';
+import {sanitize} from 'isomorphic-dompurify';
+
+
+/* eslint-disable no-bitwise */
+function getPrivsAdded(newPrivs, oldPrivs) {
```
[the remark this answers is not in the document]
the-good-boy: fixed

src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,130 @@
...
+	for (let i = 0; i < totalBits; i++) {
+		if (!(oldPrivs & (1 << i)) && (newPrivs & (1 << i))) {
+			privsAdded.push(getPrivilegeTitleFromBit(i));
+		}
+	}
+
+	return privsAdded;
+}
+
+function getPrivsRemoved(newPrivs, oldPrivs) {
```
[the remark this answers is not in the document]
the-good-boy: fixed

src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,130 @@
...
+	for (let i = 0; i < totalBits; i++) {
+		if (!(newPrivs & (1 << i)) && (oldPrivs & (1 << i))) {
+			privsRemoved.push(getPrivilegeTitleFromBit(i));
+		}
+	}
+
+	return privsRemoved;
+}
+
+function constructPrivsChangeStatement(logData) {
```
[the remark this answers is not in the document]
the-good-boy: fixed

src/client/components/pages/adminLogs.js
the-good-boy: fixed!

src/client/components/pages/parts/privs-edit-modal.js
the-good-boy: It will require some more conditions to work. I have made some changes, please see.

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -80,14 +83,47 @@ class PrivsEditModal extends React.Component {
...
+					privs: newPrivs,
+					submittable: true
+				});
+			}
+			else {
+				this.setState({
+					privs: newPrivs,
+					submittable: false
+				});
+			}
```
kellnerd, line 98:
```suggestion
			// If we have also added a note, then set submittable also true
			this.setState({
				privs: newPrivs,
				submittable: Boolean(this.state.note.length)
			});
```
Alternatively, if you are a fan of cryptic code, like me:
```suggestion
			// If we have also added a note, then set submittable also true
			this.setState({
				privs: newPrivs,
				submittable: !!this.state.note.length
			});
```
This simplification could even be extended to include the else-branch below, but GitHub doesn't allow me to make suggestions for deleted lines.

src/server/routes/adminLogs.tsx
```diff
@@ -0,0 +1,88 @@
...
+import {parseQuery} from '../helpers/utils';
+import target from '../templates/target';
+
+
+const router = express.Router();
+
+router.get('/', async (req, res, next) => {
+	const {orm} = req.app.locals;
+	const size = req.query.size ? parseInt(parseQuery(req.url).get('size'), 10) : 20;
+	const from = req.query.from ? parseInt(parseQuery(req.url).get('from'), 10) : 0;
```
kellnerd, line 37:
```suggestion
	const query = parseQuery(req.url);
	const size = parseInt(query.get('size'), 10) || 20;
	const from = parseInt(query.get('from'), 10) || 0;
```
There's no need to do the URL parsing twice. Since there are also a few occurrences of this pattern in the code from #1000 you may want to also fix those while you're at it :innocent:
The other simplification works because `parseInt(null) === NaN` and `parseInt('') === NaN` and NaN is a falsy value. Considering how often this pattern is used maybe it is also worth having a helper function `getInt(query, fallback = 0)`?

the-good-boy: Thanks for the suggestions! I have made the changes.

the-good-boy: There are some other files where we can use these helper functions, but I guess its better to make those changes in a different PR later.

the-good-boy: Thank you also for finding time to lend an extra eye on these PRs.

src/server/helpers/utils.ts
```diff
@@ -361,3 +361,7 @@ export async function parseInitialState(req, type):Promise<Record<string, any>>
 export function parseQuery(url: string) {
 	return new URLSearchParams(url.replace(/^.+?\?/, ''));
 }
+
+export function getIntFromQueryParams(query: URLSearchParams, type: string, fallback = 0) {
+	return parseInt(query.get(type), 10) || fallback;
+}
```
kellnerd, line 367: Nitpick: The parameter is usually called `name` to follow the [specification](https://url.spec.whatwg.org/#interface-urlsearchparams) :nerd_face:
```suggestion
export function getIntFromQueryParams(query: URLSearchParams, name: string, fallback = 0) {
	return parseInt(query.get(name), 10) || fallback;
}
```

src/server/routes/adminPanel.tsx
```diff
@@ -52,8 +52,9 @@ router.get('/', auth.isAuthenticated, auth.isAuthorized(ADMIN), async (req, res,
 	const {orm} = req.app.locals;
 	const query = parseQuery(req.url).get('q') ?? '';
```
kellnerd, line 53:
```suggestion
	const urlQuery = parseQuery(req.url);
	const query = urlQuery.get('q') ?? '';
```
Here's one missed occurrence of `parseQuery` which can be combined with the next one.

src/server/routes/adminPanel.tsx
```diff
@@ -52,8 +52,9 @@ router.get('/', auth.isAuthenticated, auth.isAuthorized(ADMIN), async (req, res,
...
 	const type = 'editor';
-	const size = req.query.size ? parseInt(parseQuery(req.url).get('size'), 10) : 20;
-	const from = req.query.from ? parseInt(parseQuery(req.url).get('from'), 10) : 0;
+	const urlQry = parseQuery(req.url);
+	const size = getIntFromQueryParams(urlQry, 'size', 20);
+	const from = getIntFromQueryParams(urlQry, 'from');
```
kellnerd, line 57:
```suggestion
	const size = getIntFromQueryParams(urlQuery, 'size', 20);
	const from = getIntFromQueryParams(urlQuery, 'from');
```
See my comment above. Let's be generous and grant the variable name two more letters as well :grinning:

src/client/helpers/adminLogs.tsx
```diff
@@ -0,0 +1,119 @@
...
+	const totalBits = Object.keys(PrivilegeTypes).length;
+
+	for (let i = 0; i < totalBits; i++) {
+		if (!(newPrivs & (1 << i)) && (oldPrivs & (1 << i))) {
+			privsRemoved.push(getPrivilegeTitleFromBit(i));
+		}
+	}
+
+	return privsRemoved;
+}
```
kellnerd, line 50: I have the feeling that these functions could be combined to have only one iteration over the privileges, but I don't know whether it's worth the performance gain or whether there are use case where we will need one but no the other set. Maybe it's better to keep it as is then.

kellnerd (commented):
Two more minor suggestions and this looks good to me.

src/client/helpers/adminLogs.tsx
the-good-boy: I don't think there will be any performance gains.

src/server/routes/adminPanel.tsx
the-good-boy: fixed!

src/server/routes/adminPanel.tsx
the-good-boy: haha! fixed.

src/client/components/pages/parts/privs-edit-modal.js
MonkeyDo: This bit of feedback left.

src/client/helpers/adminLogs.tsx
MonkeyDo: I think it's pretty easy to refactor actually, and will definitely be a performance gain( not one that will be visible to us humans though :D ).
Just combining putting both `if` blocks inside the same for loop and returning an object `{privsAdded, privsRemoved}` should run about twice as fast and is about the same level of readability IMO.
That would make the method name `getPrivsChanged` ?
And then you can `const {privsAdded, privsRemoved} = getPrivsChanged(newPrivs, oldPrivs);` easily.

src/server/routes/adminLogs.tsx
```diff
@@ -0,0 +1,90 @@
...
+		res.send(target({
+			markup,
+			props: escapeProps(props),
+			script: '/js/adminLogs.js',
+			title: 'Admin Logs'
+		}));
+	}
+
+	try {
+		// fetch 1 more revision than required to check nextEnabled
```
MonkeyDo, line 63:
```suggestion
		// fetch 1 more log item than required to check nextEnabled
```

src/server/routes/editor.tsx
```diff
@@ -179,18 +180,31 @@ router.post('/edit/handler', auth.isAuthenticatedForHandler, (req, res) => {
...
+			.fetch({require: true, transacting: trx})
 			.catch(Editor.NotFoundError, () => next(new error.NotFoundError('Editor not found', req)));
-		if (editor.get('privs') === req.body.newPrivs) {
+		if (editor.get('privs') === newPrivs) {
 			return next(new error.FormSubmissionError(
 				'No change to Privileges', req
 			));
 		}
-		await editor.save({privs: req.body.newPrivs});
+		await editor.save({privs: newPrivs});
```
MonkeyDo, line 197: It's for this save operation that we really need the SQL transaction:
```suggestion
		await editor.save({privs: newPrivs},{transacting: trx});
```

src/server/routes/editor.tsx
```diff
@@ -179,18 +180,31 @@ router.post('/edit/handler', auth.isAuthenticatedForHandler, (req, res) => {
...
+		const logData: AdminLogDataT = {
+			actionType,
+			adminId,
+			newPrivs,
+			note,
+			oldPrivs,
+			targetUserId
+		};
+		await createAdminLog(logData, AdminLog, trx);
+		await trx.commit();
```
MonkeyDo, line 207: This is unnecessary actually, my bad. From the [docs](https://bookshelfjs.org/api.html#Bookshelf-instance-transaction):

> Note that there is no need to explicitly call transaction.commit() or transaction.rollback() since the entire transaction will be committed if there are no errors inside the transaction block.

MonkeyDo (commented):
Looking nicer ! I have some more comments, but I think we're almost done with this PR.

src/server/routes/adminLogs.tsx
MonkeyDo: > Thank you also for finding time to lend an extra eye on these PRs.

I second that !

src/client/helpers/adminLogs.tsx
kellnerd: I would call it `getChangedPrivs` then, for what it's worth.

src/server/routes/editor.tsx
the-good-boy: I think we'll have to do this. This documentation is perhaps only for when we are using transaction in the form of a block as shown in the example. I am getting errors if I remove this.

src/client/components/pages/parts/privs-edit-modal.js
the-good-boy: I've made some changes regarding this. I will also make sure to add some more tests for checking this later.

src/server/routes/adminLogs.tsx
```diff
@@ -0,0 +1,90 @@
...
+// eslint-disable-next-line consistent-return
+router.get('/admin-logs', async (req, res, next) => {
+	const {orm} = req.app.locals;
+	const query = parseQuery(req.url);
+	const size = getIntFromQueryParams(query, 'size', 20);
+	const from = getIntFromQueryParams(query, 'from');
+
+	try {
+		const orderedRevisions = await getOrderedAdminLogs(from, size, orm);
+		res.send(orderedRevisions);
```
anshg1214, line 83:
```suggestion
		res.json(orderedRevisions);
```

src/client/components/pages/parts/admin-log-statement.js
```diff
@@ -0,0 +1,45 @@
...
+	return (
+		<div>
+			{constructAdminLogStatement(logData)}
+			{
+				note.length ?
+					<div className="small">
+						<strong>Note/Reason:&nbsp;</strong>{note}
+					</div> : null
+			}
+		</div>
```
anshg1214, line 35: You can consider using React Fragments (<>...</>) here to avoid unnecessary wrapping elements.
```suggestion
		<>
			{constructAdminLogStatement(logData)}
			{
				note.length ?
					<div className="small">
						<strong>Note/Reason:&nbsp;</strong>{note}
					</div> : null
			}
		</>
```

src/client/components/pages/parts/admin-logs-table.js
```diff
@@ -0,0 +1,87 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
+ * with this program; if not, write to the Free Software Foundation, Inc.,
+ * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+import * as bootstrap from 'react-bootstrap';
+import * as utilsHelper from '../../../helpers/utils';
```
anshg1214, line 20: You should consider importing only the components you need, instead of importing the entire package. This would otherwise lead to larger bundle sizes. 

Similarly [here](https://github.com/metabrainz/bookbrainz-site/pull/1001/files#diff-ec1b09b3858b2b94f7808ba9474ca2a6044e10b2222bae71c3be501cb1d41343R19) as well.

anshg1214 (changes requested):
The PR looks amazing now.

One more thing, in [this](https://github.com/the-good-boy/bookbrainz-site/blob/3550b5b5fd78cea485aa6efdf7f5c6256e6bbabc/src/client/components/pages/parts/privs-edit-modal.js#L77-L79) case, it would be better if you could show a text message on the modal saying, "Something went wrong".

src/server/helpers/adminLogs.ts
```diff
@@ -0,0 +1,69 @@
...
+	admin?: any,
+	adminId: number,
+	id?: number,
+	newPrivs: number,
+	note: string,
+	oldPrivs: number,
+	targetUser?: any,
+	targetUserId: number,
+	time?: any
+};
```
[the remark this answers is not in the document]
MonkeyDo: I agree, but would suggest to do this as a separate PR, so we can move forward.
Since it will require publishing a new version of the bookbrainz-data package, we might as well collect all these types that need to be moved, and do as many as possible in one go to reduce the back-and-forth work.

src/client/components/pages/parts/privs-edit-modal.js
```diff
@@ -106,6 +127,21 @@ class PrivsEditModal extends React.Component {
...
 
+		const noteField = (
+			<Form.Group>
+				<Form.Label>
+					Note/Reason:
+				</Form.Label>
+				<Form.Control
+					required
+					as="textarea"
+					defaultValue={this.state.note}
```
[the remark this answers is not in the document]
the-good-boy: I agree.

anshg1214 (approved):
Great Work 🚀

PR #1013, opened 2023-08-28
fix: Redirect users after login

anshg1214:
The users are currently redirected to home page after login.
This PR fixes this issue by redirecting the users to their original link they desire.


MonkeyDo (approved):
Thank you !
The issue was hard to find, no?

PR #1021, opened 2023-09-21
Fix linting errors and warnings

MonkeyDo:
Simple fixes.
There are two warnings left regarding todos that would need to be fixed in a separate PR  (the "do" in "todo")


kellnerd (approved):
LGTM

PR #1025, opened 2023-10-05
refactor(profile): refactor achievement to typescript

meziyum:
As the bookbrainz has been actively moving on towards typescript. I think it would be a good idea to refactor the old files.

Affected Areas: Profile Page and Achievement


src/client/components/forms/parts/achievement.tsx
```diff
@@ -57,8 +57,37 @@ const maxAchievementProgress = {
...
 
-function Achievement(props) {
-	const {achievement, counter, unlocked} = props;
+interface Props {
+	achievement: {
+		id: number;
+		name: string;
+		description: string;
+		badgeUrl: string;
+	},
```
MonkeyDo, line 66: We have a type already defined for achievement which can be imported.
It is currently being improved in #1033 (the `badgeUrl` prop needs fixing) so you will need to wait until that PR is merged before importing it

src/client/components/forms/parts/achievement.tsx
```diff
@@ -57,8 +57,37 @@ const maxAchievementProgress = {
 	30: 100
 };
 
-function Achievement(props) {
-	const {achievement, counter, unlocked} = props;
+interface Props {
```
MonkeyDo, line 60: Let's use a more precise name here to avoid confusion:

```suggestion
interface AchievementProps {
```

MonkeyDo (changes requested):
Thanks for working on this :)

And sorry it took me so long to get to reviewing it !

MonkeyDo (approved):

PR #1032, opened 2023-10-28
BB-743: Hide External Service Tab for other profiles

anshg1214:
BB-743: Hide External Service Tab for other profiles

### DIFF
--- a/src/client/controllers/editor/achievement.js
+++ b/src/client/controllers/editor/achievement.js
@@ -37,6 +37,7 @@ ReactDOM.hydrate(
 	<AppContainer>
 		<Layout {...extractLayoutProps(props)}>
 			<EditorContainer
+				user={props.user}
 				{...extractEditorProps(props)}
 			>
 				<AchievementsTab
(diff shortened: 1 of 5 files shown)


src/client/controllers/editor/achievement.js
```diff
@@ -37,6 +37,7 @@ ReactDOM.hydrate(
 	<AppContainer>
 		<Layout {...extractLayoutProps(props)}>
 			<EditorContainer
+				user={props.user}
```
MonkeyDo, line 40: I think I would rather see all these new `user={props.user}`  lines replaced with a single change to avoid passing props this way and also with the existing `extractEditorProps`.

So I think it's as simple as adding `'user'` to the `EDITOR_PROPS` here:
 https://github.com/metabrainz/bookbrainz-site/blob/9c4740de09628d2be051862b02e8795be4f4d9ce/src/client/helpers/props.js#L33-L36

MonkeyDo (changes requested):
Thanks for fixing this, I never realized :)

MonkeyDo (approved):

PR #1033, opened 2023-10-29
BB-722: Display already selected achievements in Edit Achievement Page

anshg1214:
[BB-722](https://tickets.metabrainz.org/projects/BB/issues/BB-722): Display already selected achievements in Edit Achievement Page

### DIFF
--- a/src/client/components/input/drag-and-drop.tsx
+++ b/src/client/components/input/drag-and-drop.tsx
@@ -28,8 +28,8 @@ const {useState, useCallback} = React;
 * The Achievement interface, defining the properties of an achievement.
 * @typedef
 * @property {string} name - The name of the achievement.
-* @property {string} src - The source URL of the achievement's badge image.
-* @property {string} id - The ID of the achievement.
+* @property {string} badgeUrl - The source URL of the achievement's badge image.
+* @property {number} id - The ID of the achievement.
 */
 type Achievement = {
 	name: string;
@@ -41,6 +41,7 @@ type Achievement = {
  * Props for DragAndDropImage component
  * @typedef {Object} Props
  * @property {string} name - The name of the DragAndDrop component.
+ * @property {Achievement} initialAchievement - The initial achievement to display in the card.
  */
 type Props = {
   name: string;
--- a/src/client/components/input/drag-and-drop.tsx
+++ b/src/client/components/input/drag-and-drop.tsx
@@ -33,8 +33,8 @@ const {useState, useCallback} = React;
 */
 type Achievement = {
 	name: string;
-	src: string;
-	id: string;
+	badgeUrl: string;
+	id: number;
 };
 
 /**
@@ -44,6 +44,7 @@ type Achievement = {
  */
 type Props = {
   name: string;
+  initialAchievement?: Achievement;
 };
 
 /**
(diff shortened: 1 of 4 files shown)


src/client/components/input/drag-and-drop.tsx
```diff
@@ -28,22 +28,24 @@ const {useState, useCallback} = React;
...
-* @property {string} id - The ID of the achievement.
+* @property {string} badgeUrl - The source URL of the achievement's badge image.
+* @property {number} id - The ID of the achievement.
 */
 type Achievement = {
 	name: string;
-	src: string;
-	id: string;
+	badgeUrl: string;
+	id: number;
```
MonkeyDo, line 37: We do set this to `null` in handleClick, I'm surprised Typescript doesn't scream about it.

The type should probably explicitly allow for this:
```suggestion
	id: number | null;
```

src/client/components/input/drag-and-drop.tsx
```diff
@@ -111,7 +115,18 @@ function DragAndDrop({name}: Props): JSX.Element {
 
 DragAndDrop.displayName = 'DragAndDrop';
 DragAndDrop.propTypes = {
+	initialAchievement: PropTypes.shape({
+		badgeUrl: PropTypes.string,
+		id: PropTypes.number,
```
MonkeyDo, line 120: Here we don't need the `| null` like above, as a previously selected achievement will always have an id.

src/client/components/input/drag-and-drop.tsx
```diff
@@ -111,7 +115,18 @@ function DragAndDrop({name}: Props): JSX.Element {
...
+		name: PropTypes.string.isRequired
+	}),
 	name: PropTypes.string.isRequired
 };
+DragAndDrop.defaultProps = {
+	initialAchievement: {
+		badgeUrl: '/images/blankbadge.svg',
+		name: 'drag badge to set'
```
MonkeyDo, line 128: This should probably also have `id:null` defined in the default achievement

src/server/routes/editor.tsx
```diff
@@ -531,6 +531,16 @@ router.get('/:id/achievements', async (req, res, next) => {
 		.where('editor_id', userId)
 		.fetchAll({require: false});
 
+	const achievementCol = await new AchievementUnlock()
```
MonkeyDo, line 534: While `await`ing here works, it's a departure from the rest of this function where we create promises and then `await Promise.all(…` all the promises.

I'm not attached to Promise.all, especially if there is no strong reason why we need to await all promises at the same time.

I would suggest either following the current paradigm, or refactoring the promises so that they all get `await`-ed straight away instead of using Promise.all.
It will probably mean more try-catch blocks, for example to catch `getIdEditorJSONPromise` failures

Either way works for me.

MonkeyDo (commented):
Thanks, that's been bugging me :)

Some style comments below

src/client/components/input/drag-and-drop.tsx
anshg1214: Fixed

src/server/routes/editor.tsx
anshg1214: Fixed

src/server/routes/editor.tsx
```diff
@@ -531,19 +531,32 @@ router.get('/:id/achievements', async (req, res, next) => {
...
 
 	const [unlocks, editorJSON, achievementTypes] = await Promise.all([
-		unlocksPromise, editorJSONPromise, achievementTypesPromise
+		unlocksPromise, editorJSONPromise, achievementTypesPromise, achievementCol
 	]);
 
 	const achievementJSON =
 		await setAchievementUnlockedField(achievementTypes, unlocks, userId, res.app.locals.orm);
 
+	const currAchievementJSON = achievementColToEditorGetJSON(achievementCol);
```
MonkeyDo, line 555: `achievementCol` here is the promise itself rather than the values we get after awaiting it.
I would recommend renaming it to `achievementColPromise` like we do for the other ones, and you'll need to add a new `achievementCol` on line 548 where we destructure the results of Promise.all: 
`const [unlocks, editorJSON, achievementTypes, achievementCol] = await Promise.all(`

I suspect currently the code is not working.

MonkeyDo (approved):
One more change plz :)

MonkeyDo (changes requested):
Woops.

MonkeyDo (approved):

MonkeyDo:
Thank you ! :rocket:

PR #1041, opened 2023-12-18
BB-718 CritiqueBrainz auth endpoint is called server-side

kirtanchandak:
### Problem
[BB-718: CritiqueBrainz auth endpoint is called server-side](https://tickets.metabrainz.org/browse/BB-718?filter=11910)

### Solution
The conditional check - ```typeof window !== "undefined"``` ensures that the ```getAccessToken``` method is only called when the code is executed in a browser environment (client-side), preventing the unnecessary call to the authentication endpoint on the server-side.

### Areas of Impact
The changes were reflected in ```src/client/components/pages/entities/cbReviewModal.tsx```

### DIFF
--- a/src/client/components/pages/entities/cbReviewModal.tsx
+++ b/src/client/components/pages/entities/cbReviewModal.tsx
@@ -501,6 +501,7 @@ class CBReviewModal extends React.Component<
 
 	private accessToken = '';
 
+	/* executes getAccessToken() only in a browser to avoid unnecessary server-side calls during component mounting */
 	componentDidMount = async () => {
 		if (typeof window !== "undefined") {
 			this.accessToken = await this.getAccessToken();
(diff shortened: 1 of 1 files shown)


MonkeyDo (changes requested):
Thanks for submitting a PR @kirtanchandak 
Sorry it took so long for me to get to it!

That should do the trick!
Could you please add a comment above with a short explanation of why we are doing this check?
For future developers to know what's going on.

MonkeyDo:
In the future, please assign the ticket to yourself in Jira if you are working on one, to avoid someone else working on it at the same time.

kellnerd:
For what it's worth, the ticket was assigned to @kirtanchandak at the time they submitted the PR. It's the fault of the other contributor which reassigned the ticket to themselves later without asking and ignoring the link to this PR.

MonkeyDo:
Ah, my bad, I totally read that wrong! Thank you @kellnerd for correcting me and thank you @kirtanchandak for following the rules properly :)
Sorry for accusing you !

kirtanchandak:
@MonkeyDo no worries for the misunderstanding. 
I've added the comment.

MonkeyDo (approved):
Thank you ! :rocket:

PR #1043, opened 2023-12-18
other(editor): Integrated the modification of identifiers and aliases into the primary editing workflow

Tarunmeena0901:
### Problem
 [BB-343]

### Solution

Enhanced main entity editor page:
- Introduced an identifier editing section.
- Added a section for adding aliases, improving overall editing flow.
- Aligned the remove button with identifier rows for a more polished appearance.

### Areas of Impact

~ identifier-editor.js
~ identifier-row.tsx
~ alias-editor.js
~ entity-editor.tsx

editor-page

### DIFF
--- a/src/client/entity-editor/alias-editor/alias-editor.js
+++ b/src/client/entity-editor/alias-editor/alias-editor.js
@@ -16,8 +16,8 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import {Button, Modal, OverlayTrigger, Tooltip} from 'react-bootstrap';
-import {hideAliasEditor, removeEmptyAliases} from './actions';
+import {Button, OverlayTrigger, Tooltip} from 'react-bootstrap';
+import { removeEmptyAliases} from './actions';
 import AliasModalBody from './alias-modal-body';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import PropTypes from 'prop-types';
@@ -37,7 +37,7 @@ import {faQuestionCircle} from '@fortawesome/free-solid-svg-icons';
  * @param {Array} props.languageOptions - The list of possible languages for an
  *        alias.
  * @param {Function} props.onClose - A function to be called when the button to
- *        close the editor is clicked.
+ *        add new alias is clicked.
  * @param {boolean} props.show - Whether or not the editor modal should be
  *        visible.
  * @returns {ReactElement} React element containing the rendered AliasEditor.
@@ -63,37 +63,37 @@ const AliasEditor = ({
 	);
 
 	return (
-		<Modal show={show} size="lg" onHide={onClose}>
-			<Modal.Header>
-				<Modal.Title>
-					Alias Editor {helpIconElement}
-				</Modal.Title>
-			</Modal.Header>
+		<div>
+			<div>
+			<div style={{display: 'flex'}}>
+					<h2>Add new indentifiers</h2> 
+					<div>
+						{helpIconElement}
+					</div>
+				</div>
+			</div>
 
-			<Modal.Body>
+			<div>
 				<AliasModalBody languageOptions={languageOptions}/>
-			</Modal.Body>
+			</div>
 
-			<Modal.Footer>
-				<Button variant="primary" onClick={onClose}>Close</Button>
-			</Modal.Footer>
-		</Modal>
+			<div>
+				<Button variant="primary" onClick={onClose}>Done</Button>
+			</div>
+		</div>
 	);
 };
 AliasEditor.displayName = 'AliasEditor';
 AliasEditor.propTypes = {
 	languageOptions: PropTypes.array.isRequired,
 	onClose: PropTypes.func.isRequired,
-	show: PropTypes.bool
-};
-AliasEditor.defaultProps = {
-	show: false
+
 };
 
 function mapDispatchToProps(dispatch) {
 	return {
 		onClose: () => {
-			dispatch(hideAliasEditor());
+			
 			dispatch(removeEmptyAliases());
 		}
 	};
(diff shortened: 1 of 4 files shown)


Tarunmeena0901:
i fixed the "add new identifier" in alias editing section as u can see in screenshots

src/client/entity-editor/alias-editor/alias-editor.js
```diff
@@ -37,15 +35,14 @@ import {faQuestionCircle} from '@fortawesome/free-solid-svg-icons';
...
+ *        add new alias is clicked.
  * @param {boolean} props.show - Whether or not the editor modal should be
  *        visible.
  * @returns {ReactElement} React element containing the rendered AliasEditor.
  */
 const AliasEditor = ({
 	languageOptions,
-	onClose,
-	show
+	onClose
```
MonkeyDo, line 45: The name onClose needs to be changed everywhere since the function has changed.
What does it do now?

src/client/entity-editor/identifier-editor/identifier-editor.js
```diff
@@ -67,37 +63,34 @@ const IdentifierEditor = ({
...
-		<Modal show={show} size="lg" onHide={onClose}>
-			<Modal.Header>
-				<Modal.Title>
-					Identifier Editor {helpIconElement}
-				</Modal.Title>
-			</Modal.Header>
+		<div>
+			<div>
+				<div style={{display: 'flex'}}>
+					<h2>Add new indentifiers</h2>
```
MonkeyDo, line 69:
```suggestion
					<h2>Add identifiers</h2>
```

src/client/entity-editor/identifier-editor/identifier-editor.js
```diff
@@ -67,37 +63,34 @@ const IdentifierEditor = ({
...
 };
 IdentifierEditor.displayName = 'IdentifierEditor';
 IdentifierEditor.propTypes = {
 	identifierTypes: PropTypes.array.isRequired,
-	onClose: PropTypes.func.isRequired,
-	show: PropTypes.bool
-};
-IdentifierEditor.defaultProps = {
-	show: false
+	onClose: PropTypes.func.isRequired
```
MonkeyDo, line 89: Same comment as above, what does this do now?

MonkeyDo (changes requested):
Thanks for opening a PR @Tarunmeena0901 !
Sorry it took so long to get to reviewing it.

Currently, my main feedback looking at the screenshot you posted is:
- each row is taking too much space now, both vertically and horizontally; it made some sense in the modals but in the page we need to conserve some space. For example the giant remove button is unnecessary and can be a small trash icon instead; there is too much padding in between rows; the inputs are quite large, etc.
Maybe the sections do not need to take the whole width of the page; I think both would do better with half of the page width (on desktop).
- the alias section I think would make more sense next to the name section; after all they are alternate names, it seems odd to me to move that section to the bottom of the page. Looking at the name section component, you might need to refactor the layout if you want to move things around or put the two sections side by side.

Tarunmeena0901:
I understand everyone is very busy after vacation so please take your time for reviewing my PR's 😄 
I do wanted to propose a design for this issue but I was waiting for a reply as I didn't wanted to tag any one , 
I'll be back with a design proposal

src/client/entity-editor/alias-editor/alias-editor.js
Tarunmeena0901: ok so this onClose is a dispatch function which is used to remove any empty alias row when user click on Done button (refer SS), 
I thought to change the name to onDone but it doesn't seem very descriptive to me so I thought it may be good to hear some review

MonkeyDo:
Sorry for the delay on reviewing this; I think it deserves a bit more work in terms of design, I think the identifiers section as laid out currently takes a lot of space (I think this was most likely the reason why it was put in a modal before).
It might require a more general reworking of the edit page, and as you mentioned it also needs to take the book wizard page into account.

Are you up for the challenge of proposing a mockup for improving the page design so that this new section fits better?
If not I am happy to work on it but it will most likely take me a couple of weeks to get to.

Tarunmeena0901:
> It might require a more general reworking of the edit page, and as you mentioned it also needs to take the book wizard page into account.

I guess I'll have to sit down with my pencil paper and think this out properly. This week is pretty loaded, but next week won't be. I'll ping you soon...

Tarunmeena0901:
apologies for taking so long for implementing the mockups , there was more work then i expected . the changes in git may seems confusing so i am mentioning some main changes i made
- i removed the button bar component from entity editor page as it was not needed any more but i still deleted it completely the folder and files still exist
- changes may show some files are deleted and added but that's only because of renaming of parent folder, for ex
alias-editor => alias-section
- I added a new redux action and case to add new alias and and new identifiers and removed the addAliasRow/addIdentifierRow from redux
- added a new modal to edit the existing identifiers and aliases
- renamed identifierEditor and aliasEditor variable in backend and test files to identifierSection and aliasSection

Screenshot 2024-04-24 035858

Screenshot 2024-04-24 035843

Screenshot 2024-04-23 230944

MonkeyDo:
Hello, thanks again for working on this, it's not an easy issue!
I have tested the PR locally, and I think there is some room for improvement in terms of UI and UX.
That being said, I did send you off with basically no direction on that subject, so I blame myself. The coding aspects look fine.

I need to sit on it for a bit and do some experimentation to come up with a more detailed mockup of what the pages should look like.
I reckon some of that experimentation I will have to do directly in the code to make it easier, in which case I could possibly directly push my changes if that's OK with you.
Otherwise, If you would prefer to do the changes yourself, I can do my best to give you better mockups and some clear directions for the code if you prefer.
What do you think?

Tarunmeena0901:
If you have specific mockups or ideas in mind, I am more than happy to implement them. However, if you find it easier to experiment directly with the code, please feel free to push your changes to the PR.

Given that the GSoC coding period is ongoing and you might have a lot on your plate, please let me know how you'd prefer to proceed. I'm open to either option.

PR #1044, opened 2023-12-24
fix BB-718:CritiqueBrainz auth endpoint is called server-side

AVtheking:
Problem
https://tickets.metabrainz.org/browse/BB-718

Solution
I have applied a conditional check (typeof window!=="undefined") this check ensures that this function is called from the client side .

Areas of Impact

The changes were reflected in src/client/components/pages/entities/cbReviewModal.tsx

### DIFF
--- a/src/client/components/pages/entities/cbReviewModal.tsx
+++ b/src/client/components/pages/entities/cbReviewModal.tsx
@@ -17,22 +17,21 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-import * as bootstrap from 'react-bootstrap';
+import * as bootstrap from "react-bootstrap";
 // eslint-disable-next-line import/no-internal-modules
-import * as eng from '@cospired/i18n-iso-languages/langs/en.json';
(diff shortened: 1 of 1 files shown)


AVtheking:
please review this

MonkeyDo (commented):
Hello @AVtheking and thanks for submitting a PR !

Unfortunately, a similar PR was submitted prior to yours in #1041 and I will merge that one. Despite them not assigning the ticket to themselves they did open the PR first.

The solution is the same, so good work on that :)

However for future reference, you'll see that there are a lot of unrelated formatting changes committed as well which we don't want.
I assume your editor is set up to automatically format code with Prettier or some other formatter, but we use ESLint to do so (see [the eslint config at the root](https://github.com/metabrainz/bookbrainz-site/blob/master/.eslintrc.js))

MonkeyDo:
Closing in favor of #1041

MonkeyDo:
Actually @AVtheking I was just corrected and pointed out that you had reassigned the ticket to yourself even though it was already assigned to someone. Please don't do that it's bad form and creates confusion.

AVtheking:
Ok sir

PR #1045, opened 2023-12-24
BB-779:URLs with parentheses don’t parse

sidharthkumar2019:
- Updated the urlRegex to match URLs in text.

- Added function with O(n) time complexity to count and remove trailing parantheses.

### Problem

BB-779

### Solution

Updated the urlRegex and also added a function with O(n) time complexity to count and remove trailing parantheses from a URL string.

### Areas of Impact

The changes were reflected in src/client/helpers/utils.tsx

### DIFF
--- a/src/client/helpers/utils.tsx
+++ b/src/client/helpers/utils.tsx
@@ -172,6 +172,28 @@ export function dateObjectToISOString(value: DateObject) {
 }
 
 /**
+ * Find the first index of an unbalanced paranthesis in a url string.
+ * @function firstIndexOfUnbalancedParanthesis
+ * @param {string} url - URL string.
+ * @returns {number} idx - index of the first unbalanced parathesis. If no unbalanced paranthesis found, returns -1
+ */
+function firstIndexOfUnbalancedParanthesis(url: string): number {
+	let cnt = 0;
+	for (let i = 0; i <= url.length; i++) {
+		if (url[i] === '(') {
+			cnt += 1;
+		}
+		else if (url[i] === ')') {
+			cnt -= 1;
+		}
+		if (cnt < 0) {
+			return i;
+		}
+	}
+	return -1;
+}
+
+/**
  * Convert any string url that has a prefix http|https|ftp|ftps to a clickable link
  * and then rendered the HTML string as real HTML.
  * @function stringToHTMLWithLinks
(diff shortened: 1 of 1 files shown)


sidharthkumar2019:
After these changes, the URLs were parsing correctly on my local build. 

Screenshot from 2023-12-25 00-54-11

src/client/helpers/utils.tsx
```diff
@@ -171,6 +171,34 @@ export function dateObjectToISOString(value: DateObject) {
...
+function countTrailingParantheses(url: string): number {
+    let cnt: number = 0;
+    for (const c of url) {
+        if (c === '(') {
+            cnt += 1;
+        } else if (c === ')') {
+            cnt -= 1;
+        }
+    }
+	return Math.abs(cnt);
```
Freso, line 199: I feel like this _probably_ shouldn’t be an absolute number in case it somehow does end up with a negative number? I suspect you’d want to get `-2` returned if it somehow counted two more `)` than `(` instead of `2`, which would suggest two more `(` than `)`?

I don’t know how `String.repeat()` would handle that, but… :shrug:

src/client/helpers/utils.tsx
```diff
@@ -171,6 +171,34 @@ export function dateObjectToISOString(value: DateObject) {
...
 }
 
+/**
+ * Remove trailing parantheses from a url string.
+ * @function removeTrailingParantheses
+ * @param {string} url - URL string.
+ * @returns {string} cleanUrl - sans any trailing parantheses.
+ */
+function removeTrailingParantheses(url: string, count: number): string {
+    return url.slice(0, -1*count);
```
Freso, line 181: How would this deal with something like `https://example.com/wiki/This_is_a_test_(Test))#2`. If I’m reading the code correctly, it would produce `https://example.com/wiki/This_is_a_test_(Test))#` instead of the desired `https://example.com/wiki/This_is_a_test_(Test)`.

Freso (commented):
These seem like fairly functional functions and as such should lend themselves well to unit tests… but no test of any kind included. It would probably be good to add some tests for the cases from the ticket… and maybe a few more to touch on the potential issues in my in‐line comments. :)

src/client/helpers/utils.tsx
Freso: E.g., how would this handle `https://da.wikipedia.org/wiki/Christian_Christensen_(forfatter)))))))))`? (I didn’t try and decipher the revised regex to see if this is ever something that could happen. Would be good with a test either way to ensure it works as expected/desired.)

sriramadari:
why issues section not visible?

MonkeyDo:
> why issues section not visible?

Hello @sriramadari, this isn't really the place to ask a question, but our issue tracker is in Jira: https://tickets.metabrainz.org/browse/BB

sidharthkumar2019:
@Freso Updated my logic. 
Assumption: A valid URL would have balanced parantheses.

If an unbalanced paranthesis is found in a url, the substring before would be valid.

Freso (commented):
Don’t see any obvious logic errors now. :) Would still be good to have tests to make sure the code actually does what’s expected in all the cases (and to prevent regressions), but I don’t know what the BB policy on that is.

sidharthkumar2019:
@Freso Could you please share some documentation for writing tests for this project. I couldn't find any such thing in [BB docs](https://bookbrainz-dev-docs.readthedocs.io/en/latest/index.html)

MonkeyDo (approved):
Thanks @sidharthkumar2019 !

The code looks fine, thanks @Freso for the help !

I am going to merge this as completed, but indeed some unit tests would be welcome as a follow-up PR.
We do not have specific documentation regarding testing, but we use the test framework Mocha with chai.
An intro to Mocha here: https://www.lambdatest.com/learning-hub/mocha-unit-testing

You can have a look at this test file to see what that looks like: https://github.com/metabrainz/bookbrainz-site/blob/master/test/src/server/helpers/utils.js#L19-L27

The idea is to try calling the `stringToHTMLWithLinks` function with specific inputs we want to test, and ensuring the output corresponds to what we are expecting.
In this case you would want to test a 'regular' URL, one with parenthesis, one with extra parenthesis and check that in each case it outputs the expected html string with a correctly formed link.

The tests would be in a new file `test/src/client/helpers/utils.ts` and would take this form with all the use cases you can think of:
```
describe('stringToHTMLWithLinks', () => {
	it('should parse URLs and return an anchor tag', () => {
	...
	});
	it('allows URLs with http and https protocols', () => {
	...
	});
	it('should allow for uneven number of parenthesis', () => {
	...
	});
    ...
});
```
Let me know if you need any help!

PR #1048, opened 2024-01-01
BB-766:Empty identifiers/aliases prevent submission of an entity

insane-22:
### Problem
BB-766 Empty identifiers/aliases prevent submission of an entity

### Solution
 
The "Next" Button gets disabled on the third page due to the formValid check:-
image

The formValid property is set using checks on isCoverTabEmpty and validateCoverTab:-
image

On inspecting isCoverTabEmpty, I found that in order to be able to consider form valid even if no identifiers is present, this line shouldn't be present
image

Also, since Identifiers are not compulsary, its validation should also not be compulsary:-
image

### Areas of Impact

I removed the check on size of Identifiers and changed the validation from compulsary to only when any identifier is present
image


insane-22:
This is my first time contributing to bookbrainz and I am quite new to open source. Tried fixing this bug, hope this helps.
If there are any problems with the solution or implementation, please guide me further.
Thank you for your consideration!

src/client/unified-form/validators/cover-tab.ts
```diff
@@ -40,13 +40,11 @@ export function isCoverTabEmpty(data:any) {
...
 	const authorCreditEditor = get(data, 'authorCreditEditor', {});
 	const ISBN = get(data, 'ISBN', {});
-	const identifierEditor = get(data, 'identifierEditor', {});
 	return nameSection.name?.length === 0 &&
     nameSection.sortName?.length === 0 &&
     !nameSection.language &&
     nameSection.disambiguation?.length === 0 &&
     size(authorCreditEditor) === 1 &&
     !authorCreditEditor.n0?.author &&
-    size(identifierEditor) === 0 &&
```
MonkeyDo, line 50: I don't think this should be removed.
This function checks if the cover tab is empty, which should indeed be false if the user has input an ISBN or any other identifier.

MonkeyDo (changes requested):
Hi @insane-22 thanks for opening a PR!

The root issue is not the one you identified. In fact it is not related to this specific form with he tabs, and you can replicate the same issue for example trying to create a [new work entity](https://test.bookbrainz.org/work/create):

1. fill in the name, sort name and language (which are required) and see that the submit button should be green
2. open the identifier editor and select a type but enter no value. Close the editor
3. see that the submit button is now greyed out

The real issue here is that an identifier with a type selected but no value entered should automatically be discarded and ignored rather than fail validation of the whole form.

I think the easy way to resolve this at this point in time would be to remove identifiers with a type selected but no value. We currently already remove entirely empty rows here, so it's probably as simple as removing ` && identifier.get('type') === null)` here, then finding a similar solution for the aliases editor:
https://github.com/metabrainz/bookbrainz-site/blob/c47025006298d28a7720e779f75bf6d2fcfc0253/src/client/entity-editor/identifier-editor/reducer.ts#L59-L61



Do let me know if you need any assistance.

insane-22:
Ahh, I realised i misunderstood the isssue completely and hence the solution too is incorrect (infact not of the required problem). Allow me sometime to look into this and see if I could help with my contribution.
Thanks a lot for your invaluable guidance!

insane-22:
> The real issue here is that an identifier with a type selected but no value entered should automatically be discarded and ignored rather than fail validation of the whole form.
 
I guess rather than removing  `&& identifier.get('type') === null)` , it should be `||  identifier.get('type') === null)` so that the row is considered empty when either the value or type of identier is not provided, what do you think?

MonkeyDo:
> > The real issue here is that an identifier with a type selected but no value entered should automatically be discarded and ignored rather than fail validation of the whole form.
> 
> I guess rather than removing `&& identifier.get('type') === null)` , it should be `|| identifier.get('type') === null)` so that the row is considered empty when either the value or type of identifier is not provided, what do you think?

No, I think we probably want to keep rows where the user tried to enter some information but might be incomplete (i.e. tried to paste a link, but auto-detection didn't work or something like that).
My reasoning being that we don't want to frustrate our users by deleting information they entered. I think the type selection in a drop-down does not qualify as information we want to preserve, so that check can just be removed.

insane-22:
> > > The real issue here is that an identifier with a type selected but no value entered should automatically be discarded and ignored rather than fail validation of the whole form.
> > 
> > 
> > I guess rather than removing `&& identifier.get('type') === null)` , it should be `|| identifier.get('type') === null)` so that the row is considered empty when either the value or type of identifier is not provided, what do you think?
> 
> No, I think we probably want to keep rows where the user tried to enter some information but might be incomplete (i.e. tried to paste a link, but auto-detection didn't work or something like that). My reasoning being that we don't want to frustrate our users by deleting information they entered. I think the type selection in a drop-down does not qualify as information we want to preserve, so that check can just be removed.

Ok, got it!
I do agree removing name may cause a bad user experience. Hence, row should be considered empty only when "name" is not provided. Extending similar logic to alias editor, it is updated to be considered empty when name is not provided.

insane-22:
I have made the changes, kindly review them and suggest changes if any.

insane-22:
Preview:- 

https://github.com/metabrainz/bookbrainz-site/assets/116253210/310d9f7d-802d-40bd-bd8f-087fe6cd8d49

MonkeyDo (approved):
Great, thank you ! 🚀 

Sorry for the delay in finishing reviewing the PR !

PR #1049, opened 2024-01-03
fix BB-758 : Link from edit page to entity page

Tarunmeena0901:
### Problem
 BB-758

### Solution

Using anchor tag I added a link inside the heading of entity in editor page , see the screenshot . A direct link to entity page from entity editor page makes it very  convenient while comparing the fields being edited with the original ones 

ent

https://github.com/metabrainz/bookbrainz-site/assets/97682967/e4d297f3-7e27-40a2-9c79-66d57584226f

### Areas of Impact

This PR effect the entity-editor page section of the codebase

### DIFF
--- a/src/client/entity-editor/entity-editor.tsx
+++ b/src/client/entity-editor/entity-editor.tsx
@@ -73,6 +73,7 @@ const EntityEditor = (props: Props) => {
 		entity
 	} = props;
 	const currentState = (useSelector((state) => state) as any).toJS();
+	let entityURL;
 	// eslint-disable-next-line consistent-return
 	const handleUrlChange = React.useCallback(() => {
 		if (!_.isEqual(currentState, props.intitialState) && !currentState.submissionSection.submitted) {
@@ -82,8 +83,11 @@ const EntityEditor = (props: Props) => {
 	React.useEffect(() => {
 		window.onbeforeunload = handleUrlChange;
 	}, [handleUrlChange]);
-	const entityURL = getEntityUrl(entity);
 
+	if(entity){
+	entityURL = getEntityUrl(entity);
+	}
+	
 	return (
 		<form onSubmit={onSubmit}>
 			<Card>
--- a/src/client/entity-editor/entity-editor.tsx
+++ b/src/client/entity-editor/entity-editor.tsx
@@ -27,6 +27,7 @@ import NameSection from './name-section/name-section';
 import RelationshipSection from './relationship-editor/relationship-section';
 import SubmissionSection from './submission-section/submission-section';
 import _ from 'lodash';
+import {getEntityUrl} from '../helpers/entity';
 import {submit} from './submission-section/actions';
 
 
@@ -34,6 +35,7 @@ type OwnProps = {
 	children: React.ReactElement<any>,
 	heading: string,
 	intitialState:Record<string, any>,
+	entity: any
 };
 
 type StateProps = {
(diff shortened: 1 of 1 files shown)


src/client/entity-editor/entity-editor.tsx
```diff
@@ -80,11 +84,17 @@ const EntityEditor = (props: Props) => {
...
+	}
+	
 	return (
 		<form onSubmit={onSubmit}>
 			<Card>
 				<Card.Header as="h4">
-					{heading}
+					<div>
+					 <a href={entityURL}>{heading}</a>
+					</div>
```
MonkeyDo, line 97: Following my comment above, in the case that entity is falsy we know it's not an edit but a creation, in which case we can use the `heading` directly instead of creating a broken link.
`{entityURL ? <a href={entityURL}>{heading}</a> : heading}`

src/client/entity-editor/entity-editor.tsx
```diff
@@ -27,13 +27,15 @@ import NameSection from './name-section/name-section';
...
 import _ from 'lodash';
+import {getEntityUrl} from '../helpers/entity';
 import {submit} from './submission-section/actions';
 
 
 type OwnProps = {
 	children: React.ReactElement<any>,
 	heading: string,
 	intitialState:Record<string, any>,
+	entity: any
```
MonkeyDo, line 38: I don't know where we get this prop from, but it seems to be working ❓

MonkeyDo (changes requested):
Thanks for opening this PR !

Looking pretty good. A suggestion for improvement:

src/client/entity-editor/entity-editor.tsx
Tarunmeena0901: oh nice , I think that will be much easier to understand for others too !

Tarunmeena0901: yup I think this is much better

src/client/entity-editor/entity-editor.tsx
Tarunmeena0901: so the idea of using this prop came to me when I looked at author.js which use getEntityUrl( ) to etract the Url prefix for a edit button which redirect us to edit page 

https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/author.js#L197

honestly , when I started working on this it was quite difficult to understand how data is being passed to pages in our website when they loads , but after some tracing I figure out that the most probably the entity prop is most probably coming from here
https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/helpers/props.js#L56

Tarunmeena0901:
Thank you for reviewing my PR , I have made the suggested changes please have a look 😃

src/client/entity-editor/entity-editor.tsx
Tarunmeena0901: sorry for my confusion as i am still understanding the BB codebase .... so later i found out that entity editor component is getting its props from `src/client/entity-editor/controller.js` file from this part [https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/entity-editor/controller.js/#L49-L51](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/entity-editor/controller.js/#L49-L51)

I hope this time I am right but pls do correct if I am wrong here 😅

MonkeyDo (approved):
Thank you for this improvement!

PR #1051, opened 2024-01-10
Fix BB-768 : Unset Author Credit leads to bugs

Tarunmeena0901:
### Problem

[BB-768](https://tickets.metabrainz.org/browse/BB-768)

refer to this [discussion](https://community.metabrainz.org/t/problems-with-the-this-edition-doesnt-have-an-author-tag/659459) for more details 

### Solution

- In `edition.js` and `edition-group.js` I added a check before displaying a warning of unset credit . this will not show any warning on created edition or edition-group page, if we click the checkbox while creating the edition or edition group 

- The checkbox was not getting checked and was needed to be clicked two times before disabling the edition credit , the reason was the redux initial state of `authorCreditEnable` was undefined rather than being true .

### Areas of Impact
- ./entities/edition.js
- ./entities/edition-group.js
- redux reducers.ts for edition and edition-groups
- actions.ts in author-credit-editor


src/client/entity-editor/author-credit-editor/author-credit-section.tsx
```diff
@@ -91,7 +96,7 @@ function AuthorCreditSection({
 	const authorCreditPreview = _map(authorCreditEditor, (credit) => `${credit.name}${credit.joinPhrase}`).join('');
 	const authorCreditRows = _values(authorCreditEditor);
 
-	const isValid = validateAuthorCreditSection(authorCreditRows, authorCreditEnable);
```
Tarunmeena0901, line 94: is this check necessary do this check even if author credit are unabled at first place ? 🤔

Tarunmeena0901:
well..... I was just reviewing my own PR today and find out that my solution for this problem was so naive 😆 , so I had another look at this issue and fixed it in a better way in my upcoming commit

Tarunmeena0901:
so there was two cases when authorCreditEnable was being undefined whenever the page loads
1.  when we try to create an edition
2. when we try to edit an edition group

the reason behind this in first case was that the rootstate had an empty edition section with no authorCreditEnable field to set the default state with 

editionSEc

and the reason behind the second case was we were not sending any default authorCreditEnable field in enditionGroupSection rootstate to set the default state from 

groupSec


after fix for case1 :

Screenshot 2024-02-01 155737

same the bug is fixed for edition-group edit routes

src/client/components/pages/entities/edition-group.js
```diff
@@ -79,7 +79,7 @@ EditionGroupAttributes.propTypes = {
 };
 
 
-function EditionGroupDisplayPage({entity, identifierTypes, user, wikipediaExtract}) {
+function EditionGroupDisplayPage({entity, identifierTypes, user, wikipediaExtract}, authorCreditEnable) {
```
kellnerd, line 82: This parameter and its usage below are probably a leftover from your previous attempt to solve the issue. Modern React components have no second function parameter as far as I know.
Same goes for the `EditionDisplayPage` component.

MonkeyDo:
Thanks for looking into this issue again, I appreciate your efforts!
I will review this very soon, thanks for your continued patience.

src/client/components/pages/entities/edition-group.js
Tarunmeena0901: YES ! you are correct, Kell. React components do not have a second function parameter. Thank you for pointing this out. This is indeed something I overlooked in my previous solution, where I attempted to prevent the warning about unset author credits from displaying on the edition and edition group pages. However, it's clear now that this approach won't work

Tarunmeena0901: but now as I think that when we are creating a edition or an edition group we don't allow to submit the form until the author credit is filled or "No author credit" checkbox is checked

so we have two cases here
1. when someone as set that there is no author credit 

2. author credits are provided while creating

in both the cases after creating the edition or edition group , the main page should not show any warning then why do we even have the warning ??? 
maybe we should remove it completely

MonkeyDo:
Thanks again for working on this.
I realized there is a crucial piece missing in this puzzle, and the feature won't work properly until it is resolved.

In short, I think we do not currently have a a way to store in the database that an edition has been set as "no author credit". The web page works fine, but we don't have a has_author_credit boolean column in the edition_data table, which is required if we want the checkbox to actually do something and being able to hide the warning.

Some modifications need to be done on the bookbrainz-site repo, and others on the bookbrainz-data-js repo (our ORM).
I'll get back to you when I'm able to work on this, hopefully next week.

Tarunmeena0901:
> Thanks again for working on this. I realized there is a crucial piece missing in this puzzle, and the feature won't work properly until it is resolved.
> 
> In short, I think we do not currently have a a way to store in the database that an edition has been set as "no author credit". The web page works fine, but we don't have a has_author_credit boolean column in the edition_data table, which is required if we want the checkbox to actually do something and being able to hide the warning.
> 
> Some modifications need to be done on the bookbrainz-site repo, and others on the bookbrainz-data-js repo (our ORM). I'll get back to you when I'm able to work on this, hopefully next week.


Yesss, exactly! Currently, I was trying to do just that—saving a boolean variable in the database to indicate whether author credits are set or unset for an edition or edition group. I made some schema changes to the edition_data table and edition_group_data table yesterday. I would have completed it today, but I got a little busy. I'll surely figure out the next part; give me some time, and I'll be back with a fix. ✅✅

src/client/components/pages/entities/edition-group.js
Tarunmeena0901: Oh i see now

Tarunmeena0901:
In my last commit, I added a credit_section boolean column to our edition_data and edition_group_data tables. I also added the same column to our views.

The purpose is to set credit_section to false if author credits are disabled. This signifies that the entity does not require author credits, thus no warning will be shown. If credit_section for an entity is true, along with other pre-existing checks, it means that the author_credits section exists but with empty author credits, and a warning will be displayed.

Screenshot 2024-02-08 014248


The next step is to incorporate the credit_section state in the frontend. Until then, I welcome any suggestions for improvement or a better approach to solving this aspect of the issue. 😊

Tarunmeena0901:
My apologies for finalizing this PR later than expected. I was occupied with my college responsibilities. With this commit, consider this issue resolved.

In my latest commit, I utilized the saved state of author credits in both the edition and edition_group in the frontend to determine whether to display the "Unset author credit warning." Additionally, the "Author credit enable" checkbox in the edit page is no longer set to true by default; instead, it utilizes the previously saved state of the credit section to determine whether it should be checked or unchecked.

this commit fix BB-768 as well as fix BB-767 . The only remaining task is to address the failing test and make changes in bookbrainz-data-js . 
Thank you for your patience.

Tarunmeena0901:
Can I get a little help here, please, regarding the failing build-test? It's been a while, and I'm trying to resolve this, but I can't get anywhere.

Tarunmeena0901:
i was having a look at it again and after all this time i noticed that `unified-form` is crying in the corner and i haven't made any changes there , so i did it this time and  i noticed two view in `bookbrainz.sql` file which were not present in my local postgres setup
whenever i tried to do `\dv`  , so i am not sure if those needs the the new `credit_section` field.

in `edition` and` edition-group` pages rather than displaying nothing if the author credits are not provide while creating the entity i thought it would be better to display something like `Author Credits: N/A` so if someone see it and remember the credits they may edit the entity accordingly

MonkeyDo:
> it means that the author_credits section exists but with empty author credits

While re-reading this comment, a lightbulb turned on above my head:
I suppose all we really need is to modify the entity creation code to allow creating an empty author_credit (i.e. an AC without any author_credit_name attached to it) when the "empty author credit" checkbox is checked.

Similarly, when loading an existing entity from the database, if it has an author_credit but the author_credit has no author_credit_name attached to it we can consider it's on purpose, and we don't show the "author credit unset" warning. And if we try editing that entity, the "no author credit" checkbox will be pre-checked.

There is also an author_count (INT) column in the author_credit table, which we don't currently use anywhere as far as I can tell. Perhaps manually setting it to 0 could serve the same purpose? Or maybe as a secondary check that the entity has "no AC" set on purpose? Not sure it's necessary.

Tarunmeena0901:
instead of empty string can i put my name inside author name property please,  " " => "Tarun" ? 
hehe just kidding but it will be cool to have my name in the codebase

Tarunmeena0901:
> While re-reading this comment, a lightbulb turned on above my head: I suppose all we really need is to modify the entity creation code to allow creating an empty author_credit (i.e. an AC without any author_credit_name attached to it) when the "empty author credit" checkbox is checked.

I should have asked this question earlier, but since I've started working on this now, I'm not able to comprehend that by using this method, how do we handle entities that are already present . Since they didn't have empty author credits, are we gonna edit each entity and check the 'don't have author credit' checkbox? 
This doesn't seem feasible as we have over 2000 editions and edition groups.

MonkeyDo:
One thing I'll need to figure out on the heels of this fix is a good way to allow users to keep an Edition and its Edition Group in sync. In MusicBrainz it's a simple checkbox which only appears when the author credit has been modified:

image

Tarunmeena0901:
I have the changes but I am not able to figure what is causing the build error for now

MonkeyDo:
I had some issues with the automated deployment, sorry about that !
Taking the opportunity to check the new package so I pushed the commit.

Tarunmeena0901:
ohh ok i just updated it locally but for some reason in package.json it still shows 5.0.0

MonkeyDo:
> ohh ok i just updated it locally but for some reason in package.json it still shows 5.0.0

It just wasn't published until a minute ago :sweat_smile:  Now version 5.1.1 should be installed

Tarunmeena0901:
oh ok ok 👍 i will wait some more before updating

MonkeyDo:
> oh ok ok 👍 i will wait some more before updating

The poackage is already updated, I pushed a commit for it.

However tests are still failing, but looks unrelated to the changes in your PR. Let me look into it

Tarunmeena0901:
Oh, I didn't notice that you updated the package. Thanks!!!
I also don't think the failing test is related to this PR, but let me know if it is.

MonkeyDo:
Hello!

I updated the PR and got rid of all the unrelated issues, phew!
I deployed the PR for testing on test.bookbrainz.org, and found that there is a remaining issue that is most likely related to the failing test (https://github.com/metabrainz/bookbrainz-site/runs/25781165934#r0s38)

Creating an Edition without an AC works absolutely fine, and when on the edition display page the "no AC set" warning is not shown, so that's working :100: ! See https://test.bookbrainz.org/revision/30808

However if I then edit the edition to add an AC, it looks like the AC is not saved to the database, and the "no AC set" warning is shown, which suggests the tickbox state ("this edition doesn't have an AC") is saved correctly, but not the AC itself.
See https://test.bookbrainz.org/revision/30809


Want to investigate what is happening in the noodly code of entity saving? :rofl: 
I'll go hunting too when I can but two heads are better than one !

Tarunmeena0901:
yeah sure ,  I will have a look 🫡

Tarunmeena0901:
finally my elastic search problem is fixed and i can work on the new issue we have here let me know if you want me to start a fresh pr this one has become little MESSY.

MonkeyDo:
> finally my elastic search problem is fixed and i can work on the new issue we have here let me know if you want me to start a fresh pr this one has become little MESSY.

Glad you got your local setup working :)
No need to open another PR, we can work with this one. I might end up squashing the commits so that the history in the main branch is clean.

MonkeyDo:
I found the culprit!
In processAuthorCredit we were checking if authorCreditEnabled but using the old state of the entity rather than the new one the client just sent, so if you created an edition with no AC and then tried adding an AC, authorCreditEnabled would be false and it would not create an AC.

Tarunmeena0901:
THANKS!!!! You just saved my weekend. I was going to look at it tomorrow, and this may have taken me hours to figure out the problem. Thanks again! 🙌 😇

I also noticed that you replaced some promises with async/await syntax. It looks great!✨ But here’s a twist,  I had already made those changes in the old PR for BB-672 #1058. Déjà vu, developer style! 😂

 Look, they are almost similar
 https://github.com/metabrainz/bookbrainz-site/pull/1058/files#diff-69e04ad0b67c34fe4770d91cc33d661af834447e45415e53ba9cf48023eb6537L739-R819

MonkeyDo:
Ahhaaa yes, I thought it seemed familiar!
Sorry I haven't got that one merged yet! Maybe I should do that other one first?

MonkeyDo:
P.S.: I'm rebuilding the image and deploying it to test.bookbrainz if you want to do some manual testing to check everything.

Tarunmeena0901:
> P.S.: I'm rebuilding the image and deploying it to test.bookbrainz if you want to do some manual testing to check everything.


sure ! I will check




> Ahhaaa yes, I thought it seemed familiar!
> Sorry I haven't got that one merged yet! Maybe I should do that other one first?

no problem , I will just merge the conflicting changes there once this PR is done

MonkeyDo (approved):

MonkeyDo:
OK, PR is ready to merge as far as I can tell !
I'm going to deploy it to beta for a week or so for further testing with a wider pool of users.

PR #1055, opened 2024-01-21
BB-765: Identifiers editor: each value should show a generated link from the cleaned up value

insane-22:
### Problem

BB-765: Identifiers editor: each value should show a generated link from the cleaned up value

### Solution

The component IdentifierLink present in "src/components/pages/entities/identifiers-links.js" creates a link for the identifier present.
image

I have reused this component tocreate a link whenever user adds value and type both and displayed it in the next row. Final component looks like this:-
image

### Areas of Impact

Changes have been done in: "src/client/entity-editor/identifier-editor/identifier-row.tsx"


insane-22:
Preview:-
https://github.com/metabrainz/bookbrainz-site/assets/116253210/4410a716-8b5c-4e73-a468-8454d57dd954

src/client/entity-editor/identifier-editor/identifier-row.tsx
```diff
@@ -128,6 +129,16 @@ function IdentifierRow({
...
 				</Col>
 			</Row>
+			{typeValue && valueValue && (
+				<Row>
+				<Col lg={2}>
+					Preview Link:-
+				</Col>
+				<Col>
+				<IdentifierLink typeId={typeValue} value={valueValue}/>
+				</Col>
```
MonkeyDo, line 139: I don't think the two columns are required here, it creates quite a big  space before the link on desktop.
I suggest this:
```suggestion
				<Col>
					Preview link:
					<IdentifierLink typeId={typeValue} value={valueValue}/>
				</Col>
```

Note I also removed the dash after "Preview link:"


The comparison:
image

MonkeyDo (changes requested):
Thanks for this good improvement !

I have a small suggestion regarding the layout, but other than that 🚀

insane-22:
@MonkeyDo , thank you for your suggestions. I have encorporated them in the next commit, please have a look

src/client/entity-editor/identifier-editor/identifier-row.tsx
```diff
@@ -185,8 +194,10 @@ function mapDispatchToProps(
 		onRemoveButtonClick: () => dispatch(removeIdentifierRow(index)),
 		onTypeChange: (value: {value: number}) =>
 			dispatch(updateIdentifierType(index, value && value.value)),
-		onValueChange: (event: React.ChangeEvent<HTMLInputElement>) =>
-			handleValueChange(dispatch, event, index, typeOptions)
+		onValueChange: (event: React.ChangeEvent<HTMLInputElement>) =>{
+			const { cleanedValue, action } = handleValueChange(event,index,typeOptions);
```
MonkeyDo, line 198: I'm looking at this again, and I see we don't do anything with `cleanedValue`.
Are these changes to `handleValueChange` and `mapStateToProps` required at all?

insane-22: Oh yes, I see the cleanedValue is actually doing nothing. The reason I had used it earlier was that I was planning to pass it as a prop and then pass it further to the "IdentifierLink" but later realised that this is not required, things being defined the way they are, 'typevalue' is already modified to have the cleaned upvalue.

MonkeyDo: Hi @insane-22, Do you have a minute to finish this PR off?
I think apart from the handleValueChange and mapStateToProps changes it should be ready to merge.

insane-22: yes yes @MonkeyDo , I sincerely apologise for this delay. It should have been done by now. I would be doing the changes now.

insane-22:
Done with changes and rechecked. Things are working fine. 
Sorry again for this delay, I would ensure this never happens again. 
And thanks a lot for your patience!

MonkeyDo (approved):
Thank you! :+1:

PR #980, opened 2023-03-28
BB-672: Rewrite entity folder using async/await

GandalfGwaihir:
### Problem

[BB-672](https://tickets.metabrainz.org/browse/BB-672):
Across the codebase we have asynchronous code written using promises, which is hard to read and to maintain and error-prone.

### Solution

All this code was rewritten using the async/await syntax.

### Areas of Impact

The changes were reflected in `src/server/routes/entity`

PR #1058, opened 2024-01-30
fix BB-672: rewriting promises using async await in .routes/entity folder

Tarunmeena0901:
### Problem
BB-672

### Solution

- rewrite the promises using aync await syntax in entity routes folder
- used try catch block rather than .catch() to handle error

This pull request finishes the remaining tasks from [PR-980](https://github.com/metabrainz/bookbrainz-site/pull/980). I began everything again to better understand the solution for myself.
### Areas of Impact

`src/server/routes/entity`

### DIFF
--- a/src/server/routes/entity/edition.ts
+++ b/src/server/routes/entity/edition.ts
@@ -371,9 +371,9 @@ function _setEditionTitle(res) {
 }
 
 router.get('/:bbid', middleware.loadEntityRelationships, middleware.loadWorkTableAuthors,
-	middleware.loadWikipediaExtract, (req, res, next) => {
+	middleware.loadWikipediaExtract, (req, res) => {
 		_setEditionTitle(res);
-		entityRoutes.displayEntity(req, res, next);
+		entityRoutes.displayEntity(req, res);
 	});
 
 router.get('/:bbid/revisions', (req, res, next) => {
--- a/src/server/routes/entity/edition.ts
+++ b/src/server/routes/entity/edition.ts
@@ -158,37 +158,33 @@ router.get(
 	'/create', auth.isAuthenticated, auth.isAuthorized(ENTITY_EDITOR), middleware.loadIdentifierTypes,
 	middleware.loadEditionStatuses, middleware.loadEditionFormats,
 	middleware.loadLanguages, middleware.loadRelationshipTypes,
-	(req:PassportRequest, res, next) => {
+	async (req:PassportRequest, res, next) => {
 		const {EditionGroup, Publisher, Work, Author} = req.app.locals.orm;
-		const propsPromise = generateEntityProps(
+		const propsObject = generateEntityProps(
 			'edition', req, res, {}
 		);
+		async function fetchData(entity, key) {
+			try {
+				const data = await entity.forge({bbid: req.query[key]})
+					.fetch({require: false, withRelated: 'defaultAlias'});
+				return utils.entityToOption(data.toJSON());
+			}
+			catch (err) {
+				log.debug(err);
+				return next(err);
+			}
+		}
 		if (req.query.author) {
-			propsPromise.author = Author.forge({bbid: req.query.author})
-				.fetch({require: false, withRelated: 'defaultAlias'})
-				.then((data) => data && utils.entityToOption(data.toJSON()));
+			propsObject.author = await fetchData(Author, 'author');
 		}
-
-		// Access edition-group property: can't write req.query.edition-group as the dash makes it invalid Javascript
 		if (req.query['edition-group']) {
-			propsPromise.editionGroup =
-				EditionGroup.forge({bbid: req.query['edition-group']})
-					.fetch({require: false, withRelated: 'defaultAlias'})
-					.then((data) => data && utils.entityToOption(data.toJSON()));
+			propsObject.editionGroup = await fetchData(EditionGroup, 'edition-group');
 		}
-
 		if (req.query.publisher) {
-			propsPromise.publisher =
-				Publisher.forge({bbid: req.query.publisher})
-					.fetch({require: false, withRelated: 'defaultAlias'})
-					.then((data) => data && utils.entityToOption(data.toJSON()));
+			propsObject.publisher = await fetchData(Publisher, 'publisher');
 		}
-
 		if (req.query.work) {
-			propsPromise.work =
-				Work.forge({bbid: req.query.work})
-					.fetch({require: false, withRelated: 'defaultAlias'})
-					.then((data) => data && utils.entityToOption(data.toJSON()));
+			propsObject.work = await fetchData(Work, 'work');
 		}
 
 		async function render(props) {
@@ -275,9 +271,14 @@ router.get(
 			}));
 		}
 
-		makePromiseFromObject(propsPromise)
-			.then(render)
-			.catch(next);
+		try {
+			const props = await makePromiseFromObject(propsObject);
+			return render(props);
+		}
+		catch (err) {
+			log.debug(err);
+			return next(err);
+		}
 	}
 );
 
@@ -310,7 +311,7 @@ router.post(
 				entity.editionSection.publisher = foundOption ? _.omit(foundOption, ['disambiguation']) : null;
 			}
 		}
-		const propsPromise = generateEntityProps(
+		const propsObject = generateEntityProps(
 			'edition', req, res, {}, () => entity
 		);
 		function render(props) {
@@ -325,9 +326,14 @@ router.post(
 			}));
 		}
 
-		makePromiseFromObject(propsPromise)
-			.then(render)
-			.catch(next);
+		try {
+			const props = await makePromiseFromObject(propsObject);
+			return render(props);
+		}
+		catch (err) {
+			log.debug(err);
+			return next(err);
+		}
 	}
 );
 
@@ -365,9 +371,9 @@ function _setEditionTitle(res) {
 }
 
 router.get('/:bbid', middleware.loadEntityRelationships, middleware.loadWorkTableAuthors,
-	middleware.loadWikipediaExtract, (req, res) => {
+	middleware.loadWikipediaExtract, (req, res, next) => {
 		_setEditionTitle(res);
-		entityRoutes.displayEntity(req, res);
+		entityRoutes.displayEntity(req, res, next);
 	});
 
 router.get('/:bbid/revisions', (req, res, next) => {
--- a/src/server/routes/entity/entity.tsx
+++ b/src/server/routes/entity/entity.tsx
@@ -67,7 +67,7 @@ const entityComponents = {
 	work: WorkPage
 };
 
-export function displayEntity(req: PassportRequest, res: $Response) {
+export async function displayEntity(req: PassportRequest, res: $Response, next: NextFunction) {
 	const {orm}: {orm?: any} = req.app.locals;
 	const {AchievementUnlock, EditorEntityVisits} = orm;
 	const {locals: resLocals}: {locals: any} = res;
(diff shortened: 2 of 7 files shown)


src/server/routes/entity/edition.ts
```diff
@@ -158,37 +158,33 @@ router.get(
...
 	middleware.loadEditionStatuses, middleware.loadEditionFormats,
 	middleware.loadLanguages, middleware.loadRelationshipTypes,
-	(req:PassportRequest, res, next) => {
+	async (req:PassportRequest, res, next) => {
 		const {EditionGroup, Publisher, Work, Author} = req.app.locals.orm;
-		const propsPromise = generateEntityProps(
+		const propsObject = generateEntityProps(
 			'edition', req, res, {}
 		);
+		async function fetchData(entity, key) {
```
MonkeyDo, line 166: Refactoring is good !

We do have a utility function in the ORM that does the same thing while also making sure to follow merged BBIDs:
https://github.com/metabrainz/bookbrainz-data-js/blob/master/src/func/entity.ts#L199-L212

So you should be able to use `await orm.func.entity.getEntity(orm, 'Work', bbid, ['defaultAlias']);` directly instead of rolling out this fetchData utility.

There is a small difference in that the fetch option `require` is set to true instead of false, but I think in this context it does make sense to ensure the entity we're trying to link does exist.

I think you might also have to move the try…catch block around all these `if (req.query.xyz)` so that error are caught (especially with `require:true` as stated above)

src/server/routes/entity/edition.ts
```diff
@@ -275,9 +271,14 @@ router.get(
 			}));
 		}
 
-		makePromiseFromObject(propsPromise)
-			.then(render)
-			.catch(next);
+		try {
+			const props = await makePromiseFromObject(propsObject);
```
MonkeyDo, line 275: I think `makePromiseFromObject` is not needed anymore here, as it was used to await all the promises added to propsPromise.
Since we are now directly awaiting each value I think the utility doesn't do anything.

src/server/routes/entity/entity.tsx
```diff
@@ -84,67 +84,68 @@ export function displayEntity(req: PassportRequest, res: $Response) {
...
+
+			editorEntityVisitPromise = await achievement.processPageVisit(orm, resLocals.user.id);
+		}
+		catch {
+			// error caused by duplicates we do not want in database
+			return false;
+		}
 	}
 	else {
 		editorEntityVisitPromise = new Promise(resolve => resolve(false));
```
MonkeyDo, line 102:
```suggestion
		editorEntityVisitPromise = false;
```

src/server/routes/entity/entity.tsx
```diff
@@ -84,67 +84,68 @@ export function displayEntity(req: PassportRequest, res: $Response) {
...
+				bbid: resLocals.entity.bbid,
+				editorId: resLocals.user.id
+			})
+				.save(null, {method: 'insert'});
+
+			editorEntityVisitPromise = await achievement.processPageVisit(orm, resLocals.user.id);
+		}
+		catch {
+			// error caused by duplicates we do not want in database
+			return false;
```
MonkeyDo, line 98: I don't think we want to `return` here; the original code ended up setting editorEntityVisitPromise to a value, so in this case it should just be:
```suggestion
			editorEntityVisitPromise = false;
```

src/server/routes/entity/entity.tsx
```diff
@@ -84,67 +84,68 @@ export function displayEntity(req: PassportRequest, res: $Response) {
 
 	let editorEntityVisitPromise;
```
MonkeyDo, line 85: We need to rename all these xyzPromise variables; since we are not assigning a promise to them but instead `await`ing a promise and assigning the returned value directly, they should be
`editorEntityVisitPromise` -> `editorEntityVisit`
`alertPromise` -> `alert` 
etc.

src/server/routes/entity/entity.tsx
```diff
@@ -84,67 +84,68 @@ export function displayEntity(req: PassportRequest, res: $Response) {
...
-			);
-			alertPromise = Promise.all(promiseList);
-		}
-		else {
-			alertPromise = new Promise(resolve => resolve(false));
-		}
-		return alertPromise;
-	});
+	let alertPromise;
+	const visitAlert = await editorEntityVisitPromise;
```
MonkeyDo, line 106: This is why we're rewriting this logic: the promise chaining gets really confusing ! :D

The await here doesn't make sense. `editorEntityVisitPromise` is not a promise.
You've already `await`ed a promise and set the resulting value as `editorEntityVisitPromise`.
So in this context,  the value `visitAlert` from the old code is the same as `editorEntityVisitPromise`  in the current new code (after my suggestion to above to rename it, `editorEntityVisit`)

src/server/routes/entity/entity.tsx
```diff
@@ -84,67 +84,68 @@ export function displayEntity(req: PassportRequest, res: $Response) {
...
+		});
+
+		alertPromise = Promise.all(promiseList);
+	}
+	else {
+		alertPromise = new Promise(resolve => resolve(false));
+	}
+
+	return (async () => {
+		const alert = await alertPromise;
```
MonkeyDo, line 148: In this particular case, `alertPromise` is (almost) well named since we do store `Promise.all` or `new Promise` *without awaiting them*.

I would still rename it to `alertPromises` since the main reason we have it is so we cant `await`  multiple promises with `Promise.all`.

src/server/routes/entity/entity.tsx
```diff
@@ -84,67 +84,68 @@ export function displayEntity(req: PassportRequest, res: $Response) {
...
+			}
+		});
+
+		alertPromise = Promise.all(promiseList);
+	}
+	else {
+		alertPromise = new Promise(resolve => resolve(false));
+	}
+
+	return (async () => {
```
MonkeyDo, line 147: I don't think you need to `return` anything anymore.
Previously we were returning a promise chain, but an `async` function always returns a Promise so that's already taken care of.

src/server/routes/entity/entity.tsx
```diff
@@ -264,14 +265,15 @@ function _createNote(orm, content, editorID, revision, transacting) {
...
+export async function addNoteToRevision(req: PassportRequest, res: $Response) {
 	const {orm}: {orm?: any} = req.app.locals;
 	const {Revision, bookshelf} = orm;
 	const editorJSON = req.session.passport.user;
 	const revision = Revision.forge({id: req.params.id});
 	const {body}: {body: any} = req;
-	const revisionNotePromise = bookshelf.transaction(
-		(transacting) => _createNote(
+
+	const revisionNotePromise = await bookshelf.transaction(
```
MonkeyDo, line 275: (Same comment as above to rename this variable to `revisionNote`)

We don't want to use `sendPromiseResult` here as it is mainly a utility for awaiting a promise safely and sending the result.
But since you're already `await`ing the promise here, you should wrap it in a try catch block and directly return a response like so:
`return res.status(200).send(revisionNote);`

In the catch block you can `return next(err);`

src/server/routes/entity/entity.tsx
```diff
@@ -1201,11 +1203,21 @@ export function handleCreateOrEditEntity(
...
+			req.session,
+			entityType,
+			orm,
+			editorJSON,
+			derivedProps,
+			isMergeOperation,
+			transacting
+		);
+	});
+	const entityJSON = await entityEditPromise;
```
MonkeyDo, line 1219: This does work, but why not the simpler `const entityJSON = await bookshelf.transaction(…` ?

src/server/routes/entity/entity.tsx
```diff
@@ -1201,11 +1203,21 @@ export function handleCreateOrEditEntity(
...
+	const achievementPromise = await processAchievement(orm, editorJSON.id, entityJSON);
 	return handler.sendPromiseResult(
```
MonkeyDo, line 1221: Like above, we don't want to use handler.sendPromiseResult here but instead roll out our own try…catch block and return a response directly.

src/server/routes/entity/publisher.ts
```diff
@@ -207,18 +207,18 @@ router.get('/:bbid', middleware.loadEntityRelationships, middleware.loadWikipedi
...
-			_setPublisherTitle(res);
-			res.locals.entity.editions.sort(entityRoutes.compareEntitiesByDate);
-			return entityRoutes.displayEntity(req, res);
-		})
-		.catch(next);
+	try {
+		const editions = await Publisher.forge({bbid: res.locals.entity.bbid})
+			.editions({withRelated: editionRelationsToFetch});
+		res.locals.entity.editions = editions.toJSON();
+		res.locals.entity.editions.sort(entityRoutes.compareEntitiesByDate);
```
MonkeyDo, line 215: `_setPublisherTitle` seems to be gone. I'm not sure exactly how useful it was, but it should probably go back in.

src/server/routes/entity/publisher.ts
```diff
@@ -207,18 +207,18 @@ router.get('/:bbid', middleware.loadEntityRelationships, middleware.loadWikipedi
...
+		return await entityRoutes.displayEntity(req, res);
+	}
+	catch (error) {
+		log.debug(error);
+		return next(error);
```
MonkeyDo, line 220: You've removed `next` from the arguments at the beginning of this function, this will throw an error.

src/server/routes/entity/publisher.ts
```diff
@@ -207,18 +207,18 @@ router.get('/:bbid', middleware.loadEntityRelationships, middleware.loadWikipedi
...
-			res.locals.entity.editions.sort(entityRoutes.compareEntitiesByDate);
-			return entityRoutes.displayEntity(req, res);
-		})
-		.catch(next);
+	try {
+		const editions = await Publisher.forge({bbid: res.locals.entity.bbid})
+			.editions({withRelated: editionRelationsToFetch});
+		res.locals.entity.editions = editions.toJSON();
+		res.locals.entity.editions.sort(entityRoutes.compareEntitiesByDate);
+		return await entityRoutes.displayEntity(req, res);
```
MonkeyDo, line 216: I don't think we're meant to await the function here, but rather to return a promise as we were doing before:
```suggestion
		return entityRoutes.displayEntity(req, res);
```

src/server/routes/entity/work.ts
```diff
@@ -99,26 +99,36 @@ router.get(
...
 
 		if (req.query.author) {
-			propsPromise.author =
-				Author.forge({bbid: req.query.author})
-					.fetch({require: false, withRelated: 'defaultAlias'})
-					.then((data) => data && utils.entityToOption(data.toJSON()));
+			try {
+				const data = await Author.forge({bbid: req.query.author})
+					.fetch({require: false, withRelated: 'defaultAlias'});
+				propsObject.author = utils.entityToOption(data.toJSON());
```
MonkeyDo, line 114: I think this will change, with my suggestion above to use the orm function `getEntity`, but for what it's worth compared to the previous implementation you're missing that `data &&` before calling `data.toJSON`.
This was in place because the entity fetch with `required:false` would return null or undefined, and calling `undefined.toJSON()` would throw an error.

MonkeyDo (changes requested):
Thanks for the work !

There's a few tricky cases in there, and by god I know they are hard to read ad reason through!

I left a bunch of comments, some of which will need to be applied to all the files (variable naming, utility function, etc.)

Do let me know if something is unclear or if you have questions.

src/server/routes/entity/publisher.ts
Tarunmeena0901: oops !! that was a mistake actually thank you for pointing it out

Tarunmeena0901:
Oh my, there are some rookie mistakes on my side. thank you for pointing them out

MonkeyDo:
> Oh my, there are some rookie mistakes on my side. thank you for pointing them out

Like I said, reading these promise chains and figure out what they do and when is a real pain in the ass !
It's easy to get something wrong in the process (I should know, I've made my share of "promisetakes"!)

src/server/routes/entity/entity.tsx
Tarunmeena0901: why do we have to do alertPromise = Promise.all(...)  when we already awaited the promises inside map function
```
const unlock = await new AchievementUnlock({id: achievementAlert}).fetch({
					require: true,
					withRelated: 'achievement'
				});
```
unless we want to catch individual promise rejection we may need not to use promise.all() here ........

Tarunmeena0901: same question here
```
entityJSON.relationshipSet.relationships = await Promise.all(entityJSON.relationshipSet.relationships.map(async (rel) => {
				try {
					rel.source = await commonUtils.getEntityAlias(orm, rel.source.bbid, rel.source.type);
					rel.target = await commonUtils.getEntityAlias(orm, rel.target.bbid, rel.target.type);
				}
				catch (err) {
					log.error(err);
				}
				return rel;
```

MonkeyDo: You're right, there's some subtleties here.

In short, array.map takes an array and runs the callback function on each item, then takes the return value. The resulting array would be composed of each return value.
In our case, the return value is a Promise, not the value returned from the callback function. That's why we would call `await Promise.all` to get the actual values.

Here's an example with an array of three numbers, and a callback function that awaits a timeout and returns the value.
You'll see the first arrayMapResult is an array of Promises, while the second using `await Promise.all(…` is an array of the actual return values.

image

So in the first case, that .map work like a .forEach and we can't do anything with the returned array, while we would use Promise.all if we want to do something with the returned values.
In both cases, the timeout is awaited.
Hope that clarifies it a bit!

I'll let you think through what we should be returning in which case, but do ping me if you're struggling.

Tarunmeena0901: thanks a bunch ! Concept clear 💯💯

Tarunmeena0901:
I Renamed some variables , and I also took another look at the entity.tsx file. I found some more potential improvements in terms of rewriting the logic to remove  promise chains. Although I attempted to rewrite the promises in a new syntax without altering the actual logic of the code, I still have my doubts about entity.tsx -> L745 to L817.  
please have a look @MonkeyDo 🤞🏻 🤞🏻

MonkeyDo (commented):
I had these pending comments on the PR. I will need to do another review next week, I'm finishing up for this week as we speak.

In the meantime the comments below should apply

Tarunmeena0901:
yup I thought so , working on this PR was little confusing for me back then and surely I made some rookie mistakes here ✅✅
I will fix them this weekend

src/server/routes/entity/entity.tsx
```diff
@@ -430,7 +436,7 @@ export function handleDelete(
 	const {Revision, bookshelf} = orm;
 	const editorJSON = req.session.passport.user;
 	const {body}: {body: any} = req;
-	const entityDeletePromise = bookshelf.transaction(async (transacting) => {
+	const entityDelete = bookshelf.transaction(async (transacting) => {
```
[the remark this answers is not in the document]
Tarunmeena0901: I just checked, I changed this locally but forgot to push it here
thanks for pointing this out

src/server/routes/entity/entity.tsx
```diff
@@ -481,7 +483,15 @@ export function handleDelete(
...
-	return handler.sendPromiseResult(res, entityDeletePromise, search.deleteEntity);
+	try {
+		res.send(entityDelete);
+		search.deleteEntity(entityDelete);
+		return entityDelete;
+	}
+	catch (err) {
+		log.error(err);
+		return error.sendErrorAsJSON(res, err);
+	}
```
MonkeyDo, line 494: Here there is a small subtlety which is that previously search.deleteEntity was called after the main entityDeletePromise succeeded, in such a way that an error thrown in search.deleteEntity was not sent back to the user.
Since we are now awaiting the promise further up, we don't need this try-catch block for returning the response (res.send) anymore, but we do need one for separately catching search indexing errors like so:

 ```suggestion
	res.send(entityDelete);
	try {
		search.deleteEntity(entityDelete);
	}
	catch (err) {
		log.error(err);
	}
	return entityDelete;
```


However, we might need to wrap all the code from `const entityDelete = await ...` onwards into another top-level try-catch block, and in the catch block use that `return error.sendErrorAsJSON(res, err);` .

MonkeyDo (changes requested):
Looking much better, thanks for having a look.

Just one remaining subtlety below, after that I think it's ready to merge.

src/server/routes/entity/entity.tsx
Tarunmeena0901: done 👍

 (〃￣︶￣)人(￣︶￣〃)

PR #1071, opened 2024-02-26
fix BB-782 : Unable to edit Author Credit after having set it

Tarunmeena0901:
### Problem
BB-782

### Solution

Based on the discussion in this issue [link to issue](https://tickets.metabrainz.org/browse/BB-782), I'm still unable to unveil the mystery surrounding the `isEditable` checks. However, removing it seems to improve and make things less buggy, so perhaps we're better off without it.

### Areas of Impact

author-credit-section.tsx

### DIFF
--- a/src/client/entity-editor/author-credit-editor/author-credit-section.tsx
+++ b/src/client/entity-editor/author-credit-editor/author-credit-section.tsx
@@ -58,7 +58,6 @@ type StateProps = {
 	authorCreditEditor: Record<string, AuthorCreditRow>,
 	authorCreditEnable: boolean,
 	showEditor: boolean,
-	isEditable:boolean,
 };
 
 type DispatchProps = {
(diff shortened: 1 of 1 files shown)


Tarunmeena0901:
also looked at musicbrainz entity-editor and there is no problem like this... we can edit the artist credits as per our will 

giving me another reason to believe that this was a bug

MonkeyDo:
The idea of isEditable here is to ensure users will not modify the text input directly after modifying in the author credit editor.

For the simplest case (single author, name identical), it makes it simple as typing the name to search then selecting one.
If you selected the wrong one you can simple type in the box again (isEditable is true).

However if you use the author credit editor for a more complex case (for example two authors), the resulting string representation should not be edited directly, so we disable the text input (isEditable=false) and force the user to use the AC editor instead

Does that make sense?

The same thing happens in MusicBrainz. For example, select an artist, then click "edit" to modify it (change the "Artist as credited" or add another artist) and you will see the text input becomes disabled straight away.

Closing this PR as I believe everything is working as expected, but if I misunderstood the issue do let me know and you can reopen the PR.

Tarunmeena0901:
Got it ✅✅ Now, as you explained, I think the main problem here is that after setting more than one author credit, the entire field and edit button get disabled. I think only the input field should be disabled, not the edit button, so users can still edit credits even after multiple credits are set.


image

Tarunmeena0901:
but again if that's the case then this PR wont solve this and I have to make some changes

MonkeyDo:
Yes you're absolutely right! I didn't see it was disabled once the AC is edited.
Indeed that's a bug and needs fixing, but probably best to start afresh in another PR.
Thanks for looking into it again!

PR #1073, opened 2024-02-29
Add begin date and end date attribute to relationship

aabbi15:
### Problem

Added Begin date and End date attributes to required relationship. Related to BB-490 and BB-283

### Solution

I have added a new created attribute named 'Begin date' and 'End date'. In the relationship_type__attribute_type table I have appended these attributes there for certain relationship. I have listed them in the comments as well. 

### Areas of Impact

The bookbrainz database in SQL was updated with the said attributes added. There is not much other effect on the codebase.

### DIFF
--- a/sql/migrations/2024-03-01-date-relationship-attribute/up.sql
+++ b/sql/migrations/2024-03-01-date-relationship-attribute/up.sql
@@ -1,0 +1,118 @@
+BEGIN;
+
+INSERT INTO bookbrainz.relationship_attribute_type (id, parent, root, child_order, name, description)
+VALUES
+	  (3, NULL, 1, 0, 'Begin date', 'This attribute indicates when the relationship begin.'),
+	  (4, NULL, 1, 0, 'End date', 'This attribute indicates when the relationship ended.');
+
+      
+
+
+INSERT INTO bookbrainz.relationship_type__attribute_type (relationship_type, attribute_type)
+VALUES
+        -- Author --
+        (8, 3), 
+        (8, 4),
+
+        -- Marriage --
+        (11,3),
+        (11,4),
+
+        -- Involved with --
+        (12,3),
+        (12,4),
+                
+        -- Member of Group --
+        (13,3),
+        (13,4),     
+
+        -- Subgroup --
+        (16,3),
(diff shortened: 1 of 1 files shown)


Tarunmeena0901:
hey , may be you forget but i think we also have to make changes in bookbrainz.sql file accordingly after making any sql changes. our test uses the same file to test the updated database

aabbi15:
@Tarunmeena0901 I have added it now. Kindly check.

sql/migrations/2024-03-01-date-relationship-attribute/up.sql
```diff
@@ -0,0 +1,118 @@
...
+
+        -- Art Director --
+        (28,3),
+        (28,4),
+
+        -- Employee --
+        (21,3),
+        (21,4);
+
+
```
MonkeyDo, line 113: A few suggested additions, after looking more closely at the relationship_type table:

- 18, 90 -> Pen name, Persona
- 35, 69 -> Reconstruction
- 36 -> Previous attribution
- 40 -> Owner (publishers)
- 41 -> Renamed (publishers)
- 44, 63 -> Copyright
- 45, 46 & 51 -> Licensor, Licensee
- 57 -> Reprint

sql/schemas/bookbrainz.sql
```diff
@@ -1055,4 +1055,121 @@ CREATE VIEW bookbrainz.work_import AS
 	LEFT JOIN bookbrainz.alias_set alias_set ON work_data.alias_set_id = alias_set.id
 	WHERE import.type = 'Work';
 
+
+INSERT INTO bookbrainz.relationship_attribute_type (id, parent, root, child_order, name, description)
```
MonkeyDo, line 1059: This file should not be changed in this case.
We modify it when we add or modify a table or something like that, but not when we add rows of data to the database, that is why we have separate migration scripts.
So you can remove all these additions in bookbrainz.sql

MonkeyDo (changes requested):
Sorry for only getting to reviewing this @aabbi15, and thanks for working on it !
Looks quite good, I have a suggestion or two below:

PR #1074, opened 2024-03-01
fix(profile): change Twitter to X

TheNourhan:
### Problem

BB-791

### Solution

1- changed the icon and name of Twitter to X on the index page and about us.
2- changed the X "Twitter" link  to the correct link for the bookbrainz account, the link was wrong before

### Areas of Impact

`src/client/components/pages/index.js`
`src/client/components/pages/about.js`


MonkeyDo (approved):
Thank you!

PR #1075, opened 2024-03-10
fix(BB-793): Improve the UX with the search field option entity

MohamedAli00949:
Hi there,
While searching for issues to contribute to Bookbrainz I found `BB-793` for improving the UX of the add authors to collection form, I have found it a good start for a beginner such as me so I have started to work to know where is the problem and how to solve it.

### Problem

The first time, I was trying to improve the search field element in authors to collection form such as the `BB-793` issue, then I found that the search field element at more places such as add collaborators to the collection as you can see in the images below.
Screenshot from 2024-03-09 23-53-43
Screenshot from 2024-03-10 03-29-25

### Solution

The first solution was to add a CSS class to handle the search field width to be suitable to screen width and provide a good UX, but I have found that this has happened but no effect we can notice. 
The class `Select` is responsible for handling the width of the element, but it didn't work because the cloned element took its props from the parent of the wrapper select element without taking the wrapper select props as expected.

Screenshot from 2024-03-10 03-28-11
Screenshot from 2024-03-10 03-29-06

### Areas of Impact

The PR effect on the common element `EntitySearchFieldOption` which is a part of the entity editor.

### DIFF
--- a/src/client/entity-editor/common/entity-search-field-option.js
+++ b/src/client/entity-editor/common/entity-search-field-option.js
@@ -130,7 +130,7 @@ class EntitySearchFieldOption extends React.Component {
 		if (!buttonAfter) {
 			return (
 				<>
-					{React.cloneElement(wrappedSelect, props)}
+					{React.cloneElement(wrappedSelect, wrappedSelect.props)}
 					{help && <Form.Text muted>{help}</Form.Text>}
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Thank you for finding this !

PR #1077, opened 2024-03-18
refactor(statistics) to typescript

meziyum:
### Problem
The bookbrainz project has been gradually adopting typescript. Typescript can help us prevent the code base more type safe and boost collaboration

### Solution
Refactored the statistics and any related functions to typescript.
Added the type _EditorType which can easily be imported in  the ts files wherever necessary 

### Areas of Impact
statistics page

### DIFF
--- a/src/client/components/pages/statistics.tsx
+++ b/src/client/components/pages/statistics.tsx
@@ -19,11 +19,10 @@
 import * as bootstrap from 'react-bootstrap';
 import * as utilsHelper from '../../helpers/utils';
 
-import {type Editor} from '../../../types';
 import React from 'react';
+import {type _Editor} from '../../../types';
 import {startCase as _startCase} from 'lodash';
 import {genEntityIconHTMLElement} from '../../helpers/entity';
-
 
 
 const {Table} = bootstrap;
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -9,6 +9,24 @@ export type _AuthorType = {
 
 };
 
+export interface _EditorType {
```
MonkeyDo, line 12: This should be
```suggestion
export type _Editor {
```

src/client/components/pages/statistics.tsx
```diff
@@ -19,23 +19,28 @@
 import * as bootstrap from 'react-bootstrap';
 import * as utilsHelper from '../../helpers/utils';
 
-import PropTypes from 'prop-types';
+import {_EditorType} from '../../../types';
```
MonkeyDo, line 22: And with the renaming below:
```suggestion
import {type Editor} from '../../../types';
```

MonkeyDo (changes requested):
Thanks, looks good!
Just a naming suggestion below

MonkeyDo:
Thank you!

src/types.ts
```diff
@@ -9,6 +9,24 @@ export type _AuthorType = {
 
 };
 
+export type _Editor {
```
MonkeyDo, line 12:
```suggestion
export type _Editor = {
```

PR #1081, opened 2024-03-27
Fix BB-760 : When elasticsearch configuration is missing, server crashes

Tarunmeena0901:
### Problem

[BB-760](https://tickets.metabrainz.org/projects/BB/issues/BB-760)

### Solution

As the reason behind the crashing of BB server was the absence of elastic search configuration in our configuration template , I added some checks just before we are initializng the search server in app.js .
Those checks will ensure incase of missing configuration elastic search will still get some hardcoded configuration  which will prevent it from crashing on initializing

and also incase of missing configuration error will be logged in console and a warning banner will be displayed on homepage

Screenshot 2024-03-28 001914

Screenshot 2024-03-28 001936

### Areas of Impact

- ./src/server/app.js
- ./src/api/app.js

### DIFF
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -100,7 +100,7 @@ let searchInitiated;
 // Clone search config to prevent error if starting webserver and api
 // https://github.com/elastic/elasticsearch-js/issues/33
 (async function initializeSearch() {
-    searchInitiated = await search.init(app.locals.orm, Object.assign({}, config.search));
+	searchInitiated = await search.init(app.locals.orm, Object.assign({}, config.search));
 })();
 
 // Set up constants that will remain valid for the life of the app
--- a/src/server/app.js
+++ b/src/server/app.js
@@ -34,7 +34,6 @@ import config from '../common/helpers/config';
 import express from 'express';
 import favicon from 'serve-favicon';
 import initInflux from './influx';
-import log from 'log';
 import logNode from 'log-node';
 import logger from 'morgan';
 import path from 'path';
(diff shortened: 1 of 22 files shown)


src/server/app.js
```diff
@@ -124,6 +139,24 @@ app.use((req, res, next) => {
...
+	if (!searchConfig.node || !searchConfig.auth) {
+		let msg;
+		if (process.env.DEPLOY_ENV === 'test') {
+			msg = 'Elastic search configs are not provided. Search server is not available on the test website';
+		}
+		else if (process.env.DEPLOY_ENV === 'beta') {
+			msg = 'Elastic search configs are not provided. Search server is not available on the beta website';
+		}
+		else {
+			msg = 'Elastic search configs are not provided. Search server is not available in your local environment';
```
MonkeyDo, line 151: I think we can do a single message for all environments, maybe something like 
'We could not connect to our search server, all search functionality is unavailable.'

MonkeyDo (changes requested):
Sorry for the long delay!
I have some suggestions to simplify the code a bit and deduplicate some code:

Tarunmeena0901:
There were some unexpected es lint error not releated to the changes in this PR , but I fixed them too I hope it does no make things confusing for you

MonkeyDo:
> There were some unexpected es lint error not releated to the changes in this PR , but I fixed them too I hope it does no make things confusing for you

For posterity, it's not ideal as it touches a bunch of files that are unrelated.
That being said, the changes in question are so minor that it really isn't an issue to review.

In case it's useful in the future, you can run eslint for specific files instead by passing a file path to the commend:
`npx eslint path/to/my/file`

MonkeyDo (approved):
Perfect, thanks!

MonkeyDo:
> Great solution! One thing I would like to ask: the type you mentioned here `{type ClientOptions} from '@elastic/elasticsearch';` doesn't have the properties that options should have like "auth" and "host"
> 
> ```
> options = {
> 	auth: { password: 'changeme', username: 'elastic' },
> 	node: 'http://localhost:9200',
> 	requestTimeout: 60000,
> 	host: 'localhost:9200'
> };
> ```

Thanks for pointing this out!
I took a closer look and indeed the configuration was incorrect> I took the liberty to piggyback on the PR and fix it, since there were a few places to fix it but not very obvious.
While adding the typescript type to the ES client, it also raised a separate issue which I also fixed.

Hope you don't mind my meddling!

PR #1083, opened 2024-03-30
BB-764: Unified form ISBNs bug

insane-22:
### Problem

BB-764: On the unified form, ticking and then unticking the "Automatically add ISBN" checkbox results in the generated ISBN to be left over in the "edit identifiers" modal.
Unchecking should remove the auto-generated ISBN.

### Solution

With the present code, what happens is that when the user checks the box for the first time, an action is dispatched to have the auto generated ISBN in the list of identifiers, but when he unchecks again, there was no direction to remove the added identifier. Hence, I have used this already defined removeIdentifierRow action:-
image

It requires rowId as a parameter, which we get using the addAutoIdentifier:-
image

After changes:-
https://github.com/metabrainz/bookbrainz-site/assets/116253210/ecfc1a2b-2f7e-45f0-817a-cc77243a27d9

### Areas of Impact

 src/client/unified-form/cover-tab/isbn-field.tsx
 src/client/unified-form/interface/type.ts

### DIFF
--- a/src/client/unified-form/cover-tab/isbn-field.tsx
+++ b/src/client/unified-form/cover-tab/isbn-field.tsx
@@ -53,7 +53,7 @@ function mapStateToProps(rootState:State):ISBNStateProps {
 }
 
 function mapDispatchToProps(dispatch):ISBNDispatchProps {
-	let gett:dispatchResultProps|null;
+	let autoAddedISBN:dispatchResultProps|null;
 	function onChange(value:string, autoISBN = false) {
 		const isbn10rgx = new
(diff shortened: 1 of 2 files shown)


src/client/unified-form/cover-tab/isbn-field.tsx
```diff
@@ -52,6 +53,7 @@ function mapStateToProps(rootState:State):ISBNStateProps {
 }
 
 function mapDispatchToProps(dispatch):ISBNDispatchProps {
+	let gett:dispatchResultProps|null;
```
MonkeyDo, line 56: Maybe a better name for this one would be:
```suggestion
	let autoAddedISBN:dispatchResultProps|null;
```

MonkeyDo (changes requested):
Looks good, thanks for fixing this!
Just a small naming suggestion to make the code a tad clearer.

You should also run ` npx eslint --fix src/client/unified-form/cover-tab/isbn-field.tsx` to format the file, ESLint is complaining at the formatting otherwise

MonkeyDo (approved):
Perfect, thank you for another good bugfix!

PR #1091, opened 2024-05-29
Configuration Migration of ESLint configuration file from the eslintrc format to the new flat config format

insane-22:
### Problem

According to ESLint's official docs- "The flat config file format has been the default configuration file format since ESLint v9.0.0". Codebase isn't updated with this new config

### Solution

Added new eslint.config.js file and migrated code from .eslintrc.js and .eslintignore to this new config file


eslint.config.js
```diff
@@ -0,0 +1,444 @@
+const { FlatCompat } = require("@eslint/eslintrc");
```
MonkeyDo, line 1: You can write all these imports with the import syntax to modernize it too, like so:
```js
import { FlatCompat } from'@eslint/eslintrc';
import pluginImport from 'eslint-plugin-import';
```
etc.

eslint.config.js
```diff
@@ -0,0 +1,444 @@
...
+const pluginReact = require("eslint-plugin-react");
+const pluginImport = require("eslint-plugin-import");
+const pluginNode = require("eslint-plugin-node");
+const pluginTypeScript = require("@typescript-eslint/eslint-plugin");
+
+const compat = new FlatCompat({
```
MonkeyDo, line 7: As the warning is howing here (in the [files tab of the PR](https://github.com/metabrainz/bookbrainz-site/pull/1091/files) on github), this compat configuration you create here is never used.
Either something is wrong with the configuration as it stands, or the FlatCompat and associated package are not actually needed...
Not sure which it is.

MonkeyDo (changes requested):
Thanks for starting this update!

Something was wrong with the yarn lockfile, I think maybe it got corrupted somehow, or you tried to manually correct something in there?
In any case I regenerated it so at least the tests can run.

That shows that at least the eslint.config.js file has a lot of linting issues, ironically :)
You can fix a lot of the issues automatically by running `yarn lint eslint.config.js --fix`.

More problematically, the previous configuration file eslintrc.js still exists in the repo, and I can't tell you which config is used...
I suppose it should probably be deleted in favor of the new file.
Also do a global search in the repo for the file name, to replace it with the new one where appropriate. 

I also think this PR would be the perfect occasion to update ESLint to v9 and allow us to test the changes with the target version; without that, i'm not sure the PR actually tests that the config changes will work for v9.

eslint.config.js
```diff
@@ -0,0 +1,444 @@
...
+		},
+	],
+	"import/no-mutable-exports": ERROR,
+	"import/no-named-as-default": ERROR,
+	"import/no-named-as-default-member": ERROR,
+	"import/no-named-default": ERROR,
+	"import/no-unassigned-import": ERROR,
+};
+
+module.exports = [
```
MonkeyDo, line 387: Forgot one :sweat_smile: 
```suggestion
export default [
```

insane-22:
Thank you very much @MonkeyDo for again helping while solving the issue :) 
I actually was trying a few things and that's why lock file error was there. I also tried the the upgrade to eslint v9 but there is a problem, "eslint-plugin-import" dependency does not have compatibility with Eslint v9 (https://github.com/import-js/eslint-plugin-import/issues/2948).

eslint.config.js
insane-22: yes, would make it eslint.config.mjs and add the changes

insane-22:
but at the same time eslint.config.js doesnt work with versions prior to v9, hence it doesnt make sense to leave it this way.

One work-around I could find is https://github.com/import-js/eslint-plugin-import/issues/2948#issuecomment-2148832701
or we can wait until the whole v9 ecosystem is ready to use. what in your opinion shall be the next steps?

MonkeyDo:
I read up ont he whole discussion, and it looks like v9 support in the plugin is going to take some time.
In the meantime, from what I read you should be able to use this drop-in replacement: https://github.com/un-ts/eslint-plugin-import-x
Please add a TODO comment note where you can, with a short explanation that we are using this package as a temporary replacement until v9 support is out, that it should be replaced as soon as available, and a link to the eslint-plugin-import issue.

MonkeyDo:
Bump @insane-22, do you have the time to look into the alternate plugin, or would you prefer I do?
Thanks again for looking at these config changes!

insane-22:
Hello @MonkeyDo, sorry for this delay, was caught up in a few things and thanks a lot for this reminder. 
Please allow me some more time, I will be back with the changes :)

insane-22:
Hello hello, I'm back with new sets of errors :]

"@eslint/eslintrc@3.1.0: The engine "node" is incompatible with this module. Expected version "^18.18.0 || ^20.9.0 || >=21.1.0". Got "20.6.1" "
when i changed dockerfile as "FROM metabrainz/node:20 as bookbrainz-base"
it still has node v20.6.0, which is incompatible, adding any advanced versons shows
"failed to solve: metabrainz/node:20.9.0: failed to resolve source metadata for docker.io/metabrainz/node:20.9.0: docker.io/metabrainz/node:20.9.0: not found"

MonkeyDo (approved):

MonkeyDo:
Thanks a lot for working on this, and for coming back to it multiple times :)
I'll take over from here with the ESLint 9 upgrade, if you don't mind, as it gets pretty complicated from here onwards: 7000+ errors the first time i ran it, 4000+ after I managed to make module resolution work, 1000+ after a few tweaks, and now 100+ of various issues that need a style guidelines decision to be taken...

PR #1094, opened 2024-06-05
fix(helpers): Replace deleted utility by lodash function

kellnerd:
This broke in https://github.com/metabrainz/bookbrainz-site/commit/d44d67a1e4ad54302ae3f0230cd77d402d253585

Untested, but this is the same function with which all other occurrences were replaced in the above commit. Should have been the last occurrence.


MonkeyDo (approved):
Thank you, didn't see that one (obviously)

PR #1101, opened 2024-06-25
Fix search entity type

MonkeyDo:
Broke the search autocomplete endpoint in #1093  when I forgot we can use an array of types in some search endpoints, and calling `toString` on the array of types results in a comma separated joined string which is not recognized as an entity type


MonkeyDo (commented):
Deployed to beta to fix recently deployed bug, works fine

PR #1104, opened 2024-07-16
Adding a work type lyrics

Medha170:
### Problem

BB-773 Adding a work type 'Lyrics'

### Solution

This PR shows the changes I have done in the actual database of bookbrainz which is inserting a row with lyrics and making it a child of poem. In the up.sql file of 2023-05-15-work-types.

### Areas of Impact

Added an additional column for the lyricists to publish a book.
Screenshot from 2024-07-13 22-20-16

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,7 +24,7 @@ services:
     image: postgres:12.3
     user: postgres
     ports:
-      - "127.0.0.1:5433:5432"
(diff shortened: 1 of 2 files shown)


MonkeyDo:
Hello! 
Thank you for opening a PR!
In this case, it would have required to be created in a separate migration file rather than modifying an old one, but for a single row I am just going to add it manually to the database.
Thanks again for the addition !

PR #1105, opened 2024-07-17
chore: Allow annotation revision ids to be null

MonkeyDo:
Also see https://github.com/metabrainz/bookbrainz-data-js/pull/320
 
For imported entities, we do not keep a history of all modifications.
As a result, editing import_* entities is not tied to any revisions.

This is an issue for annotations, whose `last_revision_id` column does not currently accept NULL values.

These small changes allow entering a row in the `annotation`table with `last_revision_id = NULL`, and ensure the website does not trip up if the value is missing (it is used to display the last modification time and link to the revision)

Before: 
image

After:
image


MonkeyDo:
I checked (locally) that I can have an annotation with last_revision_id = NULL, and the website does not freak out.

Modifying then the annotation creates a new revision as expected, with the annotation last_revision_id set correctly (meaning updating annotations will work after validated imports)

kellnerd (approved):
Cool, not having an annotation history or revision comparison feature (as on MusicBrainz) definitely made it easier to change this now. We can worry about these edge cases later :joy:

I only have one question: Should this PR target `import-entities` directly rather than `master` or do you want to merge `master` into `import-entities` as well? (Doing that semi-regularly also keeps the feature branch fresh until it is ready.)

MonkeyDo:
> Cool, not having an annotation history or revision comparison feature (as on MusicBrainz) definitely made it easier to change this now. We can worry about these edge cases later 😂
> 
> I only have one question: Should this PR target `import-entities` directly rather than `master` or do you want to merge `master` into `import-entities` as well? (Doing that semi-regularly also keeps the feature branch fresh until it is ready.)
> 
> Edit: I just checked and we already have merge conflicts due to `src/common/helpers/search.ts` and `yarn.lock` 🙄

Yes, the merge conflicts are inevitable...
I'll merge this into `import-entity`, there's no need for that change without the feature .

MonkeyDo:
Closing in favor of #1106

PR #1107, opened 2024-07-17
Display and preserve annotation of imported entities

kellnerd:
Display annotations for all types of imported entities and preserve the annotation during the whole "Edit & Approve" process.
The corresponding changes for the "Approve" action are in https://github.com/metabrainz/bookbrainz-data-js/pull/321.

### DIFF
--- a/src/server/routes/import-entity/edition-group.js
+++ b/src/server/routes/import-entity/edition-group.js
@@ -32,9 +32,10 @@ router.param(
 		'EditionGroupImport',
 		[
 			'editionGroupType',
-			'editions.defaultAlias',
-			'editions.disambiguation',
-			'editions.releaseEventSet.releaseEvents'
+			// TODO: Reenable edition rels once imports have BBIDs
+			// 'editions.defaultAlias',
+			// 'editions.disambiguation',
+			// 'editions.releaseEventSet.releaseEvents'
 		],
 		'Edition Group Import not found'
 	)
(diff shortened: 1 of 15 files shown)


MonkeyDo (approved):
Yes, perfect :rocket:

kellnerd:
I've found a couple more issues while I was testing edition group imports (entity type name casing strikes again). One of them is even directly related to annotations, so I thought I might add these commits to this PR as well.

src/server/routes/import-entity/edition-group.js
```diff
@@ -32,9 +32,10 @@ router.param(
 		'EditionGroupImport',
 		[
 			'editionGroupType',
-			'editions.defaultAlias',
-			'editions.disambiguation',
-			'editions.releaseEventSet.releaseEvents'
+			// TODO: Reenable edition rels once imports have BBIDs
+			// 'editions.defaultAlias',
+			// 'editions.disambiguation',
+			// 'editions.releaseEventSet.releaseEvents'
```
kellnerd, line 38: > Unexpected trailing comma (comma-dangle)

Sigh, one more example why this ESLint rule is stupid... it often forces you to pointlessly edit neighbouring lines. Can we just disable this rule or migrate to always using dangling commas?

MonkeyDo: You have my blessing to disable the rule. I've never seen it as a relevant code styling rule.

kellnerd: Cool, I've dropped the rule in 2fc11a2. We might want to do the same across all repos, but I better won't do that now or I will find some more rules which should be changed or turned off :zipper_mouth_face:

PR #1109, opened 2024-07-24
Support series imports & Fix edition imports

kellnerd:
On top of #1107 (diff: https://github.com/kellnerd/bookbrainz-site/compare/import-annotation...import-series)

Adds the missing routes, pages and data transformation functions for imported series.
Since imported entities can't have relationships yet, we have to handle the case that a series has no defined items (which wasn't necessary for regular series).
And I forgot to add the item entity type to the SQL view in one of my previous PRs.

----

Since the necessary changes to make editing and approval of edition imports work are very minor, I have included these here as well.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -204,7 +204,6 @@ const stylisticIssuesRules = {
 			properties: 'always'
 		}
 	],
-	'comma-dangle': ERROR,
 	'comma-spacing': ERROR,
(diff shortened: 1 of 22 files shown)


MonkeyDo (approved):
Looks good :+1: 

I manually added a fake series import row to test it, works fine :rocket:

PR #1133, opened 2024-10-28
Funding: add .well-known/funding-manifest-urls  file and URL

MonkeyDo:
See https://github.com/metabrainz/metabrainz.org/pull/487/files, https://floss.fund/funding-manifest/ and https://en.wikipedia.org/wiki/Well-known_URI

This link back to the MeB.org funding.json can be accessed two ways:
1. In the github repo, for use with repositoryUrl.wellKnown using "https://github.com/metabrainz/bookbrainz-site/blob/main/.well-known/funding-manifest-urls"
2. for use with webpageUrl.wellKnown, point to the new path "https://bookbrainz.org/.well-known/funding-manifest-urls"

### DIFF
--- a/.well-known/funding-manifest-urls
+++ b/.well-known/funding-manifest-urls
@@ -1,0 +1,1 @@
+https://metabrainz.org/funding.json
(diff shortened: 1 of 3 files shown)


MonkeyDo:
Closed in favor of #1135

PR #1137, opened 2024-11-07
fix(BB-806): add validation for float values in width, height, depth fields

Monika-Dangar:
### Problem

[BB-806]
Can't submit entries with float values for dimensions

### Solution

Updated input parsing from `parseInt()` to `parseFloat()` to accept float values so validation can notify users of invalid inputs.

### Areas of Impact

- **`edition-section.tsx`**: Modified input parsing and validation for dimensions.
- **`edition-section-merge.tsx`**: Applied the same changes for consistency.


config/config.json.example
MonkeyDo, line 1: This file should not be deleted.
Or at least, the deletion should not have been committed,..

.gitignore
```diff
@@ -8,6 +8,7 @@
 /reports
 /config/config.json
 /config/config.local.json
+/config.json
```
MonkeyDo, line 11: Is this required? I don't think we ever have a config file at the root of the project.
The config file is already in ignored two lines above.

sql/migrations/2024-02-07-credit-section/up.sql
MonkeyDo, line 1: Not sure why this file was committed in this PR, it does not belong here.

src/client/entity-editor/edition-section/edition-section-merge.tsx
```diff
@@ -333,13 +333,13 @@ function mapDispatchToProps(dispatch: Dispatch<Action>): DispatchProps {
 			dispatch(updateAuthorCredit(selectedAuthorCredit));
 		},
 		onDepthChange: (event) => dispatch(debouncedUpdateDepth(
-			event.target.value ? parseInt(event.target.value, 10) : null
+			event.target.value ? parseFloat(event.target.value) : null
```
MonkeyDo, line 336: These few changes are not going to be sufficient, as we have data validators running on the back-end after the submission.

See for example:
 https://github.com/metabrainz/bookbrainz-site/blob/739b2641c344938582ff06c3e13dfcbf2d2574c0/src/client/entity-editor/validators/edition.ts#L36-L38

Note that it validates integer value, so the code changes here will make submissions fail.

Please set up the project on your local machine so that you can run your changes and test them, before submitting PRs for the same.

MonkeyDo (changes requested):
The config/config.json.example file should not be deleted

sql/migrations/2024-02-07-credit-section/up.sql
Monika-Dangar: Thank you for pointing this out! It seems I mistakenly included the unnecessary files, and I’ve removed them from the PR now.

src/client/entity-editor/edition-section/edition-section-merge.tsx
Monika-Dangar: Regarding the issue with the validation not working properly, I understand that the changes I made won't work due to the back-end validation. I am currently looking into it and figuring out a proper solution so that it works seamlessly with the back-end as well. I'll update the PR as soon as I have a fix for that.

MonkeyDo:
Hi! Responding here to your message in chat, so that it's not lost in the noise, and because I have a lot to write :)

So there are three parts you're looking at:
1. one is inputing float values in the website page ()
2. one is validating the data (on  the server) when you create/edit an entity
3. finally there is storing the value in the database

For number 1. your changes using `parseFloat` should do the trick :+1: 
<hr>
For number 3, indeed there is a change needed to the database schema, as these values are currently set as SMALLINT which can only store integers:
https://github.com/metabrainz/bookbrainz-site/blob/3c448384a167abc32bf9402b8fd14f7fd476caf5/sql/schemas/bookbrainz.sql#L349-L352
We can use the REAL type instead: https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-FLOAT

To achieve this, you will need to modify the lines in the bookbrainz.sql schema file I linked to above, but also create a database update file that we can run to modify our DBs (locally and in production) in both directions.
Take for example the last migration script:
https://github.com/metabrainz/bookbrainz-site/tree/master/sql/migrations/2024-02-07-credit-section
The up.sql file should change the type of the relevant columns to REAL, while the down.sql file should change from REAL to INT (in case we need to revert the migration).

<hr>

For number 2, the best advice I have is to first work on the tests, which will show you the issue, and confirm future changes don't break anything:
To run the Edition validators test file, run the command `yarn test test/src/client/entity-editor/validators/test-edition.js`.

First, create a new test helper for validator functions, based on the existing [`testValidatePositiveIntegerFunc`](https://github.com/metabrainz/bookbrainz-site/blob/3c448384a167abc32bf9402b8fd14f7fd476caf5/test/src/client/entity-editor/validators/helpers.js#L31): the new `testValidatePositiveFloatFunc` helper function should test more or less the same test, but adapted to float values.

Then, in the [test/src/client/entity-editor/validators/test-edition.js](https://github.com/metabrainz/bookbrainz-site/blob/3c448384a167abc32bf9402b8fd14f7fd476caf5/test/src/client/entity-editor/validators/test-edition.js#L67-L85) file, you can modify the relevant tests (only for the columns you modified) to use your new testValidatePositiveFloatFunc.
At that point, tests will fail, which is what we expect.

Finally, let's modify the validators themselves:
create a new validator based on [`validatePositiveInteger`](https://github.com/metabrainz/bookbrainz-site/blob/3c448384a167abc32bf9402b8fd14f7fd476caf5/src/client/entity-editor/validators/base.ts#L72-L84) called `validatePositiveFloat`.
Then, in the [edition validators file](https://github.com/metabrainz/bookbrainz-site/blob/3c448384a167abc32bf9402b8fd14f7fd476caf5/src/client/entity-editor/validators/edition.ts#L36C1-L39C1), modify the relevant functions (validateEditionSectionDepth, validateEditionSectionWidth, etc.) to use your new validatePositiveFloat.

Now running the edition tests should pass  and that's the end of the road!

kellnerd:
Didn't we agree not to allow float values for dimensions and weights the last time this came up?
I still don't see the point to capture anything with sub-millimeter accuracy.

Monika-Dangar:
> Didn't we agree not to allow float values for dimensions and weights the last time this came up? I still don't see the point to capture anything with sub-millimeter accuracy.

To move forward, could you both provide some guidance on how you'd like me to proceed? Should I:
1. Continue with the changes as per MonkeyDo’s suggestion (allowing float values for dimensions)?
2. Revisit the decision and stick to integers for dimensions as Kellnerd suggested?

PR #1152, opened 2025-02-06
fix(auth): replaced authentication middleware with one intended for post requests. (Note: docker node 18 -> 20)

andrewsu31098:
### Problem

[BB-813](https://tickets.metabrainz.org/browse/BB-813)
The cbReviewModal is triggering a post request whose URL is being 
stored and used as a redirection link later on during authentication.
This creates a redirection bug when a user attempts to sign in on any of the pages with said cbReviewModal.
- work
- edition groups
- authors
- series

### Solution

The route handler for /critiquebrainz/refresh has been updated to use
isAuthenticatedForHandler (which authenticates without redirecting) instead of isAuthenticated which was filling in the redirect link.

4 Additional Route Handlers have been updated to protect against the same bug.
Here's a [Google Document with images](https://docs.google.com/document/d/1rCKs1Yp_Gs5eRJKHCpbxsojq6M2adevqOMN5y6Ksj-Q/edit?tab=t.0#heading=h.tn1b6d5bh6w0) of user testing showing that each of the updated endpoints is fully functional.

### Areas of Impact

 - Server `/critiquebrainz/refresh`  post endpoint has been updated.
 - Docker Node Version updated from 18->20 due to version 18 Node's incompatibility with new Mocha Version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM metabrainz/node:18 as bookbrainz-base
+FROM metabrainz/node:20 as bookbrainz-base
 
 ARG DEPLOY_ENV
 ARG GIT_COMMIT_SHA
--- a/src/client/components/pages/externalService.js
+++ b/src/client/components/pages/externalService.js
@@ -47,35 +47,34 @@ class ExternalServices extends React.Component {
 						alertDetails: 'Something went wrong. Please try again.',
 						alertType: 'danger'
 					});
-				}				
+				}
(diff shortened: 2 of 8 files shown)


MonkeyDo (approved):
Great job finding this easy solution, I missed it multiple times while reviewing the code, looking for the culprit :clap: 

I've added a few more changes myself to other POST requests using the wrong authentication middleware.

PR #1156, opened 2025-02-25
fix: cbReviewModal star rating now sets on the first click (Note: Docker Node 18->20)

andrewsu31098:
### Problem

[BB-838](https://tickets.metabrainz.org/browse/BB-838)

### Solution

The "react-simple-star-rating" does not appear to reflect immediate state changes.
The Rating component has been updated to directly set "value" in response to state changes instead of the "ratingValue" prop, now immediately displaying the star value selected by the user.

### Areas of Impact

- cbReviewModal.tsx altered. 
- Ratings show correctly on CritiqueBrainz and in the ReviewCard displayed on the work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM metabrainz/node:18 as bookbrainz-base
+FROM metabrainz/node:20 as bookbrainz-base
 
 ARG DEPLOY_ENV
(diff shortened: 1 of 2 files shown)


MonkeyDo (approved):
Thank you! Good find. :rocket:

PR #1164, opened 2025-04-05
BB-542: Create a tab view for entities 

KasukabeDefenceForce:
### Problem
[BB-542](https://tickets.metabrainz.org/browse/BB-542) : Create a tab view for entities 

### Solution
use react bootsrap's tab viewed components to create a clean and organized space for showing entity data.

#### Before
Url: https://bookbrainz.org/author/128d9490-ee19-4270-a070-32e0a36847f5

[Screencast from 2025-04-05 15-30-30.webm](https://github.com/user-attachments/assets/fc871e4d-7da7-4c4c-b4a1-09375cc5bf27)

#### After

Desktop View:

[Screencast from 2025-04-05 15-54-14.webm](https://github.com/user-attachments/assets/1c7df668-712d-48fb-adc2-4302b6888eb5)

Mobile View:

[Screencast from 2025-04-05 15-55-21.webm](https://github.com/user-attachments/assets/73aa09fb-1377-4d3a-9be5-b63246155123)

### DIFF
--- a/src/client/components/pages/entities/author.js
+++ b/src/client/components/pages/entities/author.js
@@ -40,7 +40,7 @@ import {labelsForAuthor} from '../../../helpers/utils';
 
 const {deletedEntityMessage, extractAttribute, getTypeAttribute, getEntityUrl,
 	ENTITY_TYPE_ICONS, getSortNameOfDefaultAlias, transformISODateForDisplay} = entityHelper;
-const {Button, Col, Row} = bootstrap;
+const {Button, Col, Row, Tabs, Tab} = bootstrap;
 
 function AuthorAttributes({author}) {
 	if (author.deleted) {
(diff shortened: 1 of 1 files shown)


KasukabeDefenceForce:
image

Reviews are out of alignment on mobile view. Should I fix this in this PR or in another one?

KasukabeDefenceForce:
image

In the new UI, the release date column goes beyond the width of the rest of components on the page. Should I add a horizontal scroll or is this okay?

KasukabeDefenceForce:
https://github.com/metabrainz/bookbrainz-site/pull/1164/commits/e626dc4834017b185ddd942af040bf4d4549442e here I have aligned the heading and the add a review button.

Before:
image

After:
image

MonkeyDo:
Hi @KasukabeDefenceForce nice to see you around here again!

Thanks for opening a PR for this, it's a feature I've wanted to see for a long time, and is getting pretty important now with more rich data in BB.

I think we need to discuss the tabs; I think some of the ones you have here totally make sense ("Overview", "Edition", "Collections") while others not as much ("Annnotation", "Add Work/Edition" should probably be in the overview instead).
As a general rule, we want to separate the type of information that is 1. plentiful and 2. most useful to users.

The type of tabs we have might change depending on the entity type.
For Authors, I would definitely want to see Works and Editions (maybe a Translations tab?). There are a lot of these types of relationships and they are the most looked for information (i.e. what are 'the books' by this author?)

KasukabeDefenceForce:
Hey @MonkeyDo , nice to see you too! Thanks for taking time to review my PR. 

Regarding the tabs, should I then filter out the works and translations out of relationships tab and leave relationships tab for less prominent relationships (Like Author-Author, Author-Publisher) ? 
Should series be clubbed into the works data or showed separately?

src/client/components/pages/entities/title.js
```diff
@@ -34,17 +34,18 @@ function EntityTitle({entity, handleModalToggle}) {
 	const label = getEntityLabel(entity);
 	return (
 		<div>
-			<h1>
+			<h1 className="d-flex justify-content-between">
```
MonkeyDo, line 37: Not so sure about this, it seems to cause an issue with disambiguations:
image

In a flex container, you can add `margin-left: auto` to the last element (the review button) to automatically align to the right. Maybe that can work here instead of justify-content-between on the parent div?

MonkeyDo (commented):
OK, sorry for the delays, it's a very important feature that deserves to be fully thought through, and I ran into some problems while trying to describe my vision.
There is an issue with how we define translations that prevents us from easily separating them from original works. Not exactly sure how to deal with that yet...but I'm open to ideas :)

In the meantime, some review comments below.

PR #1168, opened 2025-05-11
BB-855: Add author and publisher information in edition API

KasukabeDefenceForce:
### Problem
[BB-855](https://tickets.metabrainz.org/browse/BB-855) Add author and publisher information in edition API

### Solution
Update the `/edition/{bbid}` endpoint to share the author and publisher information as well.

### DIFF
--- a/src/api/helpers/formatEntityData.ts
+++ b/src/api/helpers/formatEntityData.ts
@@ -100,6 +100,82 @@ export function getWorkBasicInfo(work: Record<string, unknown> | null | undefine
 }
 
 /**
+ * A function to extract structured author credit details from an Edition ORM entity.
+ * @function
+ * @param {object} edition - an edition ORM entity
+ * @returns {object|null} an object containing the author credit data, or null if not available.
+ *
+ * @example
+ * 		const edition = await orm.func.entity.getEntity(orm, 'Edition', bbid, relations);
+ *		filterAuthorCredit(edition);
+ *		/* => {
+ *			"authorCount": 1,
+ *			"id": 135,
+ *			"names": [
+ *				{
+ *					"authorBBID": "e5c4e68b-bfce-4c77-9ca2-0f0a2d4d09f0",
+ *					"authorCreditID": 135,
+ *					"name": "J. K. Rowling"
+ *				}
+ *			]
+ *		}
+ */
+
+export function filterAuthorCredit(edition: any) {
+	if (_.isNil(edition) || _.isNil(edition.authorCredit)) { return null; }
+
+	const credit = edition.authorCredit;
+	return {
+		authorCount: credit.authorCount || null,
+		id: credit.id || null,
+		names: (credit.names || []).map(({authorCreditID, authorBBID, name}) => ({
+			authorBBID,
+			authorCreditID,
+			name
+		}))
+	};
+}
+
+/**
+ * A function to extract basic publisher information from an Edition ORM entity.
+ * @function
+ * @param {object} edition - an edition ORM entity
+ * @returns {object|null} an object containing publisher set data, or null if not available.
+ *
+ * @example
+ * 		const edition = await orm.func.entity.getEntity(orm, 'Edition', bbid, relations);
+ *		filterPublisherSet(edition);
+ *		/* => {
+ *			"id": 2217,
+ *			"publishers": [
+ *				{
+ *					"bbid": "d30d7df9-eb6f-4bd5-a69f-a9a3362a6f4a",
+ *					"ended": false,
+ *					"name": "Bloomsbury",
+ *					"publisherType": "Publisher",
+ *					"sortName": "Bloomsbury"
+ *				}
+ *			]
+ *		}
+ */
+
+export function filterPublisherSet(edition: any) {
+	if (_.isNil(edition) || _.isNil(edition.authorCredit)) { return null; }
+
+	const {publisherSet} = edition;
+	return {
+		id: publisherSet.id || null,
+		publishers: (publisherSet.publishers || []).map(({bbid, name, sortName, ended, publisherType}) => ({
+			bbid,
+			ended,
+			name,
+			publisherType,
+			sortName
+		}))
+	};
+}
+
+/**
  * A function to extract the basic details of an Edition ORM entity
  * @function
  * @param {object} edition - an edition ORM entity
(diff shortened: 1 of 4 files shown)


src/api/helpers/formatEntityData.ts
```diff
@@ -99,6 +99,82 @@ export function getWorkBasicInfo(work: Record<string, unknown> | null | undefine
...
+ */
+
+export function filterAuthorCredit(edition: any) {
+	if (_.isNil(edition) || _.isNil(edition.authorCredit)) { return null; }
+
+	const credit = edition.authorCredit;
+	return {
+		authorCount: credit.authorCount || null,
+		id: credit.id || null,
+		names: (credit.names || []).map(({authorCreditID, authorBBID, name}) => ({
```
MonkeyDo, line 131: Proposed refactoring with destructuring assignment:

```suggestion
	const {authorCredit} = edition;
	const {id, authorCount, names} = authorCredit;
	return {
		authorCount: authorCount ?? null,
		id: id ?? null,
		names: (names ?? []).map(({authorCreditID, authorBBID, name}) => ({
```

I also wonder if we want to return `null` or if we could instead use shorthand properties like you did below in the mapping function
```
return {
		authorCount,
		id,
....
```

MonkeyDo, line 131: Not sure what is happening here, but it looks like the mapping function just returns the same object structure, so is there a point in it?

src/api/helpers/formatEntityData.ts
```diff
@@ -99,6 +99,82 @@ export function getWorkBasicInfo(work: Record<string, unknown> | null | undefine
...
+ *		}
+ */
+
+export function filterPublisherSet(edition: any) {
+	if (_.isNil(edition) || _.isNil(edition.authorCredit)) { return null; }
+
+	const {publisherSet} = edition;
+	return {
+		id: publisherSet.id || null,
+		publishers: (publisherSet.publishers || []).map(({bbid, name, sortName, ended, publisherType}) => ({
```
MonkeyDo, line 168: Same question here as above, wondering if the mapping function does anything.

src/api/helpers/formatEntityData.ts
```diff
@@ -99,6 +99,82 @@ export function getWorkBasicInfo(work: Record<string, unknown> | null | undefine
...
+ *				{
+ *					"authorBBID": "e5c4e68b-bfce-4c77-9ca2-0f0a2d4d09f0",
+ *					"authorCreditID": 135,
+ *					"name": "J. K. Rowling"
+ *				}
+ *			]
+ *		}
+ */
+
+export function filterAuthorCredit(edition: any) {
```
MonkeyDo, line 124: In terms of naming conventions, "filter" suggests that we are manipulating the results to filter out some items using some comparison or other.
I don't think this is what this function does, so I suggest something like `getAuthorCredits`

src/api/helpers/formatEntityData.ts
```diff
@@ -145,7 +222,8 @@ export function getEditionBasicInfo(edition: any) {
 			height: _.get(edition, 'height', null),
 			languages: getLanguages(edition),
 			pages: _.get(edition, 'pages', null),
-			releaseEventDates: _.get(edition, 'releaseEventSet.releaseEvents', []).map((event) => event.date),
+			publisherSet: filterPublisherSet(edition),
```
MonkeyDo, line 225: Generally I try to avoid presenting technical terminology like 'set' to end users, for whom it will make less sense:
```suggestion
			publishers: filterPublisherSet(edition),
```

In our case, that means we want to return publisherSet.publishers directly, as I don't think we return the set id (which is an internal id) in any of the other set types.
Look for example at how we return the aliasSet.aliases in `getEntityAliases` for comparison

src/api/routes/edition.js
```diff
@@ -96,32 +98,55 @@ const editionError = 'Edition not found';
...
+ *                   authorBBID:
+ *                     type: string
+ *                     example: 'e5c4e68b-bfce-4c77-9ca2-0f0a2d4d09f0'
+ *                   authorCreditID:
+ *                     type: integer
+ *                     example: 135
+ *                   name:
+ *                     type: string
+ *                     example: 'J. K. Rowling'
+ *        publisherSet:
```
MonkeyDo, line 124: Will need to follow the name change described above

MonkeyDo (changes requested):
Thanks for the additions :)

Some comments and suggestions below.

src/api/helpers/formatEntityData.ts
KasukabeDefenceForce: The objects in names array have 5 properties:
1. authorCreditId
2. positon
3. authorBBID
4. name
5. joinphrase
and I felt we dont need the postion and joinphrase. So I am removing them in the mapping function.

src/api/helpers/formatEntityData.ts
KasukabeDefenceForce: Similarly, here I am just sharing the properties which I feel might be useful. 
Here is a list of all the available properties:
bbid, dataid, revisionId, master, annotationId, disambiguationId, disambiguation, defaultAliasId, name, sortName, beginYear, beginMonth, beginDay, endYear, endMonth, endDay, ended, areaId, typeId, publisherType, aliasSetId, identifierSetId, relationshipSetId, type, _pivot_set_id, _pivot_publisher_bbid, beginDate, endDate

src/api/helpers/formatEntityData.ts
MonkeyDo: I see, thanks for the clarifications.
I think we will need the position and joinphrase to re-compose the full author credit phrase.

For example (choosing a random recent entry with more than one author): https://bookbrainz.org/edition/c82ed17a-4627-48d1-9265-1b0a8ef34b37
Here if you had your own website where you wanted to present this edition with the author credits, you would be left with an array of names ["Daniel Keel","Daniel Kampa"], hopefully in the original order (not guaranteed).
With the position and joinphrase you can reconstitute "Daniel Keel und Daniel Kampa" like we do on the BB website itself.

MonkeyDo:
I amde some long awaited changes to the deployment setup to allow for more easily deploying different images in prod/test.

I deployed this PR to the api.test.bb domain if you want to test.

Here is an interesting output in the logs, some issue with the YAML definition in the docs:

```
Not all input has been taken into account at your final specification.
Here's the report: 


 Error in src/api/routes/edition.js :
YAMLSemanticError: Plain value cannot start with a tab character at line 75, column 1:

			  type: object
^^^^^^^^^

YAMLSemanticError: Nested mappings are not allowed in compact mappings at line 75, column 12:

			  type: object
           ^^^^^^…

YAMLSemanticError: Implicit map keys need to be on a single line at line 75, column 12:

			  type: object
           ^^^^^^…

```

MonkeyDo:
Tests are failing but I haven't looked into it yet: https://github.com/metabrainz/bookbrainz-site/runs/42306585171#r0s3

PR #1171, opened 2025-05-13
chore: Modify deployment setup

MonkeyDo:
I've been meaning to do that for a year or two now...

These changes to our deployment setup allow us to deploy an arbitrary docker image name to any (production/beta/test) environment (currently requires deploying 'beta' image to beta, 'test' image to test, etc. otherwise it breaks).

Copying how it is done in ListenBrainz.


MonkeyDo:
@amCap1712 since you are the one who made the same changes on ListenBrainz, I would ask you to please review this to make sure I didn't poop anything.
I will say that I deployed this branch to bookbrainz-test, which was not possible before I introduced those changes, so I know it is working, at least :)
I could still have missed a hundred small things though, my knowledge in this area has a lot of holes.

PR #1229, opened 2026-02-13
(CI): Update elasticsearch action in CI workflow

MonkeyDo:
Getting node version error starting today with getong/elasticsearch-action@v1.2 :
> docker: Error response from daemon: client version 1.40 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version.

v1.3 is available and could maybe fix, but we might as well try out the official ES github action


MonkeyDo:
Fixed in #1234

PR #1233, opened 2026-02-19
Fix: replace Button with anchor tag in EntityAnnotation toggle

garvthakre:
### Problem
 Related ticket: [BB-869](https://tickets.metabrainz.org/browse/BB-869)
This ticket has two issues:

1.Long annotations getting cut halfway through ,this is already fixed on the live site and locally.
2.The "Show more/less" toggle using a Bootstrap Button component ,this is still unfixed.
 
### Solution
1.Replaced the Bootstrap Button component with a plain <a> tag with role="button" for better semantics and accessibility.
2.Added e.preventDefault() to prevent the default anchor scroll behavior.
3.Removed the now-unused Button import.

### Areas of Impact
src/client/components/pages/entities/annotation.js


garvthakre:
@MonkeyDo , sorry  to ping u again but its just a small change and u already reviewed it i guess but the merge got stopped due to linting issue and i fixed that so can u just have a look at this one and tell me if there is any last  changes required too

PR #1281, opened 2026-04-07
Parameterized queries for raw SQL

eulerbutcooler:
# Problem

Four `knex.raw()` calls use JS template literal interpolation to embed values directly into SQL query strings which is not the standard way of doing a knex.raw() query.

- `src/server/helpers/collections.js` - `getCollectionItems()` interpolates `collectionId`, `size`, and `from` 
- `src/server/helpers/achievement.js` - `processSprinter()` interpolates `editorID`, getConsecutiveDaysWithEdits() interpolates `editorID` and `days`
- `src/server/routes/editor.tsx` - `getProgress()` interpolates `editorID`

All though none of these queries are prone to SQL injectioning as all values either come from an authenticated session or are validated/hardcoded it would still be a decent code improvement.

# Solution

Replace all string interpolation in knex.raw() with parameterized placeholders(?) and value arrays.

# AI usage

* [x] I did not use any AI

### DIFF
--- a/src/server/helpers/achievement.js
+++ b/src/server/helpers/achievement.js
@@ -371,11 +371,10 @@ async function processSeriesCreator(orm, editorId) {
 
 async function processSprinter(orm, editorId) {
 	const {bookshelf} = orm;
-	const rawSql =
-		`SELECT * from bookbrainz.revision WHERE author_id=${editorId} \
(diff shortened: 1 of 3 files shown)


MonkeyDo:
I am not convinced this makes the code any more readable or robust (looking for example at having to escape apostorphes), and as you mention there is no risk of SQL injection in these cases.
For these reasons, I am closing this PR, although the code did look fine.

PR #1287, opened 2026-04-10
fix: featured badge slot behavior and UI interactions

shalini-bhandari:
### Problem
The Achievements page had a few usability issues:

1. The currently selected featured badges were not shown properly when the page first loaded
2. The same badge could be selected in multiple slots
3. Dragging an already selected badge into another slot did not move it
4. The update section UI felt a bit unclear and awkward

Ticket : [BB-425](https://tickets.metabrainz.org/browse/BB-425)

### Solution

1. Fixed the initial load behavior so the already selected badges are shown correctly when the achievements page opens
2. Updated the badge slot logic so the same badge cannot appear in multiple slots
3. Improved the drag-and-drop behavior so if a selected badge is dropped into another slot, it moves from the old slot to the new one
4. slot handling logic was refactored


MonkeyDo:
AI disclosure was removed from template, and code and communications show clear signs of AI usage. Closing PR.

PR #1290, opened 2026-04-15
BB-895: fix: filter recently used to match search text instead of blocking results

atharvsp02:
# Problem

The recently used section in entity search dropdowns shows all recently used items even when the user is typing a search query. This pushes the actual search results down and the user has to scroll to see them.

Ticket: [BB-895](https://tickets.metabrainz.org/browse/BB-895)

# Solution

Filter the recently used section to only show items that match what the user is typing. If nothing matches, only search results are shown without the recently used header. Also removed `cacheOptions` because it was returning old search results and preventing recently used from showing up when searching the same text again.

# Screenshot
### Before
image

### After

Recently used shows when search box is empty
image

Only matching recently used entities shown when typing
image
# AI usage

* [x] I did not use any AI

#### If you did use AI:

# Action

No action needed.


MonkeyDo (approved):
Thanks for the quick follow-up!

PR #1160, opened 2025-03-23
BB-841: Converted Promises to async/await in component pages

Ankit-Matth:
### Problem  
[BB-841](https://tickets.metabrainz.org/browse/BB-841) :  
Several files in the `src/client/components/pages` directory used Promises for asynchronous code, making it harder to read, maintain, and debug.

### Solution  
Resolved **BB-841** by refactoring all Promises in `src/client/components/pages` to use `async/await` syntax, improving readability and maintainability.

### Areas of Impact  
The following files were updated:

- [`collection.js`](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/collection.js)  
- [`revision.js`](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/revision.js)  
- [`add-entity-to-collection-modal.js`](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/add-entity-to-collection-modal.js) 
- [`add-to-collection-modal.js`](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/add-to-collection-modal.js)  
- [`delete-or-remove-collaboration-modal.js`](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/parts/delete-or-remove-collaboration-modal.js)
- [`wikipedia-extract.tsx`](https://github.com/metabrainz/bookbrainz-site/blob/master/src/client/components/pages/entities/wikipedia-extract.tsx)

PR #1293, opened 2026-04-18
BB-841 : fix error handling and refactored handlesubmit from promise .then/catch  to async/await with try catch

garvthakre:
# Problem

 BB-841 : handleSubmit in revision.js used an old Promise .then()/.catch() chain. The .catch block incorrectly used res as if it were a response object, but a thrown Promise error is an Error object . so res.body would crash.this was mentioned by @MonkeyDo  in the PR #1160 (it is closed for now because the fork was deleted)

# Solution

Converted handleSubmit to async/await with try/catch. Error is now correctly extracted via err?.response?.body?.error || err?.message || err as recommended in the superagent docs and per @MonkeyDo's review on that PR.

# AI usage

* [x] I did not use any AI

 

# Action

 No special action required for this .

### DIFF
--- a/src/client/components/pages/revision.js
+++ b/src/client/components/pages/revision.js
@@ -163,21 +163,20 @@ class RevisionPage extends React.Component {
 		this.handleSubmit = this.handleSubmit.bind(this);
 	}
 
-	handleSubmit(event) {
+	async handleSubmit(event) {
 		event.preventDefault();
 		const data = {
 			note: this.noteInput.value
 		};
-		request.post(`/revision/${this.props.revision.id}/note`)
-			.send(data)
-			.then(() => {
-				location.reload();
-			})
-			.catch((res) => {
-				// TODO: Add proper error handling.
-				const {error} = res.body;
-				return error;
-			});
+		try {
+			await request.post(`/revision/${this.props.revision.id}/note`).send(data);
+			location.reload();
+		}
+		catch (err) {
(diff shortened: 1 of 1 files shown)


garvthakre:
hey @MonkeyDo , its a small refactor change can u have a look at this .

src/client/components/pages/revision.js
```diff
@@ -163,21 +163,20 @@ class RevisionPage extends React.Component {
...
-				return error;
-			});
+		try {
+			await request.post(`/revision/${this.props.revision.id}/note`).send(data);
+			location.reload();
+		}
+		catch (err) {
+			const error = err?.response?.body?.error || err?.message || err;
+			// eslint-disable-next-line no-console
+			console.error('Failed to submit note:', error);
```
MonkeyDo, line 178: Something changed here that you need to investigate.

Previously we were returning the error in the catch block, possibly so that something else down the line can handle it.
Now it is just being written to the console (where we would usually use a logger instead, [see examples](https://github.com/search?q=repo%3Ametabrainz%2Fbookbrainz-site%20log.error&type=code)).

You'll want to artificially trigger an error (for example throwing an error object of the right shape in the try block) and compare what happens before and after the changes in this PR, to ensure it is working as expected.

When you are refactoring code like this one-to-one, it's important to ensure that the mechanism remains unchanged, or that the reason for change is documented and tested.

MonkeyDo (changes requested):

PR #1294, opened 2026-04-30
BB-894: fix: preserve search field text on blur using controlled input

atharvsp02:
# Problem

The entity search field clears the typed text when clicking away or changing tab. This happens because the component uses `onBlurResetsInput={false}` which is a react-select v1 prop that stopped working after the v4 upgrade.

# Solution

Removed `onBlurResetsInput` and switched to controlling the input value manually with `inputValue` and `onInputChange` so blur doesn't clear the text. Also saving the search results in state so they show up again when you click back on the field.

# Screen Recordings
### Before
Clicking outside clears the typed text from the search field

https://github.com/user-attachments/assets/36154fd8-9f42-40c4-997f-d7d220b3c9cb

### After
Typed text stays in the search field and results show up again when clicking back

https://github.com/user-attachments/assets/46af3dc3-7942-43c5-977a-f5b5712e91b0

# AI usage

* [x] I did not use any AI

#### If you did use AI:

# Action

No action needed.

Let me know if you have any suggestions or changes.

### DIFF
--- a/src/client/entity-editor/common/entity-search-field-option.tsx
+++ b/src/client/entity-editor/common/entity-search-field-option.tsx
@@ -65,7 +65,13 @@ class EntitySearchFieldOption extends React.Component<Props> {
 		this.fetchOptions = this.fetchOptions.bind(this);
 		this.isArea = this.isArea.bind(this);
 		this.entityToOption = this.entityToOption.bind(this);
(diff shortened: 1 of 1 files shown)


MonkeyDo (approved):
Thank you for the good fix!

PR #1295, opened 2026-05-06
BB-841: Convert Promise chains to async/await in collection components

garvthakre:
# Problem
[BB-841](https://tickets.metabrainz.org/browse/BB-841): Several collection-related components in src/client/components/pages/ used Promise .then()/.catch() chains for asynchronous code, making the logic harder to read and maintain.

Also, some error handlers used the second argument of .then(success, error) instead of a dedicated catch block, and the handleAddToNewCollection function had nested .then() calls inside another .then(), making the control flow difficult to follow .

# Solution

Converted all Promise .then()/.catch() chains to async/await with try/catch in the following files:

1.src/client/components/pages/collection.js - handleRemoveEntities
2.src/client/components/pages/parts/add-entity-to-collection-modal.js - handleSubmit
3.src/client/components/pages/parts/add-to-collection-modal.js - handleAddToNewCollection (also flattened nested .then() chains into two sequential await calls)
4.src/client/components/pages/parts/delete-or-remove-collaboration-modal.js - handleSubmit

# AI usage

* [x] I did not use any AI

#### If you did use AI:

# Action

No special action needed

### DIFF
--- a/src/client/components/pages/collection.js
+++ b/src/client/components/pages/collection.js
@@ -140,28 +140,31 @@ class CollectionPage extends React.Component {
 		});
 	}
 
-	handleRemoveEntities() {
+	async handleRemoveEntities() {
 		if (this.state.selectedEntities.length) {
 			const bbids = this.state.selectedEntities;
 			const submissionUrl = `/collection/${this.props.collection.id}/remove`;
-			request.post(submissionUrl)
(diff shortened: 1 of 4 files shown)


MonkeyDo (approved):
Thanks!

PR #1305, opened 2026-06-01
feat(i18n): add isomorphic i18next initializer with tests

garvthakre:
# Problem

To support translations across the server-side rendering (SSR) and client-side rendering stages, This PR sets up the basic packages, the core isomorphic i18next initializer function, and its unit tests.

### Changes:
- **`package.json` / `yarn.lock`**: Added dependencies `i18next`, `react-i18next`, `i18next-browser-languagedetector`, and `i18next-http-backend`.
- **`src/common/i18n/i18n.ts`**: Implemented `createI18n` factory logic.
- **`test/src/common/i18n.js`**: Created test suite covering the initializer.

# Solution
1.  Added `i18next` and its helper packages (`react-i18next`, `i18next-browser-languagedetector`, `i18next-http-backend`)  .
2.  Added 5 unit tests in `test/src/common/i18n.js` to cover default locale, custom locales, pre-loaded resources, missing keys, and synchronous loading.
3. Created `src/common/i18n/i18n.ts` with `createI18n` to initialize i18next. It creates a fresh instance per call (important so server requests don't leak locale) and disables `initImmediate` for SSR compatibility.
 

# AI usage
 
* [ ] I did not use any AI
  => To debug and some research for testing is done by  the help of Ai  .
#### If you did use AI:

# Action
 None .Tests run completely in-memory.


garvthakre:
hey @MonkeyDo , have a look at this .let me know if u need any changes .

src/common/i18n/i18n.ts
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2026  Garv Thakre
```
MonkeyDo: I know we have a lot of old copyright in existing files, but it should be this:
 ```suggestion
 * Copyright (C) 2026  MetaBrainz Foundation
```
Applies to the other file as well.

File: test/src/common/i18n.js [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+		expect(i18n.t('nonexistent.key', {ns: 'common'})).to.equal('nonexistent.key');
+	});
+
+	it('should initialize synchronously for SSR compatibility', () => {
+		const i18n = createI18n('en', {en: {common: {'sync.test': 'Works'}}});
+		// If initImmediate were true (default), translations would load
+		// asynchronously and t() would return the raw key 'sync.test'
+		// because renderToString() is synchronous and cannot await
+		expect(i18n.t('sync.test', {ns: 'common'})).to.equal('Works');
+	});
```
MonkeyDo, line 85: I don't know that this actually tests properly, and basically looks the same as "should use pre-loaded resources for translation".
It's usually not a good idea to test for the absence of something, unless you first assert the presence of it.

So in this case, if there is a way to replicate the mechanism you describe in the comment, let's add that to compare. If there isn't, I'm not sure what we're testing exactly.

MonkeyDo (changes requested):
Nice to see the first PR rolling in :)

MonkeyDo (approved):
Looks good!
First merge of the project :)

garvthakre:
finally 😊 btw what will u do about the build error that is showing here

MonkeyDo:
> finally 😊 btw what will u do about the build error that is showing here

Looking at that now, it's unrelated to your code changes.

PR #1306, opened 2026-06-02
Update OAuth2 provider to use MetaBrainz new-oauth2

MonkeyDo:
Replaces the custom MusicBrainz oauth strategy we had in a separate repo (https://github.com/BenOckmore/passport-musicbrainz-oauth2) to use the standard OAuth2 passport strategy, with small custom code for the profile introspection.

Still need to save the access and refresh tokens in the DB, coming in the next commit.


amCap1712 (approved):

MonkeyDo:
Tested on test.bookbrainz and working fine

PR #1309, opened 2026-06-05
 feat(i18n): implement server-side i18n pipeline and layout integration

garvthakre:
# Problem

BookBrainz hardcodes all UI strings in English. We need a server-side pipeline to detect the user's language, load matching translation JSON files, and pass them to the react layout during server rendering and client hydration.

# Solution

1. Middleware (middleware.ts) : Added i18nMiddleware to scan `public/locales/` dynamically, check the bb_lang cookie or Accept-Language headers, and load the JSON files to res.locals.
2. Props (`props.ts`): Merged the detected locale and resources into page props under props.i18n.
3. HTML & Layout (`target.js & layout.js`): Passed the locale to render a dynamic <html lang=" "> tag, and wrapped the main Layout component in I18nextProvider.
4. Docker & Tests : Copied public/ into the Dockerfile so the build has access to locale assets. Added a Mocha test  in `test/src/server/helpers/middleware.js` verifying the fallbacks and default behaviors.

## How it Works :
1. Detect: The server reads the language cookie or browser header.
2. Load: It reads matching translation files (like common.json) from public/locales/.
3. Pass: The server embeds these translations in the HTML page source inside the page props (props.i18n).
4. Hydrate: The browser reads those props to initialize the translations, preventing page flickering.

# AI usage
 
* [ ] I did not use any AI

Details : used for debugging and some research for tests 
#### If you did use AI:

# Action

=> Set up the BookBrainz translation components on translations.metabrainz.org pointing to `public/locales/en/` once this merges.


garvthakre:
hey @MonkeyDo , so finally made the second PR , and this was built on top of the '[origin/i18n-initializer'](https://github.com/metabrainz/bookbrainz-site/pull/1309/commits/3c578df33ad67b3aa272c6c64e03ebcaa19d4e8f)'  branch and  so to solve the linting error , i merged the branch  and also the old commit will gone after that PR gets merged .
i think the things gets abit messy here , but the commits are clean , i think that will help you to review easily.let me know your opinion.

test/src/server/helpers/middleware.js
```diff
@@ -0,0 +1,119 @@
...
+		expect(nextCalled).to.be.true;
+	});
+
+	it('should set i18nResources as a plain object', async () => {
+		const req = createMockReq();
+		const res = createMockRes();
+		let nextCalled = false;
+
+		function next() {
+			nextCalled = true;
```
MonkeyDo, line 96: For this test and the next one, you're not checking the nextcalledf is true

docker-compose.api.yml
```diff
@@ -26,7 +26,7 @@ services:
     image: postgres:18
     user: postgres
     ports:
-      - "127.0.0.1:5432:5432"
+      - "127.0.0.1:5434:5432"
```
MonkeyDo: Why were these changes to the docker-compose required?
You shouldn't have to change them, and in the context of this PR certainly not the api docker-compose file.
I would revert both of these files.

MonkeyDo (changes requested):
Looking good!
Some minor things, but apart from that should be ready to merge

docker-compose.api.yml
garvthakre: really sorry for these one i change the port for my local use and forget to change it back .

test/src/server/helpers/middleware.js
garvthakre: added nextcalled have a look and let me know if it is alright

docker-compose.api.yml
MonkeyDo: The two files have not been reverted

garvthakre: done ;)

MonkeyDo (approved):
Good to go, thank you!

PR #1312, opened 2026-06-13
feat(i18n): migrate name-section components to react-i18next

garvthakre:
# Problem
This is part of the BookBrainz internationalization (i18n) project. In this PR, migrate the `name-section` components of the entity editor away from hardcoded English strings to use `react-i18next` for localization.
 
# Solution

* Migrated name section components (`name-field`, `sort-name-field`, `language-field`, `disambiguation-field`, and `name-section`) to use the `useTranslation` hook or `withTranslation` HOC.
* Extracted and updated the translation keys in `public/locales/en/entityEditor.json`.
* Created empty locale JSON stub files (`entities.json`, `errors.json`, `pages.json` for both `en` and `fr`) to prevent console warnings/404 fetch errors in the browser.
* Updated the server-side i18n middleware to load all namespaces with an English fallback.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

note : use ai to do repetitive tasks
# Action
None


src/client/entity-editor/name-section/name-section.js
```diff
@@ -134,17 +135,17 @@ class NameSection extends React.Component {
...
 					disambiguationDefaultValue
 				) ?
 					<Alert variant="warning">
-					We found the following&nbsp;
-						{_.startCase(entityType)}{exactMatches.length > 1 ? 's' : ''} with
-					exactly the same name or alias:
-						<br/><small className="text-muted">Click on a name to open it (Ctrl/Cmd + click to open in a new tab)</small>
+						{translate('nameSection.duplicateWarning')}&nbsp;
+						{_.startCase(entityType)}{exactMatches.length > 1 ? 's' : ''} {translate('nameSection.duplicateWithSameName')}
+						<br/><small className="text-muted">{translate('nameSection.clickToOpen')}</small>
```
MonkeyDo: Would this be a case for interpolation like you did for the  duplicateSuggestion below?
Something like this maybe ? - pseudocode but you get the idea

```ts
let formattedEntityName = _.startCase(entityType)
if(exactMatches.length > 1){
formattedEntityName += 's'
}

"We found the following {{entityType}} with exactly the same name or alias:"

{translate('nameSection.duplicateWarning', { entityType: formattedEntityName })}
```


Or even starting to use this feature for the pluralization?
https://www.i18next.com/translation-function/plurals

MonkeyDo: Or mayyybe even combining pluralizing with some nesting?
https://www.i18next.com/translation-function/nesting

Something like this I suppose? (Untested)

```json
  "nameSection.duplicateWarning": "We found the following $t({{entityType}}, {\"count\": $t(count)}) with exactly the same name or alias:",
  
  "author_one": "author",
  "author_other": "authors",
  
  "edition_one": "edition",
  "edition_other": "editions",
  [...]

 ```
 
 
 ```ts
{translate('nameSection.duplicateWarning', { count: exactMatches.length, entityType: entityType })}
 ```

MonkeyDo:
Don't know if I'm testing it wrong, but with locale and currentLocale set to FR, the translations don't seme to be working for the name section:
image

Alias editor still correctly shows french content.

If it's not user-error on my part, and if you are testing these PRs locally, you need to test them more thoroughly

garvthakre:
> Don't know if I'm testing it wrong, but with locale and currentLocale set to FR, the translations don't seme to be working for the name section: image
> 
> Alias editor still correctly shows french content.
> 
> If it's not user-error on my part, and if you are testing these PRs locally, you need to test them more thoroughly

the most probable reason is that there is no fr tanslations and the language is going back to its default english language (basically if the tanslations is not present in the json then it sets back to english and shows this ) and the way i use to test is that i just put a,b,c ..... in  fr json and any  word  with same character count to check it ,is there any other way that i should use to test it too .

MonkeyDo:
> the most probable reason is that there is no fr tanslations and the language is going back to its default english language (basically if the tanslations is not present in the json then it sets back to english and shows this ) and the way i use to test is that i just put a,b,c ..... in fr json and any word with same character count to check it ,is there any other way that i should use to test it too .

Ah yes, I see public/locales/en/entityEditor.json has a bunch of entries that public/locales/fr/entityEditor.json does not have.

src/client/entity-editor/name-section/name-section.js
garvthakre: i have refactored it to use i18next interpolation and pluralization with _one/_other suffixes .i think this  would works for this one.

MonkeyDo (approved):
Name section looks good! 👍 

I'd like us to focus next on CI (looking at #1315) and integration with weblate, so I can start testing the translator workflow with those empty string in french translations.

In any case, that did give us a good test for fallback when there is no translation available, I suppose.

garvthakre:
> Name section looks good! 👍
> 
> I'd like us to focus next on CI (looking at #1315) and integration with weblate, so I can start testing the translator workflow with those empty string in french translations.
> 
> In any case, that did give us a good test for fallback when there is no translation available, I suppose.

Thanks for the merge ,and yeah  i started working at ( https://github.com/metabrainz/bookbrainz-site/pull/1315 ),i will push the updated code soonn.

PR #1314, opened 2026-06-15
feat(i18n): Migrate Button Bar sub-components to react-i18next

garvthakre:
# Problem

This is part of the BookBrainz internationalization (i18n) project. In this PR,migrate the Button Bar components (`AliasButton` and `IdentifierButton`) of the entity editor away from hardcoded English strings to use `react-i18next` for localization.
 

# Solution

* Migrated the button bar sub-components (`alias-button.js` and `identifier-button.js`) to use `react-i18next` with the `useTranslation` hook.
* Refactored the dynamic label layout of the buttons using an IIFE `const` pattern to satisfy strict ESLint and `@typescript-eslint` scoping rules.
* Updated the translation keys in `public/locales/en/entityEditor.json` using the standard dot notation.

* [x] I have run the code and manually tested the changes

# AI usage

:beginner: If you used AI at all, you [must disclose it](https://github.com/metabrainz/guidelines/?tab=readme-ov-file#ai-use-policy), and what for, such as:
>Used Copilot when writing the tests I added to `testfile.js`

* [x] I did not use any AI
* [ ] I have used AI in this PR (add more details below)

#### If you did use AI:
* [ ] I used AI tools for communication
* [ ] I used AI tools for coding
* [ ] I understand all the changes made in this PR

# Action

None

PR #1315, opened 2026-06-16
ci: add workflow to check for uncommitted translation changes

garvthakre:
# Problem

This is part of the BookBrainz internationalization (i18n) project.  

When developers modify text in React components, they might forget to run `yarn run parse-i18n` locally to update the translation catalog JSON files. This results in the locale files being out of sync with the actual codebase.

# Solution

* Created a new GitHub Actions workflow at `.github/workflows/i18n-extract.yml` to run key extraction on pushes to `master`/`i18n` and on all pull requests.
* The workflow runs key extraction and uses `git diff --exit-code public/locales/` to fail the build if any translation files are modified by the parser, alerting the developer that their local JSON catalogs are out of sync.

# AI usage

* [x] I did not use any AI

#### If you did use AI:

# Action
None


MonkeyDo:
What do you think would be a good way to test this new workflow?
I can see it ran successfully, but I'd like to see it fail for confirmation.

You could push a commit to this branch to make it fail, and once we are in agreement revert it?

MonkeyDo:
I mentioned it in the BB chat, but for posterity here as well:
The i18next-parser library is deprecated since late February this year, and replaced with i18next-cli.

It might be worth replacing it straight away in this PR, considering the focus on that particular tool.

They suggest it could be as easy as running `npx i18next-cli migrate-config`, so worth a try in any case

garvthakre:
Hey @MonkeyDo , so i replaced the parser with cli ,and the working is a bit different in this bcos cli parses all TS files in TSX mode, where the old angle-bracket typecast syntax is invalid in wikimedia.ts , so i refactored it too. and now i am adding a dummy key in my next PR so the workflow gets failed .

src/client/entity-editor/submission-section/submission-section.js
MonkeyDo, line 1: Great, thanks for the demonstration, that looks good.
You can revert this commit 👍

garvthakre: done.

MonkeyDo (approved):
Ready to merge once #1314 is merged

garvthakre:
yupp wokring on it..

src/server/helpers/wikimedia.ts
```diff
@@ -96,11 +96,11 @@ export async function getAvailableWikipediaArticles(wikidataId: string, {
 	const articles = Object.values(item.sitelinks)
 		// only keep Wikipedia pages
 		.filter((link) => link.site.endsWith('wiki'))
-		.map((page) => <WikipediaArticle>({
+		.map((page) => ({
 			// drop project suffix
 			language: page.site.replace(/wiki$/, ''),
 			title: page.title
-		}));
+		} as wikipedia));
```
MonkeyDo: Don't know where that came from, but this type does not exist.
```suggestion
		} as WikipediaArticle));
```

MonkeyDo (commented):
Small details, ready to merge after I commit small fixes.

PR #1322, opened 2026-06-19
i18n: Migrate Author and Publisher Editor Sections

garvthakre:
# Problem

This PR migrates the Author and Publisher editor section components (including their merge views) to support internationalization (i18n) using `react-i18next` and JSON locale files.

# Solution

1. **Hook Pattern (`useTranslation`)**:
   - Migrated `PublisherSection` and `PublisherSectionMerge` using the `useTranslation('entityEditor')` hook, since their translation strings are entirely defined and rendered directly inside their JSX views.
2. **HOC Pattern (`withTranslation`)**:
   - Migrated `AuthorSection` and `AuthorSectionMerge` using the `withTranslation('entityEditor')` High-Order Component (HOC) wrapper.
3. **Shared Helper (`utils.tsx`)**:
   - Updated `labelsForAuthor` in `src/client/helpers/utils.tsx` to support the passed `translate` function, resolving labels from the `entityEditor` namespace with default English fallbacks.
4. **Locale files**:
   - Added translation keys in `public/locales/en/entityEditor.json` and `public/locales/en/common.json` with correct default English strings.

# AI usage

* [x] I did not use any AI

#### If you did use AI:

# Action

None


public/locales/en/entityEditor.json
```diff
@@ -14,6 +14,26 @@
...
+  "authorSection.endDateLabel.person": "Date of death",
+  "authorSection.endedLabel.group": "Dissolved?",
+  "authorSection.endedLabel.person": "Died?",
+  "authorSection.genderLabel": "Gender",
+  "authorSection.heading": "What else do you know about the Author?",
+  "authorSection.Merge.genderLabel": "Gender",
+  "authorSection.Merge.typeLabel": "Type",
+  "authorSection.typeLabel": "Type",
+  "authorSectionMerge.genderLabel": "Gender",
+  "authorSectionMerge.typeLabel": "Type",
```
MonkeyDo: These look like they're missing a dot, but also duplicate the correct one a couple of lines above them

src/client/entity-editor/publisher-section/publisher-section.tsx
```diff
@@ -139,13 +141,12 @@ function PublisherSection({
 		<div>
 			{!isUnifiedForm && heading}
 			<p className="text-muted">
-				All fields optional — leave something blank if you don&rsquo;t
-				know it
+				{translate('publisherSection.allFieldsOptional')}
```
MonkeyDo: I think this should be more generic and not attached to the publisherSection.
It is used in other pages section as well (like the author section above, but also work section and edition group sectionm), so let's not duplicate it.

src/client/entity-editor/publisher-section/publisher-section.tsx
```diff
@@ -123,14 +124,15 @@ function PublisherSection({
...
+	const {t: translate} = useTranslation('entityEditor');
 	const publisherTypesForDisplay = publisherTypes.map((type) => ({
 		label: type.label,
 		value: type.id
 	}));
 	const typeOption = publisherTypesForDisplay.filter((el) => el.value === typeValue);
 	const {isValid: isValidBeginDate, errorMessage: errorMessageBeginDate} = validatePublisherSectionBeginDate(beginDateValue);
 	const {isValid: isValidEndDate, errorMessage: errorMessageEndDate} = validatePublisherSectionEndDate(beginDateValue, endDateValue, endedChecked);
-	const heading = <h2>What else do you know about the Publisher?</h2>;
+	const heading = <h2>{translate('publisherSection.heading')}</h2>;
```
MonkeyDo: Same comment here, let's avoid duplication by using `"What else do you know about the {entity}"` and reuse it across components.
Not sure what the key name should be, but we'll want to avoid the generic "heading" in favor of something like "whatElseHeading" maybe ? Open to better ideas...

This might be too much for this instance, but we could also start using [custom formatters](https://www.i18next.com/translation-function/formatting#adding-custom-format-function)  to allow passing the lowercase entity name and have it be capitalized with `"What else do you know about the {{entity, capitalize}}"`

src/client/entity-editor/publisher-section/publisher-section-merge.tsx
```diff
@@ -184,7 +191,7 @@ function PublisherSectionMerge({
 					currentValue={formattedEndDateValue}
 					error={!isValidEndDate}
 					errorMessage={errorMessageEndDate}
-					label="Date Dissolved"
+					label={translate('publisherSection.dateDissolved')}
```
MonkeyDo, line 194: I will note here that we will need the same keys for the display page for each entity type.
Basically, taking "Date Dissolved" as an example, it also appears in the heading of the display page: https://bookbrainz.org/publisher/27d0cee1-88d5-4aa5-af3b-a91cb13d7e74

I think maybe we will want to reuse the same key to avoid duplication, and it is probably the right time to think about how they are organized ("entityEditor" namespace and "publisherSection" key name suggests use only in the editing pages).
We need to figure out the right place to put these reused translation.

I'm thinking also for example about "Area" above. This is used in most entity edit and display pages, making it perhaps more of a "common" namespace key.
What would be your suggestion?

MonkeyDo (changes requested):
Now that we are continuing to add more keys, I want to review the organization a bit to reduce duplication as much as possible. 
Most of my comments below are on that topic, I want us to find the right balance before going forward with adding more duplicate keys and making it harder for translators.

public/locales/en/entityEditor.json
garvthakre: they are of different pages but i will use a common term for both now .

src/client/entity-editor/publisher-section/publisher-section-merge.tsx
garvthakre: Hi @MonkeyDo Based on your comments, here is what I think we can do:

1. I will search for duplicate strings across the project. If the same string appears at least 3 times, I will move it to `common.json`.
2. If we have duplicate strings within the same section folder (for example, in `author-section`), I'll  define a single, generic key name and reuse it in both places (rather than having separate keys for the edit and merge views).

What do you think of this approach?

src/client/entity-editor/publisher-section/publisher-section.tsx
garvthakre: okee yeah i guess this is also a case and i should keep this one maybe in common and yeah let me think of a  more generic name for this one .

src/client/entity-editor/publisher-section/publisher-section-merge.tsx
garvthakre: hey @MonkeyDo , i made the changes following the same approach, if this sounds good, i will push the commit .

MonkeyDo (approved):
Looking good, thank you!

PR #1325, opened 2026-06-21
fix(auth): Use epoch timestamp as default for active_at column

srajang1805:
Changes the default value of the editor.active_at column from timezone('UTC'::text, now()) to TIMESTAMP 'epoch'.

Previously, new editors were created with active_at defaulting to the current timestamp, implying they were "active" at the moment of account creation. This was incorrect — a brand new user has no activity yet. Using epoch (1970-01-01 00:00:00 UTC) as the default provides a proper sentinel value, so the "last active" display only shows a meaningful date after the user has performed their first action.

Changes:

sql/migrations/2026-06-active-at-epoch/up.sql — new migration
sql/migrations/2026-06-active-at-epoch/down.sql — rollback migration
sql/schemas/bookbrainz.sql — updated schema default

### DIFF
--- a/sql/migrations/2026-06-active-at-epoch/down.sql
+++ b/sql/migrations/2026-06-active-at-epoch/down.sql
@@ -1,0 +1,2 @@
+ALTER TABLE bookbrainz.editor
+	ALTER COLUMN active_at SET DEFAULT timezone('UTC'::TEXT, now());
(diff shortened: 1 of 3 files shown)


MonkeyDo:
Hello, why are you opening PRs with my commits?

srajang1805:
@MonkeyDo Apologies for the confusion. The branch was created from an upstream branch that already contained a commit authored by you, and I mistakenly opened the PR from it. It was an oversight on my part. I'm sorry for the mix-up and any confusion it caused.

PR #1326, opened 2026-06-21
 fix: Improve error handling in entity loader and collections

srajang1805:
**catch-entity-loader-errors** — Improves error handling across three areas:

### `src/server/helpers/middleware.ts`
Properly classifies entity loader errors — `NotFoundError` from the ORM model is caught and returned as a `404`, while other errors (DB failures, etc.) are wrapped in a `SiteError` with the BBID and entity type for debugging.

### `src/common/helpers/error.js`
`SiteError` constructor now accepts an `Error` object or a string as `detailedMessage`, preserving the original error context.

### `src/server/routes/collections.js`
- Removed unnecessary `// eslint-disable-next-line consistent-return` comments
- Replaced `throw` with `return next(...)` for `BadRequestError`
- Added explicit `return` before `res.send()`
- Changed `/collections` endpoint to use `error.sendErrorAsJSON()` instead of passing errors to `next()`


MonkeyDo:
Hello, why are you opening PRs with my commits?

PR #1332, opened 2026-07-02
i18n: Migrate editor and registration page components to react-i18next

garvthakre:
# Problem
This PR migrates the registration and editor-related page components to react-i18next and replaces hardcoded UI strings with localized text from the shared locale catalog.

# Solution

- `registration-details.js`: Registration details form text and related labels.
- `collections-table.js`: Collection table UI copy.
- `editor-achievements.js`: Achievements page labels and messages.
- `editor-profile.js`: Profile page content and action text.
- `revisions-table.js`: Revisions table headings and related strings.

# AI usage

* [ ] I did not use any AI
note: to perform some repetitive tasks and process 
#### If you did use AI:

# Action
None


src/client/components/pages/parts/collections-table.js
```diff
@@ -128,26 +131,26 @@ class CollectionsTable extends React.Component {
...
 							<thead>
 								<tr>
-									<th width="16%">Name</th>
-									<th width="33%">Description</th>
-									<th width="16%">Entity Type</th>
-									<th width="16%">Entities</th>
+									<th width="16%">{translate('pages:collections.headerName')}</th>
+									<th width="33%">{translate('pages:collections.headerDescription')}</th>
+									<th width="16%">{translate('pages:collections.headerEntityType')}</th>
+									<th width="16%">{translate('pages:collections.headerEntities')}</th>
```
MonkeyDo: These all look standard to me. At the very least, "name" should use the existing key.
If the others don't already exist in the common namespace that is where they should live IMO as they are strings we are likely to reuse.

src/client/components/pages/parts/revisions-table.js
```diff
@@ -154,7 +157,7 @@ RevisionsTable.defaultProps = {
 	showEntities: false,
 	showRevisionEditor: false,
 	showRevisionNote: false,
-	tableHeading: 'Recent Activity'
+	tableHeading: null
```
MonkeyDo, line 160: Why not use `translate('pages:revisions.recentActivity')` here? Do wwe not have access to the translate function?
If not, let's remove this prop entirely and use the translated string directly in the h1 component

MonkeyDo (commented):
Some preliminary comments, I will review again after merging the previous 2 PRs, as it is getting  hard to review

src/client/components/pages/parts/revisions-table.js
```diff
@@ -23,17 +23,19 @@ import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
...
 
 function RevisionsTable(props) {
-	const {results, showEntities, showRevisionNote, showRevisionEditor, tableHeading} = props;
+	const {results, showEntities, showRevisionNote, showRevisionEditor} = props;
+	const {t: translate} = useTranslation('pages');
 	return (
 		<div>
 			<div>
-				<h1 className="text-center">{tableHeading}</h1>
+				<h1 className="text-center">{translate('pages:revisions.recentActivity')}</h1>
```
MonkeyDo: This is incorrect. The RevisionsTable component is used in a few places, not always with teh same heading.

src/client/components/pages/editor-revision.js
```diff
@@ -79,7 +79,7 @@ EditorRevisionPage.defaultProps = {
 	showRevisionEditor: false,
 	showRevisionNote: true,
 	size: 20,
-	tableHeading: 'Recent Activity'
+	tableHeading: null
```
MonkeyDo, line 87: Here's the other incorrect part, this should pass the translated string

garvthakre: oopsie cant believe i did the same mistake again .

MonkeyDo: I thinkk I also misunderstood at some point and suggested removing tableHeading entirely from RevisionsTable without realizing it is reused in a few pages.

src/client/components/pages/parts/revisions-table.js
MonkeyDo: I was totally wrong here!
This component is used in a few places, so the tableHeading is required

src/client/components/pages/editor-revision.js
MonkeyDo: Some more thoughts on this: you will not have access to the translation fucntion in the defaultProps, so getting that default value translated is tricky.

We have a couple of options:
1. Set up EditorRevisionPage to be translatable, change the defaultProp.tableHeading to `null` and set a default value in the render function instead, something like `const tableHeading = this.props.tableHeading ?? translate('pages:revisions.recentActivity'); `

2. More complicated, probably not as good: refactor the component to be a functional component rather than a class component, which comes with the removal of `defaultProps` and having the default value defined in the function definition instead:
`function EditorRevisionPage({[...other props here], tableHeading='Recent Activity' }) { [...rest of the component] }`

garvthakre: got it ! so i restored the `tableheading ` as a reusable default to  `EditorRevisionPage`'s render method and also update the othe rpages to pass their translated heading explicitly. and entityRevision was left unchanged since its already provides its own custom heading.

garvthakre:
hey @MonkeyDo ,can u  have a look at this one ..

MonkeyDo (approved):
Did a tiny bit of cleanup, good to go now ! 🚀

PR #1353, opened 2026-08-04
Review database dumps

MonkeyDo:
# Problem
1. We were leaking OAuth tokens in the DB dumps :facepalm: both critiquebrainz ~and metabrainz oauth tokens~ (the new MeB tokens were not dumped, see point 3)
2. We were not dumping collections, as we needed some changes to dump only public ones (see also #1187 and #513 ) 
3. Dumps seem to have been broken since May

# Solution

We needed to implement a few changes to the dump generation:
1. make the table denylist into an allowlist, listing each table that we wanmt to see exported
2. for some tables, we want to filter the data and only export what should be public (no oauth tokens, no private collections)

We have to create dumps in multiple parts, to allow for exporting:
- the schema and other DB objects (not exported when using `--table`, the allowlist mechanism)
  - Don't export `_editor_entity_visits` table
- the data we need using `--table` allowlist
- the filtered data (excluded from the previous section with `--exclude-table-data`)
  - editor table without OAuth tokens
  - user_collection and associated tables, filtered for public collections only
- finally the triggers, indices, constraints, etc.

I also removed the uuid-ossp extension, as UUID generation is now supported natively by postgres. Restoring the dumps was problematic with that extension, as the extension needed to be created before the import.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

Used Codex to help with the dump creation script, specifically how to break it down into multiple parts
Manually tested everything from A to Z, both locally and in production to create a new dump with no private data

# Action

* [x] Create a new database dump without private data, synced to public FTP 
* [ ] Create a full dump with all the data, synced to our private backup server

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,6 +81,7 @@ RUN touch /etc/service/cron/down
 
 ADD ./docker/services/cron/crontab /etc/cron.d/bookbrainz
 RUN chmod 0644 /etc/cron.d/bookbrainz && crontab -u bookbrainz /etc/cron.d/bookbrainz
+RUN mkdir -p /tmp/logs
 
 # Build JS project and assets
 RUN ["yarn", "run", "build"]
--- a/docker/services/cron/crontab
+++ b/docker/services/cron/crontab
@@ -1,7 +1,4 @@
 # m h d m w
 
-# Database backup creation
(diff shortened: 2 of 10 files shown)


MonkeyDo:
@amCap1712 I would love your review on this, since you've also been looking at bhte LB dumps

PR #1360, opened 2026-08-07
i18n : refactoring the namespace json to translation json

garvthakre:
# Problem

Merge the  5 separate translation files (`common.json`, `entityEditor.json`, `pages.json`, `entities.json`, `errors.json`) in one combined translation.json.
 
# Solution

1. **Merged Translation Files**:
   * Consolidated all flat localization files into a single nested `translation.json` structure per locale (`en` and `fr`).
   * Cleaned up the old flat namespace JSON files.

2. **Configuration Updates**:
   * Updated `i18next.config.js` to target `translation.json` with `keySeparator: '.'` and `nsSeparator: false`.
   * Updated runtime config in `src/common/i18n/i18n.ts` and node middleware in `src/server/helpers/middleware.ts` to load the unified translations.

3. **Codebase Migration**:
   * Replaced all occurrences of `useTranslation('namespace')` and `withTranslation('namespace')` with parameterless `useTranslation()` / `withTranslation()` across ~50 component files.
   * Migrated all flat translation keys to the new nested dot-path notation (e.g. `t('button.submit')` to `t('common.button.submit')`).
   * Updated `test/src/common/i18n.js` to reflect the new nested JSON structure and dot-path lookups.

# AI usage

* [ ] I did not use any AI

#### If you did use AI:

 
# Action

*None.*

### DIFF
--- a/src/client/components/pages/parts/collections-table.js
+++ b/src/client/components/pages/parts/collections-table.js
@@ -50,7 +50,7 @@ class CollectionsTable extends React.Component {
 			<DropdownButton
 				className="margin-bottom-d5"
 				id="entity-type-select"
-				title={this.props.type ? translate(`common:entityType.${_.camelCase(this.props.type)}`) : translate('common:type')}
+				title={this.props.type ? translate(`common.entityType.${_.camelCase(this.props.type)}`) : translate('common.type')}
 				variant="primary"
 				onSelect={this.handleEntitySelect}
 			>
@@ -60,7 +60,7 @@ class CollectionsTable extends React.Component {
 						key={entityType}
 					>
 						{genEntityIconHTMLElement(entityType)}
-						{translate(`common:entityType.${_.camelCase(entityType)}`)}
+						{translate(`common.entityType.${_.camelCase(entityType)}`)}
 					</Dropdown.Item>
 				))}
 				<Dropdown.Divider/>
@@ -68,7 +68,7 @@ class CollectionsTable extends React.Component {
 					eventKey={null}
 					key="allTypes"
 				>
-					{translate('pages:collections.allTypes')}
+					{translate('pages.collections.allTypes')}
 				</Dropdown.Item>
 			</DropdownButton>
 		);
@@ -86,7 +86,7 @@ class CollectionsTable extends React.Component {
 					variant="warning"
 				>
 					<FontAwesomeIcon icon={faPlus}/>
-					&nbsp;{translate('pages:collections.createButton')}
+					&nbsp;{translate('pages.collections.createButton')}
 				</Button>
 			);
 		}
@@ -103,7 +103,7 @@ class CollectionsTable extends React.Component {
 					type="button"
 					variant="success"
 				>
-					{translate('pages:collections.myCollectionsButton')}
+					{translate('pages.collections.myCollectionsButton')}
 				</Button>
 			);
 		}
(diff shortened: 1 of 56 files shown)


src/client/components/pages/parts/collections-table.js
```diff
@@ -131,26 +131,26 @@ class CollectionsTable extends React.Component {
 						>
 							<thead>
 								<tr>
-									<th width="16%">{translate('common:name')}</th>
-									<th width="33%">{translate('common:description')}</th>
-									<th width="16%">{translate('common:entityType')}</th>
-									<th width="16%">{translate('common:entities')}</th>
+									<th width="16%">{translate('common.name')}</th>
+									<th width="33%">{translate('common.description')}</th>
+									<th width="16%">{translate('common.entityType._self')}</th>
```
MonkeyDo: This doesn't look right.

src/client/entity-editor/relationship-editor/relationship-editor.tsx
```diff
@@ -428,7 +428,7 @@ class RelationshipModal
 		// We are disabling this rule because we are already sanitizing the html here
 		return (
 			<Form.Group>
-				<Form.Label>{translate('common:Relationship')}</Form.Label>
+				<Form.Label>{translate('common.Relationship')}</Form.Label>
```
MonkeyDo: This capitalization isn't right.
Didn't cathc that in the previous passes, but it should be renamed to "relationship" (and in the source file as well)

MonkeyDo (changes requested):
Some minor things

src/client/components/pages/parts/collections-table.js
garvthakre: i didn't got what i need to fix it here .

src/client/entity-editor/relationship-editor/relationship-editor.tsx
garvthakre: done ...

src/client/components/pages/parts/collections-table.js
MonkeyDo: Sorry, that wasn't clear. `common.entityType._self` should likely be `common.entityType` instead. Not sure why the `._self` was added.

garvthakre: well it was because of the format of the translation.json ( because entity type has strings inside that so thats why i wrote `._self` ) , can you please have a look at line 32 ..

MonkeyDo: Oh I see, totally missed that, thank you for the explanation.

MonkeyDo: OK, in that case I think it would be clearer to those who come after to separate `common.entityType` = "Entity type" (note the capitalization, should be left to CSS if we want Title Case)
and separately
`common.entityTypes.XXXX`>

So entityType a single string, and entityTypes (with an 's') an object of the various types

src/client/entity-editor/relationship-editor/relationship-editor.tsx
MonkeyDo: I think you forgot to push that commit.

garvthakre: no i didnt push the commit yet , will push after fixing the both bug.

src/client/components/pages/parts/collections-table.js
garvthakre: okay got it

garvthakre: but then in this way i need to make alot of changes in files too and also `entitytypes` not feels good for eg (`entitytypes.area` is not looking good and consistent whereas `entitytype.area` is more better option here naturally i guess ) , dont u think it will be better if i write `entitytype.singular` and `entitytype.plural`.

MonkeyDo: Then for a simpler solution I suggest renaming the string for "Entity type" to something like `common.entityTypeLabel` or some other generic name, and keeping it out of the `common.entityType` object.
It does not belong in the same object, because it is not itself an entity type.

garvthakre: i think `common.entityTypeLabel` is good here .

src/client/entity-editor/relationship-editor/relationship-editor.tsx
garvthakre: okay i fixed both the bug , can u have a look at it.

MonkeyDo (approved):
Good to go, thank you!

